PR #1, opened 2019-05-17
libdrgn: add /usr/lib/debug/boot in the vmlinux_paths

sdimitro:
Ubuntu-based distros tend to put vmlinux with debug info
under `/usr/lib/debug/boot/vmlinux-<version>`.

Side-notes:
Took the liberty of re-ordering the paths to alphabetical order.

### DIFF
--- a/libdrgn/program.c
+++ b/libdrgn/program.c
@@ -487,9 +487,10 @@ static struct drgn_error *find_elf_symbol(Elf *elf, Elf_Scn *symtab_scn,
 }
 
 static const char * const vmlinux_paths[] = {
-	"/usr/lib/debug/lib/modules/%s/vmlinux",
 	"/boot/vmlinux-%s",
 	"/lib/modules/%s/build/vmlinux",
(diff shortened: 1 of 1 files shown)


libdrgn/program.c
```diff
@@ -487,9 +487,10 @@ static struct drgn_error *find_elf_symbol(Elf *elf, Elf_Scn *symtab_scn,
 }
 
 static const char * const vmlinux_paths[] = {
-	"/usr/lib/debug/lib/modules/%s/vmlinux",
```
osandov: I think my reasoning with the order was that the files under /usr/lib/debug always have debug information, but maybe in /boot or /lib/modules they don't, so it'd be best to check /usr/lib/debug first. Could you keep the order and maybe add a comment like "Check /usr/lib/debug first, since those will always have debug information."

osandov:
I have a conflicting change in the works, so I went ahead and merged this with the added comment. Thanks!

osandov:
I have a conflicting change in the works, so I went ahead and merged this with the added comment. Thanks!

PR #2, opened 2019-06-05
helpers: fix for_each_cpu to use new attributes

naota:
for_each_cpu is using old attributes "sizeof" and "size". Fix them to
use new ones.

Before patch:
>>> for x in for_each_online_cpu(prog): print(x)
...
  File "/home/naota/drgn/drgn/helpers/linux/cpumask.py", line 29, in for_each_cpu
    word_bits = 8 * bits.type_.type.sizeof()
AttributeError: '_drgn.Type' object has no attribute 'sizeof'

After patch:
>>> for x in for_each_online_cpu(prog): print(x)
...
0
1
2
3
4
5
6
7

### DIFF
--- a/drgn/helpers/linux/cpumask.py
+++ b/drgn/helpers/linux/cpumask.py
@@ -26,8 +26,8 @@ def for_each_cpu(mask):
     :rtype: Iterator[int]
     """
     bits = mask.bits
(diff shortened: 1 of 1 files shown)


osandov:
Thanks for the fix! This reminds me that I need to add tests for the helpers.

PR #5, opened 2019-06-18
cli: Assume --kernel if no target is specified

htejun:
cli currently requires one of --kernel, --core or --pid to be
specified.  Let's assume --kernel so that we can get away with older
env(1) which doesn't understand -S option when writing debug scripts
which uses drgn as the interpreter.  While this is a bit of a kludge,
other options aren't useful on the shebang lines anyway given that
they need an extra dynamic argument.


osandov:
Thanks, @htejun!

PR #6, opened 2019-06-28
Added path_lookup function

carmeli-tamir:
The function returns a struct path object of a given path name.
I was wondering whether to call it kern_path, the name is debatable.

For example, d_path(path_lookup(prog, '/etc/shadow') == '/etc/shadow' :)


osandov:
I really like this trick, but since it's only applicable to live systems (and doesn't play too well with mount namespaces), I don't know that it's suitable to be a generic helper. I'm pretty sure I wrote a proper path walk at one point, I can dig that up and add it to the helpers.

osandov:
@carmeli-tamir could you try out the latest commit? I think it'll do the trick.

PR #7, opened 2019-07-01
libdrgn: read kdumps (new dependecy: libkdumpfile)

sdimitro:
## Intro

Note that this is a WIP to get feedback on the design/organization of the code.
After writing this in C, I realized that it could have been easily implemented in
Python too since libkdumpfile provides wrappers for Python. I'm not sure what
would be preferable at this point but it wouldn't be a problem for me rewriting
this in Python.

## Preliminary Feature Testing

Look at the `jiffies` of kdump in `crash`:
```
$ crash ../dump/vmlinux-4.15.0-47-generic ../dump/dump.201904221843

crash 7.2.1
Copyright (C) 2002-2017  Red Hat, Inc.
... < cropped output > ...
crash> p jiffies
jiffies = $2 = 4294918568
```

Using the same address to read the `jiffies` from drgn prototype we get the same value:
```
$ sudo drgn -c dump/dump.201904221843 -s dump/vmlinux-4.15.0-47-generic
could not get debugging information for:
kernel (could not find vmlinux)
kernel modules (could not find loaded kernel modules: could not find 'struct module')
drgn 0.0.1 (using Python 3.6.7)
For help, type help(drgn).
>>> import drgn
>>> from drgn import cast, container_of, execscript, NULL, Object, reinterpret
>>> prog['jiffies']
(volatile unsigned long)4294918568
>>>
```

## Testing that feature is indeed an optional dependency

The above testing was done by enabling the feature through `setup.py` like this:
```
--- a/setup.py
+++ b/setup.py
@@ -40,6 +40,7 @@ class my_build_ext(build_ext):
             args = [
                 os.path.relpath('libdrgn/configure', self.build_temp),
                 '--disable-static', '--with-python=' + sys.executable,
+                '--with-libkdumpfile=yes',
             ]
             try:
                 subprocess.check_call(args, cwd=self.build_temp)
```

Without the feature trying to open the same crash dump we get the following error:
```
$ sudo drgn -c dump/dump.201904221843 -s dump/vmlinux-4.15.0-47-generic
  File "/usr/local/bin/drgn", line 11, in <module>
    load_entry_point('drgn==0.0.1', 'console_scripts', 'drgn')()
  File "/usr/local/lib/python3.6/dist-packages/drgn-0.0.1-py3.6-linux-x86_64.egg/drgn/internal/cli.py", line 83, in main
    prog.set_core_dump(args.core)
ValueError: drgn configured without libkdumpfile
```

## Work left
* finalize ifdef structure for making this an optional dependency
* figure out if automated testing is possible?
* style fixes
* copy Petr Tesarik (maintainer of libkdumpfile) in the final rounds of code review to ensure proper use of the library (and also give a heads up that we exist as an official consumer)

### DIFF
--- a/libdrgn/Makefile.am
+++ b/libdrgn/Makefile.am
@@ -52,9 +52,14 @@ libdrgnimpl_la_SOURCES = binary_search_tree.h \
 			 vector.h
 
 libdrgnimpl_la_CFLAGS = -fvisibility=hidden -fopenmp $(libelf_CFLAGS) \
-			$(libdw_CFLAGS) $(libkdumpfile_CFLAGS)
+			$(libdw_CFLAGS)
 libdrgnimpl_la_CPPFLAGS = -D_GNU_SOURCE
-libdrgnimpl_la_LIBADD = $(libelf_LIBS) $(libdw_LIBS) $(libkdumpfile_LIBS)
+libdrgnimpl_la_LIBADD = $(libelf_LIBS) $(libdw_LIBS)
+
+if WITH_LIBKDUMPFILE
+libdrgnimpl_la_CFLAGS += $(libkdumpfile_CFLAGS)
+libdrgnimpl_la_LIBADD += $(libkdumpfile_LIBS)
+endif
 
 lib_LTLIBRARIES = libdrgn.la
 
--- a/libdrgn/Makefile.am
+++ b/libdrgn/Makefile.am
@@ -18,7 +18,6 @@ libdrgnimpl_la_SOURCES = binary_search_tree.h \
 			 hash_table.h \
 			 internal.c \
 			 internal.h \
-			 kdump_reader.c \
 			 kdump_reader.h \
 			 language.h \
 			 language_c.c \
@@ -57,6 +56,7 @@ libdrgnimpl_la_CPPFLAGS = -D_GNU_SOURCE
 libdrgnimpl_la_LIBADD = $(libelf_LIBS) $(libdw_LIBS)
 
 if WITH_LIBKDUMPFILE
+libdrgnimpl_la_SOURCES += kdump_reader.c
 libdrgnimpl_la_CFLAGS += $(libkdumpfile_CFLAGS)
 libdrgnimpl_la_LIBADD += $(libkdumpfile_LIBS)
 endif
--- a/libdrgn/Makefile.am
+++ b/libdrgn/Makefile.am
@@ -18,6 +18,8 @@ libdrgnimpl_la_SOURCES = binary_search_tree.h \
 			 hash_table.h \
 			 internal.c \
 			 internal.h \
+			 kdump_reader.c \
+			 kdump_reader.h \
 			 language.h \
 			 language_c.c \
 			 lexer.c \
@@ -50,9 +52,9 @@ libdrgnimpl_la_SOURCES = binary_search_tree.h \
 			 vector.h
 
 libdrgnimpl_la_CFLAGS = -fvisibility=hidden -fopenmp $(libelf_CFLAGS) \
-			$(libdw_CFLAGS)
+			$(libdw_CFLAGS) $(libkdumpfile_CFLAGS)
 libdrgnimpl_la_CPPFLAGS = -D_GNU_SOURCE
-libdrgnimpl_la_LIBADD = $(libelf_LIBS) $(libdw_LIBS)
+libdrgnimpl_la_LIBADD = $(libelf_LIBS) $(libdw_LIBS) $(libkdumpfile_LIBS)
 
 lib_LTLIBRARIES = libdrgn.la
 
--- a/libdrgn/configure.ac
+++ b/libdrgn/configure.ac
@@ -48,6 +48,22 @@ python3-devel or python3-dev) or specify the location of the Python development
 headers by setting the PYTHON_CPPFLAGS environment variable.])])
 	    CPPFLAGS="$save_CPPFLAGS"])
 
+AC_ARG_WITH([libkdumpfile],
+	    [AS_HELP_STRING([--with-libkdumpfile@<:@=ARG@:>@],
+			    [leverage libkdumpfile to read crash dumps using
+			     the makedumpfile format. ARG may be yes or no
+			     @<:default=no:>@])],
+           [], [with_libkdumpfile=no])
+AS_IF([test "x$with_libkdumpfile" != xno],
+      [
+       PKG_CHECK_MODULES(libkdumpfile, [libkdumpfile])
+       AM_CONDITIONAL([WITH_LIBKDUMPFILE], true)
+       AC_DEFINE(LIBKDUMPFILE, true)
+      ],
+      [
+       AM_CONDITIONAL([WITH_LIBKDUMPFILE], false)
+      ])
+
 PKG_PROG_PKG_CONFIG
 PKG_CHECK_MODULES(libelf, [libelf])
 AC_SUBST([libelf_CFLAGS])
@@ -56,9 +72,6 @@ dnl We need dwarf_cu_getdwarf() which was added in elfutils 0.160.
 PKG_CHECK_MODULES(libdw, [libdw >= 0.160])
 AC_SUBST([libdw_CFLAGS])
 AC_SUBST([libdw_LIBS])
-PKG_CHECK_MODULES(libkdumpfile, [libkdumpfile])
-AC_SUBST([libkdumpfile_CFLAGS])
-AC_SUBST([libkdumpfile_LIBS])
 
 AC_CONFIG_FILES([Makefile])
 AC_OUTPUT
--- a/libdrgn/configure.ac
+++ b/libdrgn/configure.ac
@@ -56,6 +56,9 @@ dnl We need dwarf_cu_getdwarf() which was added in elfutils 0.160.
 PKG_CHECK_MODULES(libdw, [libdw >= 0.160])
 AC_SUBST([libdw_CFLAGS])
 AC_SUBST([libdw_LIBS])
+PKG_CHECK_MODULES(libkdumpfile, [libkdumpfile])
+AC_SUBST([libkdumpfile_CFLAGS])
+AC_SUBST([libkdumpfile_LIBS])
 
 AC_CONFIG_FILES([Makefile])
 AC_OUTPUT
--- a/libdrgn/configure.ac
+++ b/libdrgn/configure.ac
@@ -49,20 +49,18 @@ headers by setting the PYTHON_CPPFLAGS environment variable.])])
 	    CPPFLAGS="$save_CPPFLAGS"])
 
 AC_ARG_WITH([libkdumpfile],
-	    [AS_HELP_STRING([--with-libkdumpfile@<:@=ARG@:>@],
-			    [leverage libkdumpfile to read crash dumps using
-			     the makedumpfile format. ARG may be yes or no
-			     @<:default=no:>@])],
-           [], [with_libkdumpfile=no])
-AS_IF([test "x$with_libkdumpfile" != xno],
-      [
-       PKG_CHECK_MODULES(libkdumpfile, [libkdumpfile])
-       AM_CONDITIONAL([WITH_LIBKDUMPFILE], true)
-       AC_DEFINE(LIBKDUMPFILE, true)
-      ],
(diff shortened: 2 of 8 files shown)


sdimitro:
Seems like Travis is failing because the images do not have `libkdumpfile` installed

libdrgn/program.c
```diff
@@ -513,6 +514,54 @@ drgn_program_check_initialized(struct drgn_program *prog)
...
+	if (err)
+		goto out_kdump;
+
+	err = drgn_kdump_get_osrelease(ctx, prog->vmcoreinfo.osrelease);
+	if (err)
+		goto out_kdump;
+
+	err = drgn_kdump_get_page_size(ctx, &prog->vmcoreinfo.page_size);
+	if (err)
+		goto out_kdump;
```
osandov: This is the only part I don't like at a high level. drgn already has code for parsing vmcoreinfo, so I'd rather use that and avoid having to add a new call here for every field we start using. We'll probably just need to do some minimal parsing straight from the fd.

libdrgn/configure.ac
```diff
@@ -56,6 +56,9 @@ dnl We need dwarf_cu_getdwarf() which was added in elfutils 0.160.
 PKG_CHECK_MODULES(libdw, [libdw >= 0.160])
 AC_SUBST([libdw_CFLAGS])
 AC_SUBST([libdw_LIBS])
+PKG_CHECK_MODULES(libkdumpfile, [libkdumpfile])
```
osandov: Since libkdumpfile doesn't appear to be packaged for most distros, let's make this dependency optional.

libdrgn/program.c
```diff
@@ -534,6 +583,11 @@ drgn_program_set_core_dump(struct drgn_program *prog, const char *path)
 	if (prog->core_fd == -1)
 		return drgn_error_create_os(errno, path, "open");
 
+	if (has_kdump_signature(prog->core_fd, &err))
```
osandov: As a side note, normal ELF vmcores don't include vmalloc or per CPU addresses in the program headers, so drgn currently can't read from them. It seems like libkdumpfile can, so perhaps as a followup change, we should also use libkdumpfile for normal ELF vmcores.

libdrgn/kdump_reader.c
```diff
@@ -0,0 +1,286 @@
...
+		return false;
+
+	return true;
+}
+
+struct drgn_error *drgn_kdump_init(kdump_ctx_t **ctx, int fd)
+{
+	kdump_ctx_t *context = kdump_new();
+	if (!ctx) {
+		close(fd);
```
osandov: In general, whatever scope created a given resource should clean up that resource unless there's a clear reason to transfer ownership. Plus, it looks like this is already being closed on error by the caller.

libdrgn/kdump_reader.c
```diff
@@ -0,0 +1,286 @@
...
+
+#include <fcntl.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+#include "kdump_reader.h"
+#include "program.h"
+
+bool has_kdump_signature(int fd, struct drgn_error **err)
```
osandov: The calling convention in drgn is that if an error may be returned, the error is always the return value, and anything else is returned through a return parameter.

libdrgn/kdump_reader.c
```diff
@@ -0,0 +1,286 @@
...
+{
+	char signature[SIG_LEN];
+
+	int n = read(fd, signature, sizeof (signature));
+	if (n != sizeof (signature)) {
+		*err = drgn_error_format(DRGN_ERROR_FAULT,
+		                         "short read from file");
```
osandov: This needs the usual handling for short reads and `EINTR`. And, this should use `drgn_error_create_os()`.

libdrgn/kdump_reader.c
```diff
@@ -0,0 +1,286 @@
...
+	return (!strcmp(kdump_arch_attr, KDUMP_ARCH_X86_64) ||
+	        !strcmp(kdump_arch_attr, KDUMP_ARCH_AARCH64) ||
+	        !strcmp(kdump_arch_attr, KDUMP_ARCH_ALPHA) ||
+	        !strcmp(kdump_arch_attr, KDUMP_ARCH_IA64) ||
+	        !strcmp(kdump_arch_attr, KDUMP_ARCH_PPC64) ||
+	        !strcmp(kdump_arch_attr, KDUMP_ARCH_S390X));
+}
+
+struct drgn_error *drgn_kdump_get_arch(kdump_ctx_t *ctx,
+                                       enum drgn_architecture_flags *arch)
```
osandov: My stack trace work is going to conflict with this bit since I'm changing around how drgn represents the architecture, but I can work that out on my end.

libdrgn/program.c
sdimitro: Great idea. Decided to use `kdump_vmcoreinfo_raw()` provided by `libkdumpfile` to get the whole thing as a string and then used `parse_vmcoreinfo()` which is provided by `drgn`. Does this work better?

libdrgn/kdump_reader.c
sdimitro: I'll keep an eye too in case your change makes it first.

sdimitro:
Asked a couple of questions above, and applied the rest of the feedback in a separate commit. As this is still a WIP, I'll leave each round of feedback as a separate commit and squash them into one commit before merging.

libdrgn/configure.ac
sdimitro: I did not forget about this. Coming from illumos, I'm new to autotools+pkgconf so this may take me some time.

sdimitro: Just pushed the work to ensure that this is an optional dependency. This is my first attempt at doing anything like this so please do not hesitate to ask for corrections or nitpick.

libdrgn/Makefile.am
```diff
@@ -18,6 +18,8 @@ libdrgnimpl_la_SOURCES = binary_search_tree.h \
 			 hash_table.h \
 			 internal.c \
 			 internal.h \
+			 kdump_reader.c \
```
osandov: Instead of having it here, we can add `libdrgnimpl_la_SOURCES += kdump_reader.c` in the `if WITH_LIBKDUMPFILE` block, that way you don't need to put the big ifdef in `kdump_reader.c`.

libdrgn/program.c
```diff
@@ -164,14 +165,61 @@ drgn_program_check_initialized(struct drgn_program *prog)
 	return NULL;
 }
 
+#ifdef LIBKDUMPFILE
```
osandov: Can you move this function to `kdump_reader.c`, again so we don't need the ifdef? Then we can also make a bunch of the kdump helpers static and will only need an ifdef in `kdump_reader.h` for this function.

libdrgn/program.c
```diff
@@ -185,6 +233,14 @@ drgn_program_set_core_dump(struct drgn_program *prog, const char *path)
 	if (prog->core_fd == -1)
 		return drgn_error_create_os("open", errno, path);
 
+#ifdef LIBKDUMPFILE
+	err = has_kdump_signature(prog->core_fd, &is_kdump);
```
osandov: How about we:
1. Move `has_kdump_signature()` to `program.c` so it's always available.
2. Add a dummy inline `drgn_program_set_kdump()` in `kdump.h` which is only defined `ifndef LIBKDUMPFILE` which just returns an error explaining that drgn was configured without libkdumpfile.

That way, the user gets a useful error message (and as a bonus we avoid an ifdef in the main logic).

libdrgn/program.h
```diff
@@ -28,10 +28,12 @@
  * @{
  */
 
+#define OSRELEASE_LEN 128
```
osandov: This change is no longer necessary.

libdrgn/kdump_reader.c
```diff
@@ -0,0 +1,199 @@
...
+
+	kdump_addrspace_t as = (physical) ? KDUMP_KPHYSADDR : KDUMP_KVADDR;
+	kdump_status ks = kdump_read(ctx, as, address, buf, &nread);
+	if (ks != KDUMP_OK) {
+		return drgn_error_format(DRGN_ERROR_OTHER,
+		                         "kdump_read failed: %s",
+					 kdump_get_err(ctx));
+	}
+
+	if (nread != count) {
```
osandov: Can `kdump_read()` validly return a short read, or would this be a bug in libkdumpfile?

libdrgn/kdump_reader.c
```diff
@@ -0,0 +1,199 @@
...
+}
+
+struct drgn_error *drgn_read_kdump(void *buf, uint64_t address, size_t count,
+                                   uint64_t offset, void *arg, bool physical)
+{
+	kdump_ctx_t *ctx = arg;
+	size_t nread = count;
+
+	/* zero-out buffer before doing anything */
+	memset(buf, 0, count);
```
osandov: This shouldn't be necessary unless there's a bug in libkdumpfile.

libdrgn/kdump_reader.c
sdimitro: You are right, I checked the code and it cannot validly return a short read,
as it has its own loop (snippet below).

from https://github.com/ptesarik/libkdumpfile/blob/3943558af6fca25785faca69078b6e1dee1f4fac/src/kdumpfile/read.c :
```
...
kdump_status
kdump_read(kdump_ctx_t *ctx, kdump_addrspace_t as, kdump_addr_t addr,
            void *buffer, size_t *plength)
{
        kdump_status ret;

        clear_error(ctx);
        rwlock_rdlock(&ctx->shared->lock);
        ret = read_locked(ctx, as, addr, buffer, plength);
        rwlock_unlock(&ctx->shared->lock);
        return ret;
}
...
/**  Internal version of @ref kdump_read
 * @param         ctx      Dump file object.
 * @param[in]     as       Address space of @p addr.
 * @param[in]     addr     Any type of address.
 * @param[out]    buffer   Buffer to receive data.
 * @param[in,out] plength  Length of the buffer.
 * @returns                Error status.
 *
 * Use this function internally if the shared lock is already held
 * (for reading or writing).
                                                                                                                     62,0-1        58%
read_locked(kdump_ctx_t *ctx, kdump_addrspace_t as, kdump_addr_t addr,
            void *buffer, size_t *plength)
{
        struct page_io pio;
        size_t remain;
        kdump_status ret;

        ret = KDUMP_OK;
        remain = *plength;
        while (remain) {
                size_t off, partlen;

                pio.addr.as = as;
                pio.addr.addr = page_align(ctx, addr);
                ret = get_page(ctx, &pio);
                if (ret != KDUMP_OK)
                        break;

                off = addr % get_page_size(ctx);
                partlen = get_page_size(ctx) - off;
                if (partlen > remain)
                        partlen = remain;
                memcpy(buffer, pio.chunk.data + off, partlen);
                put_page(ctx, &pio);
                addr += partlen;
                buffer += partlen;
                remain -= partlen;
        }

        *plength -= remain;
        return ret;
}
...
```

osandov:
I pushed a tweak to the configure script so that we automatically use libkdumpfile if it's available. A few final things:

* Let's rename `kdump_reader.[ch]` to `kdump.[ch]` since they have more logic than just the reader.
* I don't see where we're freeing the kdump context. We can probably just stuff it in `struct program` as a `void *` and close it if it's not `NULL`.
* Before I merge this, I'll reformat the code a bit to the Linux kernel coding style, but you don't need to worry about that.

osandov:
I rebased this on top of the platform change in the devel branch. Can you please test it out? If everything works well, I'll merge it.

sdimitro:
Just tested the PR and the output is the same as the description (I tried a few other globals and types from modules that I loaded with `-s`) so from that perspective we are good to go.

I remember you mentioning on our DMs yesterday, that you went ahead and did the freeing of the kdump context yourself? I didn't see that in the changes that I pulled (neither the renaming `s/kdump_reader/kdump/` for the new files). Do you want me to go ahead and do it. or you do have changes but they are still local?

osandov:
Ah, I pushed the changes to the `devel` branch on my repo rather than updating your branch.

sdimitro:
Gave the `devel` branch a run. Works the same :-)

osandov:
Merged, thanks!

PR #13, opened 2019-09-21
Add support for GCC dwarf2 data

jwadams:
libdrgn/dwarf_info_cache.c uses dwarf_formudata() to pull constants
(like structure member offsets) out of the dwarf data.  Unfortunately,
GCC in -gdwarf-2 mode generates structure constants as blocks with
a single "DW_OP_plus_uconst" operation, which doesn't work with
dwarf_formudata().

This patch introduces a drgn_formudata() wrapper around
dwarf_formudata(), which upon failure sees if the attribute is a
block of that form, and if so returns the proper value.  We have all
the existing dwarf_formudata() callsites use the new wrapper.  This
allows for debugging kernels with dwarf-2 data (which is the default
for x86 linux)

### DIFF
--- a/libdrgn/dwarf_info_cache.c
+++ b/libdrgn/dwarf_info_cache.c
@@ -42,6 +42,53 @@ static bool drgn_type_realloc(struct drgn_type **type, size_t capacity,
 	return true;
 }
 
+/*
+ * We use dwarf_formudata() to pull constants out of dwarf attributes.
+ * Unfortunately, GCC's dwarfv2 generator for structure member offsets uses
+ * simple "blocks" instead of the constants that dwarf_formudata() expects.
+ *
+ * So if dwarf_formudata() fails, we handle the special case of the attribute
+ * being a block consisting of a single DW_OP_plus_uconst operation.
+ */
+static int drgn_formudata(Dwarf_Attribute *attr, Dwarf_Word *return_uval)
+{
+	Dwarf_Block block;
+
+	/* dwarf4 common case: constant value */
+	if (dwarf_formudata(attr, return_uval) == 0) {
+		return (0);
+	}
+	/* Try for the dwarf2 case, bailing out if anything looks odd */
+	if (dwarf_formblock(attr, &block) == 0) {
+		unsigned char *const beg = block.data;
+		unsigned char *const end = beg + block.length;
+		unsigned char *cur;
+		int nth;
+		Dwarf_Word uval;
+		uval = 0;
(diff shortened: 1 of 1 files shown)


libdrgn/dwarf_info_cache.c
```diff
@@ -42,6 +42,53 @@ static bool drgn_type_realloc(struct drgn_type **type, size_t capacity,
...
+static int drgn_formudata(Dwarf_Attribute *attr, Dwarf_Word *return_uval)
+{
+	Dwarf_Block block;
+
+	/* dwarf4 common case: constant value */
+	if (dwarf_formudata(attr, return_uval) == 0) {
+		return (0);
+	}
+	/* Try for the dwarf2 case, bailing out if anything looks odd */
+	if (dwarf_formblock(attr, &block) == 0) {
```
osandov: Can we use `dwarf_getlocation()`? See `drgn_object_from_dwarf_variable()` for an example.

libdrgn/dwarf_info_cache.c
```diff
@@ -42,6 +42,53 @@ static bool drgn_type_realloc(struct drgn_type **type, size_t capacity,
...
 
+/*
+ * We use dwarf_formudata() to pull constants out of dwarf attributes.
+ * Unfortunately, GCC's dwarfv2 generator for structure member offsets uses
+ * simple "blocks" instead of the constants that dwarf_formudata() expects.
+ *
+ * So if dwarf_formudata() fails, we handle the special case of the attribute
+ * being a block consisting of a single DW_OP_plus_uconst operation.
+ */
+static int drgn_formudata(Dwarf_Attribute *attr, Dwarf_Word *return_uval)
```
osandov: How about naming this `drgn_dwarf_uconst()` to separate its naming from any particular `DW_FORM`?

jwadams: Yeah, this would be fine.

libdrgn/dwarf_info_cache.c
jwadams: We could; my only worry is the overhead of the memory allocation (internal to libdw, which caches them, presumably forever) for each every structure member we ever see.  It looks like it would work, though.  If you're not worried about that overhead, it would certainly make this function simpler.

osandov: Yeah let's do the simpler way for now. If it ends up causing issues, we can bring back your optimized version.

osandov:
I ran into this again recently, so I finally got around to applying your approach with some updates. (Belated) thanks!

PR #31, opened 2019-12-14
libdrgn: fix regression in kernel module loading

sdimitro:
Commit f327552229658d85b791c6b0613efb96b8a9da24 introduced a
regression resulting in kernel modules not loaded at the right
offset. This patch fixes the regression.


osandov:
I tweaked the wording a bit. I'll merge it once tests pass. Thanks!

PR #33, opened 2019-12-21
examples: Add tcp_sock.py

rdna:
Add example script that goes over all TCP sockets and for every socket
prints:
- TCP state;
- IP:port pair;
- cgroup v2 path.

The script shows basic operations with sockets including getting `struct
sock`, `struct inet_sock`, socket fields and cgroup v2 of socket.

It provides output similar to `ss -nt` or `ss -lnt`:
```
% sudo python36 -m drgn examples/linux/tcp_sock.py | shuf -n 5
TIME_WAIT                                       [::1]:2560                                      [::1]:44041
ESTABLISHED                                     [::1]:54178                                     [::1]:11101  system.slice/async-distillery-server.service
LISTEN                                           [::]:3201                                       [::]:0      system.slice/fbflow.service
ESTABLISHED           [2401:db00:21:718a:face:0:31:0]:2406            [2401:db00:12:a106:face:0:1e:0]:46306  system.slice/configerator_proxy2.service
LISTEN                                        0.0.0.0:11150                                   0.0.0.0:0      system.slice/system-mcrouter.slice/mcrouter@tao.service
```

Since it requires iterating over sockets, necessary `hlist_nulls`
helpers are added as well.

Some functions from `tcp_sock.py` (`sk_fullsock`, `sk_nulls_for_each`,
`inet_sk`) are generally useful and can be moved to a library later.

### DIFF
--- a/drgn/helpers/linux/list_nulls.py
+++ b/drgn/helpers/linux/list_nulls.py
@@ -1,0 +1,66 @@
+# SPDX-License-Identifier: GPL-3.0+
+
+"""
+Nulls Lists
+-----------
+
+The ``drgn.helpers.linux.list_nulls`` module provides helpers for working with
+the special version of lists (``struct hlist_nulls_head`` and ``struct
+hlist_nulls_node``) in :linux:`include/linux/list_nulls.h` where the end of
+list is not a ``NULL`` pointer, but a "nulls" marker.
+"""
+
+from drgn import container_of
+
+
+__all__ = [
+    'is_a_nulls',
+    'hlist_nulls_empty',
+    'hlist_nulls_entry',
+    'hlist_nulls_for_each_entry',
+]
+
+
+def is_a_nulls(pos):
+    """
+    .. c:function:: bool is_a_nulls(struct hlist_nulls_node *pos)
+
+    Return whether a a pointer is a nulls marker.
+    """
+    return bool(pos.value_() & 1)
+
+
+def hlist_nulls_empty(head):
+    """
+    .. c:function:: bool hlist_nulls_empty(struct hlist_nulls_head *head)
+
+    Return whether a nulls hash list is empty.
+    """
+    return is_a_nulls(head.first)
+
+
+def hlist_nulls_entry(pos, type, member):
+    """
+    .. c:function:: type *hlist_nulls_entry(struct hlist_nulls_node *pos, type, member)
+
+    Return an entry in a nulls hash list.
+
+    The nulls hash list is assumed to be non-empty.
+    """
+    return container_of(pos, type, member)
+
+
+def hlist_nulls_for_each_entry(type, head, member):
+    """
+    .. c:function:: hlist_nulls_for_each_entry(type, struct hlist_nulls_head *head, member)
+
+    Iterate over all the entries in a nulls hash list specified by ``struct
+    hlist_nulls_head`` head, given the type of the entry and the ``struct
+    hlist_nulls_node`` member in that type.
+
+    :return: Iterator of ``type *`` objects.
+    """
+    pos = head.first
+    while not is_a_nulls(pos):
+        yield hlist_nulls_entry(pos, type, member)
+        pos = pos.next
--- a/examples/linux/tcp_sock.py
+++ b/examples/linux/tcp_sock.py
@@ -1,0 +1,113 @@
+"""List all TCP sockets and their cgroup v2 paths"""
+
+import ipaddress
+import os
+import socket
+import struct
+
+from drgn import Object, cast, container_of
+from drgn.helpers.linux import (
+    get_tcp_states,
+    hlist_for_each,
+    hlist_nulls_empty,
+    sk_fullsock,
+    sk_nulls_for_each,
+)
+
+
+TcpStates = get_tcp_states(prog)
+
+
+def inet_sk(sk):
+    return cast('struct inet_sock *', sk)
+
+
+def _ipv4(be32):
+    return ipaddress.IPv4Address(struct.pack('I', be32.value_()))
+
+
+def _ipv6(in6_addr):
+    return ipaddress.IPv6Address(struct.pack('IIII', *in6_addr.in6_u.u6_addr32))
+
+
+def _brackets(ip):
+    if ip.version == 4:
+        return '{}'.format(ip.compressed)
+    elif ip.version == 6:
+        return '[{}]'.format(ip.compressed)
+    return ''
+
+
+def _ip_port(ip, port):
+    return '{:>40}:{:<6}'.format(_brackets(ip), port)
+
+
+def _cgroup_path(cgroup):
+    cgroup_path = ''
+
+    while cgroup.level > 0:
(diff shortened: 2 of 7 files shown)


rdna:
[Test failure](https://travis-ci.org/osandov/drgn/jobs/627962419?utm_medium=notification&utm_source=github_status
) seems to be unrelated, `apt-get` failed to install something.
Not sure how to restart it though ..

rdna:
In the last pushes: split commonly useful functions to work with nulls lists, network sockets and TCP states into corresponding helper modules.

drgn/helpers/linux/list_nulls.py
```diff
@@ -0,0 +1,66 @@
...
+    Return an entry in a nulls hash list.
+
+    The nulls hash list is assumed to be non-empty.
+    """
+    return container_of(pos, type, member)
+
+
+def hlist_nulls_for_each_entry(head, type, member):
+    """
+    .. c:function:: hlist_nulls_for_each_entry(type *tpos, struct hlist_nulls_node *pos, struct hlist_nulls_head *head, member)
```
osandov: The C signatures are for documenting the helper itself rather than the original C function/macro it came from, so this could be:

```
.. c:function:: hlist_nulls_for_each_entry(type, struct hlist_nulls_head *head, member)
```

On a related note, the names in the C signatures should be consistent with the names in the Python signature for the other helpers.

drgn/helpers/linux/tcp_states.py
```diff
@@ -0,0 +1,54 @@
...
+The ``drgn.helpers.linux.tcp_states`` module provides definitions for the TCP
+protocol sk_state field defined in :linux:`include/net/tcp_states.h`.
+"""
+
+__all__ = [
+    'TcpStates'
+]
+
+
+class TcpStates(object):
```
osandov: Python 3 has the [`enum`](https://docs.python.org/3/library/enum.html) module which we could use here.

examples/linux/tcp_sock.py
```diff
@@ -0,0 +1,110 @@
...
+    elif ip.version == 6:
+        return '[{}]'.format(ip.compressed)
+    return ''
+
+
+def _ip_port(ip, port):
+    return '{:>40}:{:<6}'.format(_brackets(ip), port)
+
+
+def _cgroup_path(cgroup):
```
osandov: This seems like it could be a useful helper.

examples/linux/tcp_sock.py
```diff
@@ -0,0 +1,110 @@
...
+            state,
+            _ip_port(src_ip, src_port),
+            _ip_port(dst_ip, dst_port),
+            cgroup_path
+        ))
+
+        # Uncomment to print whole struct:
+        #   print(sk)
+        #   print(inet)
+        #   print(cgroup)
```
osandov: The indentation looks too deep here.

drgn/helpers/linux/tcp_states.py
```diff
@@ -0,0 +1,54 @@
...
+    SYN_RECV = 3
+    FIN_WAIT1 = 4
+    FIN_WAIT2 = 5
+    TIME_WAIT = 6
+    CLOSE = 7
+    CLOSE_WAIT = 8
+    LAST_ACK = 9
+    LISTEN = 10
+    CLOSING = 11
+    NEW_SYN_RECV = 12
```
osandov: Are these values guaranteed not to change? If not, then we should grab them from the debug info instead of hard-coding. E.g., something like this:

```python
import enum


def get_tcp_states(prog):
    try:
        return prog.cache['TcpStates']
    except KeyError:
        pass
    # The enum type is anonymous, but we can get it from one of the
    # enumerators.
    tcp_states_type = prog['TCP_ESTABLISHED'].type_.enumerators
    enumerators = [
        (name[4:] if name.startswith('TCP_') else name, value)
        for (name, value) in tcp_states_type
        if name != 'TCP_MAX_STATES'
    ]
    TcpStates = enum.IntEnum('TcpStates', enumerators)
    prog.cache['TcpStates'] = TcpStates
    return TcpStates
```

drgn/helpers/linux/list_nulls.py
rdna: Yeah, I mistakenly assumed that it's copy/paste of C signature. Makes sense, I'll cleanup all signatures here and in other places.

examples/linux/tcp_sock.py
rdna: Oops, code move failure, will fix, thanks.

drgn/helpers/linux/tcp_states.py
rdna: Oh, wow.

TIL: that it's possible to get a `enum` from `prog`
TIL: `prog.cache`

These values should never change, since e.g. glibc header `<netinet/tcp.h>` and kernel's `include/uapi/linux/bpf.h` use same enumerations, but new states (even though just one was added since 2005) may appear with delay in new places, e.g. the last enumeration `TCP_NEW_SYN_RECV` was introduced in 2015 but it's not present on my devbox in `<netinet/tcp.h>`. 

The only reason I defined them manually is I didn't know how to get them automatically and getting them automatically is much better IMO since for example it avoids possible problems with "recently" introduced  `TCP_NEW_SYN_RECV`.

I'll switch to your version. Thanks!

examples/linux/tcp_sock.py
rdna: Yep, I have a plan to add a bunch of cgroup related helpers, but wanted to do it in a separate PR when I have cycles + not to make this one too big. LMK if it doesn't work for some reason.

drgn/helpers/linux/tcp_states.py
rdna: While switching to the automatic enum generation I realized that it would require changing all interfaces that uses TCP states to add additional `prog` argument (e.g. `sk_fullsock`).

That's probably not a big deal, but I'm curious if it's possible to get access to `prog` variable w/o passing it as an argument to helper. The benefit would be have helpers with signatures similar to those in kernel (e.g. same `sk_fullsock`)

osandov: As long as the helper takes a `drgn.Object`, you can grab it from `obj.prog_` (e.g., `sk.prog_` for `sk_fullsock`).

rdna:
In the last push:
- consistent order of args in `hlist_nulls_for_each_entry`;
- fixed `c:function` signatures;
- auto generation of TCP states `enum`;
- fixed indentation in the example.

osandov:
Thanks, this is great. The Linux kernel helpers are more fine-grained than we need for drgn, so I moved the tcp_states helpers to a generic tcp.py and renamed sock.py to net.py. I also added an `sk_tcpstate` helper and some unit tests. I'll merge this once tests pass.

osandov:
Oops, the test failure is because in older kernels, the TCP state definitions don't get compiled in if the kernel isn't compiled with TCP. Easy fix, we shouldn't try to access them until we've created the socket.

PR #35, opened 2020-01-03
cgroup v2 helpers

rdna:
Add a bunch of cgroup v2 helpers and use them in examples.

### DIFF
--- a/drgn/helpers/linux/bpf.py
+++ b/drgn/helpers/linux/bpf.py
@@ -1,0 +1,55 @@
+# SPDX-License-Identifier: GPL-3.0+
+
+"""
+BPF
+---
+
+The ``drgn.helpers.linux.bpf`` module provides helpers for working with BPF
+interface in :linux:`include/linux/bpf.h`, :linux:`include/linux/bpf-cgroup.h`,
+etc.
+"""
+
+
+import itertools
+
+from drgn.helpers import enum_type_to_class
+from drgn.helpers.linux.list import list_for_each_entry
+
+
+__all__ = [
+    "cgroup_bpf_prog_for_each",
+    "cgroup_bpf_prog_for_each_effective",
+]
+
+
+def cgroup_bpf_prog_for_each(cgrp, bpf_attach_type):
+    """
+    .. c:function:: cgroup_bpf_prog_for_each(struct cgroup *cgrp, int bpf_attach_type)
+
+    Iterate over all cgroup bpf programs of the given attach type attached to
+    the given cgroup.
+
+    :return: Iterator of ``struct bpf_prog *`` objects.
+    """
+    progs_head = cgrp.bpf.progs[bpf_attach_type]
+    for pl in list_for_each_entry(
+        "struct bpf_prog_list", progs_head.address_of_(), "node"
+    ):
+        yield pl.prog
+
+
+def cgroup_bpf_prog_for_each_effective(cgrp, bpf_attach_type):
+    """
+    .. c:function:: cgroup_bpf_prog_for_each(struct cgroup *cgrp, int bpf_attach_type)
+
+    Iterate over all effective cgroup bpf programs of the given attach type for
+    the given cgroup.
+
+    :return: Iterator of ``struct bpf_prog *`` objects.
+    """
+    prog_array_items = cgrp.bpf.effective[bpf_attach_type].items
+    for i in itertools.count():
+        prog = prog_array_items[i].prog.read_()
+        if not prog:
+            break
+        yield prog
--- a/drgn/helpers/linux/cgroup.py
+++ b/drgn/helpers/linux/cgroup.py
@@ -12,11 +12,16 @@ supported.
 
 from drgn import NULL, cast, container_of
 from drgn.helpers.linux.kernfs import kernfs_name, kernfs_path
+from drgn.helpers.linux.list import list_for_each_entry
 
 __all__ = [
     "cgroup_name",
     "cgroup_parent",
     "cgroup_path",
+    "css_for_each_child",
+    "css_for_each_descendant_pre",
+    "css_next_child",
+    "css_next_descendant_pre",
     "sock_cgroup_ptr",
 ]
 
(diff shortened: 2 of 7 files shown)


drgn/helpers/linux/cgroup.py
```diff
@@ -0,0 +1,77 @@
...
+
+__all__ = [
+        'cgroup_name',
+        'cgroup_parent',
+        'cgroup_path',
+        'sock_cgroup',
+]
+
+
+def sock_cgroup(skcd):
```
osandov: The kernel has `sock_cgroup_ptr`, how about we make this return a `struct cgroup *` instead of `struct cgroup` and rename it for consistency?

drgn/helpers/linux/cgroup.py
```diff
@@ -0,0 +1,77 @@
...
+    """
+    .. c:function:: struct cgroup *sock_cgroup(struct sock_cgroup_data *skcd)
+
+    Get cgroup of a socket by ``struct sock_cgroup_data`` member embedded to
+    every ``struct sock``.
+
+
+    :return: ``struct cgroup`` object.
+    """
+    return Object(skcd.prog_, 'struct cgroup', address=skcd.val)
```
osandov: This can be `cast('struct cgroup *', skcd.val)`.

drgn/helpers/linux/cgroup.py
```diff
@@ -0,0 +1,77 @@
...
+
+    :return: ``struct cgroup`` object or None.
+    """
+    parent_css = cgrp.self.parent
+    if parent_css:
+        return container_of(parent_css, 'struct cgroup', 'self')
+    return None
+
+
+def cgroup_name(cgrp):
```
osandov: sysfs also uses kernfs, so it'd be worth it to add `kernfs_name` and `kernfs_path` helpers and use those for `cgroup_name` and `cgroup_path`. Would you mind doing that?

drgn/helpers/linux/cgroup.py
```diff
@@ -0,0 +1,77 @@
...
+    Get cgroup path.
+
+    :rtype: bytes
+    """
+    names = []
+    while cgrp is not None and cgrp.level >= 0:
+        names.append(cgroup_name(cgrp))
+        cgrp = cgroup_parent(cgrp)
+    names.reverse()
+    return os.path.join(b'', *names)
```
osandov: This is really pedantic, but I'd prefer `b'/'.join(names)` over `os.path.join()`, because we're joining a Linux path, not a path on whatever operating system drgn is running on. And who knows, maybe someone will port drgn to Windows and use it to debug a Linux core dump :)

drgn/helpers/linux/cgroup.py
rdna: Lol, that what I initially did, but then I decided to make it "more python-friendly". Sure, can convert it back to `_ptr` version.

drgn/helpers/linux/cgroup.py
rdna: `kernfs_name` is simple, I'll do it

`kernfs_path` is more involved so I'll do simplified version of it, that later can be extended if needed. By "simplified" I mean I won't implement the whole logic of `kernfs_path_from_node_locked` that is used by `kernfs_path` underneath but will handle only scenario when `from == NULL`, i.e. the path is built from starting root, not from a random `kn` in the fs. LMK if it's not a good idea for some reason.

drgn/helpers/linux/cgroup.py
rdna: Yeah, I saw it in other helpers, but wasn't sure "why" :) OK, makes sense, I'll change it -- not a big deal.

rdna:
In v2:
- renamed `sock_cgroup` to `sock_cgroup_ptr` and changed return type to ptr;
- use `cast` instead of `Object` in `sock_cgroup_ptr`;
- implement `kernfs` helpers and use them in `cgroup_path`, `cgroup_name`;
- avoid `os.path.join`;
- implement additional cgroup v2 helpers to traverse hierarchy and an example -- @osandov please  lmk if it's not a good idea to add new commits into this PR, I can move them into separate one, not a big deal.

drgn/helpers/linux/cgroup.py
```diff
@@ -0,0 +1,183 @@
...
+
+    :return: ``struct cgroup_subsys_state *`` object or None.
+    """
+    css_flags = _get_css_flags(parent.prog_)
+
+    if pos is None:
+        next_ = container_of(
+            parent.children.next, 'struct cgroup_subsys_state', 'sibling'
+        )
+    elif not (pos.flags.value_() & css_flags.CSS_RELEASED):
```
osandov: You don't need the `.value_()` here, drgn objects support all of the usual operators. This can be changed in a few other places, too.

drgn/helpers/linux/cgroup.py
```diff
@@ -0,0 +1,183 @@
...
+        next_ = container_of(
+            pos.sibling.next, 'struct cgroup_subsys_state', 'sibling'
+        )
+    else:
+        for next_ in list_for_each_entry(
+            'struct cgroup_subsys_state',
+            parent.children.address_of_(),
+            'sibling'
+        ):
+            if next_.serial_nr.value_() > pos.serial_nr.value_():
```
osandov: Likewise here, you don't need `.value_()`. Although in this case, it would be a bit more efficient to read `pos.serial_nr` once (under the covers, each access needs to read from `/proc/kcore` or the core dump):

```
pos_serial_nr = pos.serial_nr.value_()
for ...:
    if next_.serial_nr > pos_serial_nr:
        break
```

drgn/helpers/linux/cgroup.py
```diff
@@ -0,0 +1,183 @@
...
+                break
+
+    if next_.sibling.address_of_() != parent.children.address_of_():
+        return next_
+    return None
```
osandov: Hm, I don't know how consistent I've been with this, but I think that for helpers that return a pointer, we should return `NULL` instead of `None`. You can use `NULL('struct cgroup_subsys_state *')`

drgn/helpers/linux/bpf.py
```diff
@@ -0,0 +1,76 @@
...
+
+    :rtype: enum.IntEnum
+    """
+    enum_name = 'BpfAttachTypes'
+    try:
+        return prog.cache[enum_name]
+    except KeyError:
+        enumerators = [
+            (name, value)
+            for (name, value) in prog['__MAX_BPF_ATTACH_TYPE'].type_.enumerators
```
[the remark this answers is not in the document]
rdna: > Since this type isn't anonymous, we can use `prog.type('enum bpf_attach_type')` instead of `prog['__MAX_BPF_ATTACH_TYPE'].type`. Also, we probably want to remove the redundant `BPF_` prefix?

Oh, `prog.type` -- that's definitely better. I tried to do it with `prog['enum bpf_attach_type']`, failed and decided to go with what worked. Sure, will switch to it.

As for `BPF_` prefix IMO it's better to keep it so that e.g. I can copy/paste it from UAPI to my drgn script. Also it simplifies auto-generating types for enums in general -- just use the name as is.

We did a different thing for TCP states enum, but that one is very common in many user-space utils and people got used to `LISTEN` more than `TCP_LISTEN` so there it makes sense.



> At this point I wonder if we want to factor this out into a utility function in `drgn/helpers/__init__.py`. Something like `get_enum(name, type, prefix)`? Feel free to suggest a better name.

Yep, I had same thoughts, but .. I was lazy :) I added the function in `drgn/helpers/__init__.py` and wen can bikesched on names in the next update.

drgn/helpers/linux/cgroup.py
```diff
@@ -0,0 +1,183 @@
...
+    .. c:function:: char *cgroup_path(struct cgroup *cgrp)
+
+    Get cgroup path.
+
+    :rtype: bytes
+    """
+    return kernfs_path(cgrp.kn)
+
+
+def _get_css_flags(prog):
```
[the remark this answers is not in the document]
rdna: Right, it's only for internal usage.

The main reason I did it is to cache the enum to avoid `read()`ing it from kern mem all the time during iteration, but I admin that I don't know the whole picture when drgn reads and when it doesn't so maybe it's premature optimization.

I'll remove it but lmk if caching makes sense here.

rdna:
In v3:
- introduce `get_enum` and switch to it to get enum types;
- use `enum bpf_attach_type` instead of an enumerator to get the type;
- yield `prog` instead of `item.prog` in `cgroup_bpf_prog_for_each_effective`;
- don't use `value_()` where not needed;
- remove `css` enum and access enumerators directly;
- switch from `None` to `NULL`;
- switch to `itertools.count()`.

drgn/helpers/linux/cgroup.py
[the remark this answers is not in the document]
osandov: Enums are available in the kernel debug info, so we don't have to read kernel memory for it. So in this case, caching won't help much. I should document some of the low-level details of drgn that can help people writing helpers :)

drgn/helpers/__init__.py
```diff
@@ -56,3 +57,25 @@ def escape_string(buffer: Iterable[int], escape_single_quote: bool = False,
                                     escape_double_quote=escape_double_quote,
                                     escape_backslash=escape_backslash)
                    for c in buffer)
+
+
+def get_enum(prog: object, name : str, type : str, ignore: Iterable[str] = [],
```
osandov: `prog: drgn.Program`. Also, let's make this take the type as a `drgn.Type` instead of a string. Then the caller can deal with doing `prog.type` or `prog[...].type_` as appropriate.

drgn/helpers/linux/cgroup.py
```diff
@@ -0,0 +1,169 @@
...
+
+def sock_cgroup_ptr(skcd):
+    """
+    .. c:function:: struct cgroup *sock_cgroup_ptr(struct sock_cgroup_data *skcd)
+
+    Get cgroup of a socket by ``struct sock_cgroup_data`` member embedded to
+    every ``struct sock``.
+
+
+    :return: ``struct cgroup *`` object.
```
osandov: This annotation isn't necessary since it's documented by the C signature.

drgn/helpers/linux/cgroup.py
```diff
@@ -0,0 +1,169 @@
...
+    .. c:function:: struct cgroup *cgroup_parent(struct cgroup *cgrp)
+
+    Get parent of cgroup if it exists.
+
+    :return: ``struct cgroup`` object or None.
+    """
+    parent_css = cgrp.self.parent
+    if parent_css:
+        return container_of(parent_css, 'struct cgroup', 'self')
+    return None
```
osandov: Ah here's the place I was thinking of when I said I wasn't consistent with this. Let's make this return a `NULL` object, too, then we can get rid of the `:return:` annotation and instead note that it returns `NULL` if the cgroup doesn't have a parent.

rdna:
In v4:
- better typing in `get_enum`;
- moar `None` to `NULL` in cgroup helpers;
- cleanup `:return:` annotations in cgroup helpers documentation.

osandov:
I ended up taking a different approach for the enum types. I think it makes more sense to work with drgn Objects directly in the helpers and allow the example scripts to create the `IntEnum` types for their own use. I'll merge this as soon as tests pass. Thanks again!

PR #37, opened 2020-01-18
FaultError address attribute and OutOfBoundsError

sdimitro:
Test sample (address of FaultError):
```
    >>> a = Object(prog, 'void *', address=0x1)
    >>> try:
    ...     a.read_()
    ... except drgn.FaultError as err:
    ...     print(err.address)
    ...     print(err.message)
    ...     print(err)
    ...
    1
    could not find memory segment
    could not find memory segment: 0x1
```

Test sample (OutOfBounds):
```
    >>> prog['init_task'].comm.read_()
    (char [16])"swapper/0"
    >>> prog['init_task'].comm.read_()[15]
    (char)0
    >>> prog['init_task'].comm.read_()[16]
    _drgn.OutOfBoundsError: out of bounds of value
```

TODO:
- introduce/edit existing tests
- Write documentation for new address attribute
- code checkstyle


libdrgn/error.c
```diff
@@ -117,6 +119,36 @@ LIBDRGN_PUBLIC struct drgn_error *drgn_error_format(enum drgn_error_code code,
...
+	va_end(ap);
+
+	if (ret == -1)
+                return &drgn_enomem;
+
+	err = drgn_error_create_nodup(code, message);
+	if (err == &drgn_enomem)
+		return err;
+
+	err->address = malloc(sizeof(*err->address));
```
osandov: Why does this need to be malloc'd? Can't we just make `err->address` a `uint64_t` rather than a `uint64_t *`?

If the concern is distinguishing between an address of 0 and an unspecified address, then let's add the "out of bounds" error first so that the address is always specified here.

libdrgn/memory_reader.c
```diff
@@ -240,9 +240,9 @@ struct drgn_error *drgn_memory_reader_read(struct drgn_memory_reader *reader,
 		segment = drgn_memory_segment_tree_search_le(tree,
 							     &address).entry;
 		if (!segment || segment->address + segment->size <= address) {
-			return drgn_error_format(DRGN_ERROR_FAULT,
-						 "could not find memory segment containing 0x%" PRIx64,
-						 address);
+			return drgn_error_format_read(DRGN_ERROR_FAULT, address,
+						      "could not find memory segment containing 0x%" PRIx64,
+						      address);
```
osandov: Duplicating the address between the message and the message is kind of icky. How about we make this `drgn_error_create_fault(const char *message, uint64_t address)` and have `string_builder_append_error()` format it specially as "message: address" like it does for OS errors?

libdrgn/python/error.c
```diff
@@ -123,6 +125,13 @@ DRGNPY_PUBLIC void *set_drgn_error(struct drgn_error *err)
 		PyErr_SetString(PyExc_LookupError, err->message);
 		break;
 	case DRGN_ERROR_FAULT:
+		if (err->address) {
+			address = PyLong_FromUnsignedLongLong(*err->address);
+			if (!address)
+				PyErr_NoMemory();
+		}
+		// XXX: retval?
+		(void) PyObject_SetAttrString(FaultError, "address", address);
```
osandov: This is setting the `address` attribute on the `FaultError` class itself. Ideally we would have the equivalent of the following:

```python
class FaultError(Exception):
    def __init__(self, message, address):
        self.message = message
        self.address = address

    def __str__(self):
        return f'{self.message}: {self.address:#x}'
```

Then here we could use something like
```c
PyObject *exc = PyObject_CallFunction(&FaultError_type, "sK", err->message, err->address);
PyErr_SetObject(&FaultError_type, exc);
```

This means that `FaultError_type` needs to be defined manually rather than with `PyErr_NewExceptionWithDoc`. The best example I could find was [here](https://github.com/python/cpython/blob/47be7d0108b4021ede111dbd15a095c725be46b7/Modules/_ctypes/_ctypes.c#L5515).

sdimitro:
@osandov applied feedback and split the change into to commits. This is ready for another iteration.

libdrgn/python/module.c
```diff
@@ -10,6 +10,52 @@ PyObject *FaultError;
...
+	message = PyObject_GetAttrString(self, "message");
+	if (!message)
+		return NULL;
+
+	address = PyObject_GetAttrString(self, "address");
+	if (!address)
+		return NULL;
+
+	return PyUnicode_FromFormat("%S: %S",
+				    message, PyNumber_ToBase(address, 16));
```
osandov: This is leaking references to the message and address attributes and the return value from `PyNumber_ToBase()`. I'd write it like this:

```c
static PyObject *FaultError_str(PyObject *self)
{
	PyObject *message, *address, *fmt, *ret = NULL;

	message = PyObject_GetAttrString(self, "message");
	if (!message)
		return NULL;

	address = PyObject_GetAttrString(self, "address");
	if (!address)
		goto out_message;

	fmt = PyUnicode_FromString("%s: %#x");
	if (!fmt)
		goto out_address;

	ret = PyUnicode_Format(fmt, message, address);

	Py_DECREF(fmt);
out_address:
	Py_DECREF(address);
out_message:
	Py_DECREF(message);
	return ret;
}
```

libdrgn/python/module.c
```diff
@@ -10,6 +10,52 @@ PyObject *FaultError;
 PyObject *MissingDebugInfoError;
 PyObject *OutOfBoundsError;
 
+int FaultError_init(PyObject *self, PyObject *args, PyObject *kwds)
+{
+	PyObject *address, *message;
+
+	if (!_PyArg_NoKeywords(Py_TYPE(self)->tp_name, kwds))
```
osandov: Out of curiosity, why not allow keyword arguments here?

libdrgn/python/module.c
```diff
@@ -10,6 +10,52 @@ PyObject *FaultError;
...
+
+	address = PyObject_GetAttrString(self, "address");
+	if (!address)
+		return NULL;
+
+	return PyUnicode_FromFormat("%S: %S",
+				    message, PyNumber_ToBase(address, 16));
+}
+
+static PyTypeObject PyExc_FaultError = {
```
osandov: I should've mentioned this on the last revision, the `PyExc_` prefix makes it look like a standard Python error. We can call this `FaultError_type`. Also, the definition of `FaultError_type` and its methods would make more sense in `error.c`.

libdrgn/python/module.c
```diff
@@ -155,10 +201,11 @@ DRGNPY_PUBLIC PyMODINIT_FUNC PyInit__drgn(void)
...
 		goto err;
 
-	FaultError = PyErr_NewExceptionWithDoc("_drgn.FaultError",
-					       drgn_FaultError_DOC, NULL, NULL);
-	if (!FaultError)
+	PyExc_FaultError.tp_base = (PyTypeObject*)PyExc_Exception;
+	if (PyType_Ready(&PyExc_FaultError) < 0)
 		goto err;
+	Py_INCREF(&PyExc_FaultError);
+	FaultError = (PyObject *)&PyExc_FaultError;
```
osandov: We don't need the `FaultError` variable anymore, let's get rid of it and use `&FaultError_type` casted to the appropriate types in the couple of places we currently reference `FaultError`

libdrgn/python/module.c
```diff
@@ -155,10 +201,11 @@ DRGNPY_PUBLIC PyMODINIT_FUNC PyInit__drgn(void)
 	if (add_module_constants(m) == -1)
 		goto err;
 
-	FaultError = PyErr_NewExceptionWithDoc("_drgn.FaultError",
-					       drgn_FaultError_DOC, NULL, NULL);
-	if (!FaultError)
+	PyExc_FaultError.tp_base = (PyTypeObject*)PyExc_Exception;
```
osandov: Style nit: `(PyTypeObject *)`.

libdrgn/python/module.c
```diff
@@ -10,6 +10,52 @@ PyObject *FaultError;
...
+}
+
+static PyTypeObject PyExc_FaultError = {
+	PyVarObject_HEAD_INIT(NULL, 0)
+	.tp_name = "_drgn.FaultError",
+	.tp_basicsize = sizeof(PyBaseExceptionObject),
+	.tp_str = (reprfunc)FaultError_str,
+	.tp_flags = Py_TPFLAGS_DEFAULT | Py_TPFLAGS_BASETYPE,
+	.tp_doc = drgn_FaultError_DOC,
+	.tp_base = 0,
```
osandov: This is implicit and can be removed.

libdrgn/error.c
```diff
@@ -117,6 +120,37 @@ LIBDRGN_PUBLIC struct drgn_error *drgn_error_format(enum drgn_error_code code,
...
+LIBDRGN_PUBLIC struct drgn_error *drgn_error_create_fault(const char *message,
+							  uint64_t address)
+{
+	struct drgn_error *err;
+	int ret;
+
+	err = drgn_error_create(DRGN_ERROR_FAULT, message);
+	if (!err)
+		return err;
+	err->address = address;
```
osandov: One last thing I noticed: `drgn_error_create()` never returns `NULL`, but it may return `&drgn_enomem`. This should be

```c
if (err != &drgn_enomem)
	err->address = address;
```

libdrgn/error.c
```diff
@@ -117,6 +120,37 @@ LIBDRGN_PUBLIC struct drgn_error *drgn_error_format(enum drgn_error_code code,
...
+	char *message;
+	int ret;
+
+	va_start(ap, format);
+	ret = vasprintf(&message, format, ap);
+	va_end(ap);
+	if (ret == -1)
+                return &drgn_enomem;
+	err = drgn_error_create_nodup(DRGN_ERROR_FAULT, message);
+	err->address = address;
```
osandov: Same thing here, `if (err != &drgn_enomem)`.

osandov:
I fixed an unused variable warning and switched `FaultError_init` to `PyArg_ParseTupleAndKeywords()`. I'll merge it as soon as tests pass, thanks!

sdimitro:
Cool, thanks Omar!

PR #38, opened 2020-01-24
BPF prog and maps helpers

rdna:
Add a couple of simple BPF helpers to iterate over BPF progs and maps and an example of usage.


osandov:
Thank you!

PR #40, opened 2020-01-29
examples/tcp_sock: Fix missed import

rdna:
In the previous commit used import was removed by mistake. Restore them.

Fixes: 7dd1853 ("examples: Switch tcp_sock to cgroup helpers")


osandov:
Huh, I could've sworn I tested it. Anyways, thanks!

PR #42, opened 2020-02-06
libdrgn: Add support for looking up reading kernel parameters

### LINKED ISSUE #36: Add helper for getting kernel CONFIG options
If the kernel was compiled with `CONFIG_IKCONFIG=y`, then the gzipped kernel configuration exists in kernel memory (see [here](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/kernel/configs.c?h=v5.5-rc5)). We can't access it currently:

```
>>> prog.variable('kernel_config_data')
LookupError: could not find variable 'kernel_config_data'
```

This is because `kernel_config_data` and `kernel_config_end` are defined in assembly. We'll need to add an interface for looking up ELF symbols by name.

jgkamat:
Since this is my first diff, I would appreciate it if you took a closer look than you normally do, to make sure I'm not overlooking anything.

This should be following the methodology set in #36 pretty exactly. Let me know if anything can be improved though!

I'm not sure if these tests will be fine, if CONFIG_IKCONFIG is not set (as is the case on most systems) I think the test will fail as is. Do you have a way to skip the test in this case, without preventing it from failing if the functionality is actually broken?

Closes #36

### DIFF
--- a/drgn/helpers/linux/kconfig.py
+++ b/drgn/helpers/linux/kconfig.py
@@ -1,0 +1,66 @@
+# SPDX-License-Identifier: GPL-3.0+
+
+"""
+Kconfig
+-------
+
+The ``drgn.helpers.linux.kconfig`` module provides helpers for reading the
+Linux kernel build configuration.
+"""
+
+import gzip
+import types
+from typing import Mapping
+
+__all__ = ("get_kconfig",)
+
+
+def get_kconfig(prog) -> Mapping[str, str]:
+    """
+    Get the kernel build configuration as a mapping from the option name to the
+    value.
+
+    >>> get_kconfig(prog)['CONFIG_SMP']
+    'y'
+    >>> get_kconfig(prog)['CONFIG_HZ']
+    '300'
+
+    This is only supported if the kernel was compiled with ``CONFIG_IKCONFIG``.
+    Note that most Linux distributions do not enable this option.
+    """
+    try:
+        return prog.cache["kconfig_map"]
+    except KeyError:
+        pass
+
+    try:
+        start = prog.symbol("kernel_config_data").address
+        size = prog.symbol("kernel_config_data_end").address - start
+    except LookupError:
+        # Before Linux kernel commit 13610aa908dc ("kernel/configs: use .incbin
+        # directive to embed config_data.gz") (in v5.1), the data is a variable
+        # rather than two symbols.
+        try:
+            kernel_config_data = prog["kernel_config_data"]
+        except KeyError:
+            raise LookupError(
+                "kernel configuration data not found; kernel must be compiled with CONFIG_IKCONFIG"
+            )
+        # The data is delimited by the magic strings "IKCFG_ST" and "IKCFG_ED"
+        # plus a NUL byte.
+        start = kernel_config_data.address_ + 8
+        size = len(kernel_config_data) - 17
+
+    data = prog.read(start, size)
+    result = {}
+    for line in gzip.decompress(data).decode().splitlines():
+        if not line or line.startswith("#"):
+            continue
+        name, _, value = line.partition("=")
+        if value:
+            result[name] = value
+
+    # Make result mapping 'immutable', so changes cannot propagate to the cache
+    result = types.MappingProxyType(result)
+    prog.cache["kconfig_map"] = result
+    return result
--- a/libdrgn/program.c
+++ b/libdrgn/program.c
@@ -961,6 +961,70 @@ drgn_program_find_symbol_by_address(struct drgn_program *prog, uint64_t address,
 	return NULL;
 }
 
+struct find_symbol_by_name_arg {
+	const char *name;
+	struct drgn_symbol **ret;
+	struct drgn_error *err;
+	bool bad_symtabs;
+};
+
+static int find_symbol_by_name_cb(Dwfl_Module *dwfl_module, void **userdatap,
+				  const char *module_name, Dwarf_Addr base,
+				  void *cb_arg)
+{
+	struct find_symbol_by_name_arg *arg = cb_arg;
+	int symtab_len, i;
+
+	symtab_len = dwfl_module_getsymtab(dwfl_module);
+	i = dwfl_module_getsymtab_first_global(dwfl_module);
+	if (symtab_len == -1 || i == -1) {
+		arg->bad_symtabs = true;
+		return DWARF_CB_OK;
+	}
+	for (; i < symtab_len; i++) {
+		GElf_Sym elf_sym;
+		GElf_Addr elf_addr;
+		const char *name;
+
+		name = dwfl_module_getsym_info(dwfl_module, i, &elf_sym,
+					       &elf_addr, NULL, NULL, NULL);
+		if (name && strcmp(arg->name, name) == 0) {
+			struct drgn_symbol *sym;
+
+			sym = malloc(sizeof(*sym));
+			if (sym) {
+				sym->name = name;
+				sym->address = elf_addr;
+				sym->size = elf_sym.st_size;
+				*arg->ret = sym;
+			} else {
+				arg->err = &drgn_enomem;
+			}
+			return DWARF_CB_ABORT;
+		}
+	}
+	return DWARF_CB_OK;
+}
+
+LIBDRGN_PUBLIC struct drgn_error *
+drgn_program_find_symbol_by_name(struct drgn_program *prog,
+			const char *name, struct drgn_symbol **ret)
+{
+	struct find_symbol_by_name_arg arg = {
+		.name = name,
+		.ret = ret,
+	};
+
+	if (prog->_dicache &&
+	    dwfl_getmodules(prog->_dicache->dindex.dwfl, find_symbol_by_name_cb,
+			    &arg, 0))
+		return arg.err;
+	return drgn_error_format(DRGN_ERROR_LOOKUP,
+				 "could not find symbol with name '%s'%s", name,
+				 arg.bad_symtabs ?
+				 " (could not get some symbol tables)" : "");
+}
+
 LIBDRGN_PUBLIC struct drgn_error *
 drgn_program_element_info(struct drgn_program *prog, struct drgn_type *type,
 			  struct drgn_element_info *ret)
--- a/libdrgn/program.c
+++ b/libdrgn/program.c
@@ -906,9 +906,10 @@ drgn_program_find_object(struct drgn_program *prog, const char *name,
 				      ret);
 }
 
-bool drgn_program_find_symbol_internal(struct drgn_program *prog,
-				       uint64_t address, Dwfl_Module *module,
-				       struct drgn_symbol *ret)
+bool drgn_program_find_symbol_by_address_internal(struct drgn_program *prog,
+						  uint64_t address,
+						  Dwfl_Module *module,
+						  struct drgn_symbol *ret)
 {
 	const char *name;
 	GElf_Off offset;
@@ -943,15 +944,16 @@ struct drgn_error *drgn_error_symbol_not_found(uint64_t address)
 }
 
 LIBDRGN_PUBLIC struct drgn_error *
-drgn_program_find_symbol(struct drgn_program *prog, uint64_t address,
-			 struct drgn_symbol **ret)
+drgn_program_find_symbol_by_address(struct drgn_program *prog, uint64_t address,
+				    struct drgn_symbol **ret)
 {
 	struct drgn_symbol *sym;
 
 	sym = malloc(sizeof(*sym));
 	if (!sym)
 		return &drgn_enomem;
-	if (!drgn_program_find_symbol_internal(prog, address, NULL, sym)) {
+	if (!drgn_program_find_symbol_by_address_internal(prog, address, NULL,
+							  sym)) {
 		free(sym);
 		return drgn_error_symbol_not_found(address);
 	}
--- a/libdrgn/python/program.c
+++ b/libdrgn/python/program.c
@@ -718,20 +718,27 @@ static StackTrace *Program_stack_trace(Program *self, PyObject *args,
 	return ret;
 }
 
-static PyObject *Program_symbol(Program *self, PyObject *args, PyObject *kwds)
+static PyObject *Program_symbol(Program *self, PyObject *arg)
 {
-	static char *keywords[] = {"address", NULL};
 	struct drgn_error *err;
-	struct index_arg address = {};
 	struct drgn_symbol *sym;
 	PyObject *ret;
 
-	if (!PyArg_ParseTupleAndKeywords(args, kwds, "O&", keywords,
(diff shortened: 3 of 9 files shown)


libdrgn/program.c
```diff
@@ -995,3 +1001,64 @@ drgn_program_member_info(struct drgn_program *prog, struct drgn_type *type,
...
+
+	for (int i = symtab_offset; i < symtab_len; i++) {
+		const char *name = dwfl_module_getsym_info(
+			dwfl_module, i, &elf_sym, &elf_addr, NULL, NULL, NULL);
+
+		if (strcmp(sym->name, name) == 0) {
+			sym->symbol->name = name;
+			sym->symbol->address = elf_addr;
+			sym->symbol->size = elf_sym.st_size;
+			return DWARF_CB_ABORT;
```
jgkamat: I don't like this method of signalling success on ABORT, but I'm not sure if there's another way to cleanly get the exit status from the caller. Would you prefer setting a field in `sym` to determine success instead?

libdrgn/program.c
```diff
@@ -995,3 +1001,64 @@ drgn_program_member_info(struct drgn_program *prog, struct drgn_type *type,
...
+	const char *module_name,
+	Dwarf_Addr base, void *arg)
+{
+	GElf_Sym elf_sym;
+	GElf_Addr elf_addr;
+	struct drgn_string_symbol *sym;
+	int symtab_offset, symtab_len;
+
+	sym = arg;
+	symtab_len = dwfl_module_getsymtab(dwfl_module);
```
jgkamat: I'm not sure if I should be handling errors for these `dwfl` calls in any way. Am I missing a check on these calls (and what would that look like)?

drgn/helpers/linux/kernel_config.py
```diff
@@ -0,0 +1,54 @@
...
+    except KeyError:
+        pass
+
+    try:
+        start = prog.symbol("kernel_config_data").address
+        end = prog.symbol("kernel_config_data_end").address
+    except LookupError as e:
+        print("Kernel configuration data not found. Ensure your kernel is "
+              "compiled with CONFIG_IKCONFIG set.", file=sys.stderr)
+        raise e
```
osandov: I consider the helpers library code, so rather than printing, this information should go in an exception:

```
except LookupError:
    raise LookupError("kernel configuration data not found; kernel must be compiled with CONFIG_IKCONFIG")
```

libdrgn/program.c
```diff
@@ -995,3 +1001,64 @@ drgn_program_member_info(struct drgn_program *prog, struct drgn_type *type,
...
+
+	err = drgn_program_get_dwfl(prog, &dwfl);
+	if (err)
+		return err;
+
+	if (dwfl_getmodules(dwfl, drgn_dwarf_module_iterate_sym_table, &arg, 0)) {
+		*ret = arg.symbol;
+		return NULL;
+	}
+	// If we did a full iteration, we didn't find anything.
```
osandov: The allocated symbol is being leaked in the not found case.

tests/helpers/linux/test_kernel_config.py
```diff
@@ -0,0 +1,15 @@
+import os
+import os.path
+import typing
+
+from drgn.helpers.linux import kernel_config
+
+from tests.helpers.linux import LinuxHelperTestCase
+
+
+class TestBlock(LinuxHelperTestCase):
```
osandov: `TestKconfig`?

tests/helpers/linux/test_kernel_config.py
```diff
@@ -0,0 +1,15 @@
...
+    def test_kernel_config(self):
+        # TODO skip this test when CONFIG_IKCONFIG is not set.
+        m = kernel_config.kconfig(self.prog)
+        assert isinstance(m, typing.Mapping)
+        assert m['CONFIG_IKCONFIG'] in {'y', 'm'}
```
osandov: `self.assertTrue()` and `self.assertIn()` to get nicer output on failure.

osandov:
As for the test case, let's only run it if `/proc/config.gz` exists. If not, the test can call `self.skipTest()`.

We could also base it on the existence of the `kernel_config_data` symbol (e.g., by looking at `/proc/kallsyms` or with `prog.symbol()`) to support the test on systems with `CONFIG_IKCONFIG` but not `CONFIG_IKCONFIG_PROC`. However, this has the major downside that if the symbol name changes or is removed for some reason, the test will start skipping silently rather than failing. In my opinion the stricter config requirement is the better option.

libdrgn/program.c
jgkamat: Thanks for catching this, it totally slipped my mind

jgkamat:
Seems like travis dosen't like the skipping that I implemented. It's possible I made a mistake, or that travis dosen't have CONFIG_IKCONFIG set but somehow has /proc/config.gz present (?). I'll take a look at that hopefully later today.

osandov:
I'll take a look at the code later today, but I think what's going on is that `CONFIG_IKCONFIG` works differently on older kernels -- see https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=13610aa908dcfce77135bb799c0a10d0172da6ba. We'll want the helper to support that, too.

libdrgn/program.c
```diff
@@ -995,3 +995,83 @@ drgn_program_member_info(struct drgn_program *prog, struct drgn_type *type,
...
+	sym = arg;
+	sym->err = NULL;
+
+	symtab_len = dwfl_module_getsymtab(dwfl_module);
+	symtab_offset = dwfl_module_getsymtab_first_global(dwfl_module);
+
+	if (symtab_len == -1 || symtab_offset == -1) {
+		sym->err = drgn_error_format(DRGN_ERROR_LOOKUP,
+			"could not look up symbol table for %s.", module_name);
+		return DWARF_CB_ABORT;
```
osandov: You know what, it's probably not a great idea to fail all symbol lookups hard because one file has a bad symbol table. Let's just skip the module in this case by returning `DWARF_CB_OK`.

libdrgn/program.c
```diff
@@ -995,3 +995,83 @@ drgn_program_member_info(struct drgn_program *prog, struct drgn_type *type,
...
+	}
+
+	for (int i = symtab_offset; i < symtab_len; i++) {
+		const char *name = dwfl_module_getsym_info(
+			dwfl_module, i, &elf_sym, &elf_addr, NULL, NULL, NULL);
+
+		if (strcmp(sym->name, name) == 0) {
```
osandov: `dwfl_module_getsym_info()` returns `NULL` on error, so let's make this `if (name && strcmp(sym->name, name) == 0)` (again, skipping bad data rather than failing hard).

libdrgn/program.c
```diff
@@ -995,3 +995,83 @@ drgn_program_member_info(struct drgn_program *prog, struct drgn_type *type,
...
+
+	struct drgn_symbol *sym;
+
+	err = drgn_program_get_dwfl(prog, &dwfl);
+	if (err)
+		return err;
+
+	sym = malloc(sizeof(*sym));
+	arg.name = name;
+	arg.symbol = sym;
```
osandov: Rather than always allocating the symbol upfront, let's allocate it only when we find it in the callback.

libdrgn/program.c
```diff
@@ -995,3 +995,83 @@ drgn_program_member_info(struct drgn_program *prog, struct drgn_type *type,
...
+drgn_program_find_symbol_by_name(struct drgn_program *prog,
+			const char *name, struct drgn_symbol **ret)
+{
+	struct drgn_string_symbol arg;
+	struct drgn_error *err = NULL;
+	Dwfl *dwfl;
+
+	struct drgn_symbol *sym;
+
+	err = drgn_program_get_dwfl(prog, &dwfl);
```
osandov: If the program doesn't have a `Dwfl`, then there must be no symbols, so we can fail fast.

libdrgn/python/util.c
```diff
@@ -103,6 +103,26 @@ int index_converter(PyObject *o, void *p)
 	}
 }
 
+int string_int_converter(PyObject *o, void *p)
```
osandov: Until we need this in more places, I'd rather fold it into `Program_symbol()`.

drgn/helpers/linux/kernel_config.py
```diff
@@ -0,0 +1,52 @@
+# SPDX-License-Identifier: GPL-3.0+
+
+"""
+Kconfig
+------
+
+
+The ``drgn.helpers.linux.kconfig`` module provides helpers for reading kernel
```
osandov: This says the module is `kconfig` but the file is called `kernel_config.py`. Let's stick with `kconfig.py`.

drgn/helpers/linux/kernel_config.py
```diff
@@ -0,0 +1,52 @@
...
+configuration from memory. This is only supported if CONFIG_IKCONFIG is set.
+"""
+
+
+import sys
```
osandov: This is unused now.

drgn/helpers/linux/kernel_config.py
```diff
@@ -0,0 +1,52 @@
...
+import gzip
+import types
+
+from drgn import NULL, cast, container_of
```
osandov: Ditto.

libdrgn/program.c
jgkamat: Isn't this fail-fast behavior happening already below? Or is there another way errors are reported from this function?

libdrgn/python/util.c
jgkamat: I think it would be a little more messy to have this directly inline in `Program_symbol` (and it would make `Program_symbol` rather large/complex. I decided to move the function into program.c but keep it in it's own function (rather than inline) to make it easier to refactor out if needed. Does that sound fine?

libdrgn/program.c
osandov: Sorry, I mean something like what `drgn_program_find_symbol_by_address_internal()` does: if there is no `prog->_dicache`, immediately return the not found error instead of going through the trouble of creating the `dicache`/`dindex`/`dwfl`.

libdrgn/python/util.c
osandov: Hm it's only a couple more lines of code, right?

```c
if (PyUnicode_Check(arg)) {
	const char *name;

	name = PyUnicode_AsUTF8(arg);
	if (!name)
		return NULL;
	err = drgn_program_find_symbol_by_name(&self->prog, name, &sym);
} else {
	struct index_arg address = {};

	if (!index_converter(arg, &address))
		return NULL;
	err = drgn_program_find_symbol_by_address(&self->prog,
						  address.uvalue, &sym);
}
```

`Program_symbol()` is not doing much other than interpreting the argument and passing it through to `libdrgn`.

libdrgn/python/program.c
```diff
@@ -718,19 +718,24 @@ static StackTrace *Program_stack_trace(Program *self, PyObject *args,
 	return ret;
 }
 
-static PyObject *Program_symbol(Program *self, PyObject *args, PyObject *kwds)
+static PyObject *Program_symbol(Program *self, PyObject *address_or_name)
 {
 	static char *keywords[] = {"address", NULL};
```
osandov: This is unused now.

libdrgn/program.c
```diff
@@ -995,3 +995,86 @@ drgn_program_member_info(struct drgn_program *prog, struct drgn_type *type,
...
+
+	sym = arg;
+	sym->err = NULL;
+
+	symtab_len = dwfl_module_getsymtab(dwfl_module);
+	symtab_offset = dwfl_module_getsymtab_first_global(dwfl_module);
+
+	if (symtab_len == -1 || symtab_offset == -1) {
+		sym->err = drgn_error_format(DRGN_ERROR_LOOKUP,
+			"could not look up symbol table for %s.", module_name);
```
osandov: If there are multiple modules with bad symbol tables, this leaks the previous error. I think it would suffice to have a flag that indicates that some symbol tables couldn't be read.

libdrgn/program.c
```diff
@@ -995,3 +995,86 @@ drgn_program_member_info(struct drgn_program *prog, struct drgn_type *type,
...
+		return DWARF_CB_OK;
+	}
+
+	for (int i = symtab_offset; i < symtab_len; i++) {
+		const char *name = dwfl_module_getsym_info(
+			dwfl_module, i, &elf_sym, &elf_addr, NULL, NULL, NULL);
+
+		if (name && strcmp(sym->name, name) == 0) {
+			// Leaked to caller
+			sym->symbol = malloc(sizeof(struct drgn_symbol));
```
osandov: This can be even simpler if we pass the `struct drgn_symbol **ret` through the callback.

drgn/helpers/linux/kconfig.py
```diff
@@ -0,0 +1,59 @@
...
+        return prog.cache["kconfig_options_map"]
+    except KeyError:
+        pass
+
+    try:
+        start = prog.symbol("kernel_config_data").address
+        size = prog.symbol("kernel_config_data_end").address - start
+        raw_data = prog.read(start, size)
+    except LookupError:
+        try:
```
osandov: We should limit the `try` block to only contain the statement that we want to handle, i.e., `prog["kernel_config_data"]`.

drgn/helpers/linux/kconfig.py
```diff
@@ -0,0 +1,59 @@
...
+            # Try reading data in format present before kernel 5.0.
+            start = prog['kernel_config_data'].address_
+            size = len(prog['kernel_config_data'])
+            raw_data = prog.read(start, size)
+            trim_left = raw_data.find(b'IKCFG_ST') + len(b'IKCFG_ST')
+            trim_right = raw_data.rfind(b'IKCFG_ED')
+            if trim_left == -1 or trim_right == -1:
```
osandov: I don't think we need to actually find the delimiters; they're guaranteed to be at the start and end of the array.

drgn/helpers/linux/kconfig.py
```diff
@@ -0,0 +1,59 @@
...
+                raise KeyError("IKCFG not Found")
+            raw_data = raw_data[trim_left:trim_right]
+        except KeyError:
+            raise LookupError("kernel configuration data not found; kernel must be compiled with CONFIG_IKCONFIG")
+
+    raw_options = gzip.decompress(raw_data).decode().split('\n')
```
osandov: `splitlines()`

drgn/helpers/linux/kconfig.py
```diff
@@ -0,0 +1,59 @@
...
+
+    result = {}
+
+    for line in raw_options:
+        if not line or line.startswith('#'):
+            continue
+        split_option = line.split('=')
```
osandov: This is slightly cleaner using `line.partition('=')`.

Issue #24, opened 2019-12-04
Support stack traces for kdump files


osandov:
drgn uses the `NT_PRSTATUS` ELF note in core dumps to get the initial registers for stack unwinding. (I think?) kdump files include this note as well, but libkdumpfile doesn't have an interface to get it. It should suffice to add `kdump_prstatus_raw()` to libkdumpfile similar to `kdump_vmcoreinfo_raw()`, although we might want a generic `kdump_note_raw()` instead.

sdimitro:
As mentioned in the meeting I can pick this up!

sdimitro:
This can be closed as https://github.com/osandov/drgn/pull/45 has been merged

PR #44, opened 2020-02-12
Fix leak in kdump code

sdimitro:
prog->kdump_ctx is never really initialized, and the kdump_ctx struct
allocated in drgn_program_set_kdump() is leaked.

WARNING:
I found this bug while working towards https://github.com/osandov/drgn/issues/24 . Once I applied the changes on this PR and tested my bits, `drgn` would hang on exit. That was due to a bug in libkdumpfile as I found later. I submitted a PR for the latter bug here -> https://github.com/ptesarik/libkdumpfile/pull/26
Please do not merge this until https://github.com/ptesarik/libkdumpfile/pull/26 is merged.


sdimitro:
@osandov This should be good to go now

PR #46, opened 2020-02-20
libdrgn: Autodetect language type from compile units

jgkamat:
I was originally expecting this bit to be rather simple, but it ended up becoming much more complicated due to several issues:

1. A program can have multiple compilation units, which have different languages. 
2. Some methods are at "program level" which need to have some assumed language associated with them.
3. Many scenarios (objects created from python, mocks in existing tests, etc) all don't have a real "program" to autodetect the language from. These cases all needed to fall back exactly like before (currently at least).

The solution I have so far is: 

1. Store a language variable in each type, generated on type creation in `from_dwarf` methods. This variable is initalized to c by default and is not set if we cannot find a language
2. Store a language variable in the drgn_program, which is used in the couple places we don't have a type to work from. This is initilaized after the debug information is loaded , and is not set otherwise. The getter for this function returns c if it is unset.

This is pretty rough (sorry about that), but this is the best way I can think of how to do this at the moment. Let me know if you find any way this can be improved though!

I currently don't like:

1. Setting the language on debug information load. I'm not sure if this is the "correct" place to do that.
2. Setting the language in the from_dwarf methods. It feels a little out-of-place, but I can't think of a better place to have it.
3. (currently) no way to check if there was an error deducing types.
4. (currently) no fallback if `main` is not found in the top-level program to find a "default compile unit". I'm not sure (yet) how to pick a random/first one (and I'm not sure if that's a good idea).
5. (currently) We cannot override the language passed to types (although it should be easy to add). I could add an argument to these initializer functions, but that would require updates to all the callers (which seem like a lot). Is that worth it?
6. Most of the tests use mocks which cause the language detection to fail. It would be nice to have them run this code, even though it's not the purpose of those tests.

The big missing part is tests currently. I'll try to write some as soon as the high level structure is a little more finalized.

I'm going to leave the actual C++ switch for another commit/review cycle because this is very complicated/large as is.


libdrgn/dwarf_info_cache.c
```diff
@@ -628,6 +628,7 @@ drgn_compound_type_from_dwarf(struct drgn_dwarf_info_cache *dicache,
 		*ret = type;
 		return NULL;
 	}
+	drgn_update_language_from_dwarf(type, die);
```
osandov: Even though it will cause churn, I think it'd be cleaner to pass the language to all of the initializers. This way, we're explicit about the places that we're setting a default language for now. Plus, drgn_types should be immutable, and having a separate setter for the language makes that less obvious.

So, let's add `const struct drgn_language *lang` as the last argument to all of the initializers, make use of it in the DWARF code, and explicitly pass in `&drgn_language_c` where we don't yet have that information.

libdrgn/language.h
```diff
@@ -155,6 +156,21 @@ drgn_element_format_object_flags(enum drgn_format_object_flags flags)
...
 }
 
+/** Return the drgn language spec corresponding to the given dwfl language */
+static const struct drgn_language *drgn_dwarf_to_language(int dwfl_lang)
+{
+	switch (dwfl_lang) {
+	case DW_LANG_C:
+	case DW_LANG_C89:
+	case DW_LANG_C99:
+	case DW_LANG_C11:
```
osandov: I hate that the DWARF spec makes this not future proof by design, but oh well.

libdrgn/language.h
```diff
@@ -155,6 +156,21 @@ drgn_element_format_object_flags(enum drgn_format_object_flags flags)
...
+		return &drgn_language_c;
+	// TODO Provide proper fallback errors instead of defaulting to C always.
```
osandov: I don't think we'll ever want to make this fail hard. I'd rather have a partially functional debugger for an unknown language than one that doesn't start up at all. Long term, we probably want a `drgn_language_unknown` that has sane default operations, but for now C is fine.

libdrgn/object.c
```diff
@@ -837,7 +837,7 @@ LIBDRGN_PUBLIC struct drgn_error *
 drgn_format_object(const struct drgn_object *obj, size_t columns,
 		   enum drgn_format_object_flags flags, char **ret)
 {
-	const struct drgn_language *lang = &drgn_language_c;
+	const struct drgn_language *lang = drgn_type_lang(obj->type);
```
osandov: This is repeated often enough that I think it merits an inline `drgn_object_lang()` helper in `drgn.h`.

libdrgn/object.c
```diff
@@ -1075,7 +1075,7 @@ drgn_object_cast(struct drgn_object *res,
 		 struct drgn_qualified_type qualified_type,
 		 const struct drgn_object *obj)
 {
-	const struct drgn_language *lang = &drgn_language_c;
+	const struct drgn_language *lang = drgn_type_lang(obj->type);
```
osandov: For casting, do we want to use the language of the old object or the language of the new type? I think the latter.

libdrgn/object.c
```diff
@@ -1168,7 +1168,8 @@ LIBDRGN_PUBLIC struct drgn_error *drgn_object_cmp(const struct drgn_object *lhs,
 						  const struct drgn_object *rhs,
 						  int *ret)
 {
-	const struct drgn_language *lang = &drgn_language_c;
+	// TODO handle the case where LHS and RHS have different types
+	const struct drgn_language *lang = drgn_type_lang(lhs->type);
```
osandov: I think this is pretty much the sanest way to handle it, unless you have any better ideas.

libdrgn/program.c
```diff
@@ -50,6 +50,27 @@ drgn_program_platform(struct drgn_program *prog)
...
+	err = drgn_program_find_object(prog, "main", NULL,
+				       DRGN_FIND_OBJECT_ANY, &res);
+	if (!err) {
+		prog->lang = drgn_type_lang(res.type);
+		return;
+	}
+
+	// We could not deduce type or find result from "main".
+	// TODO have a better fallback here.
+	prog->lang = &drgn_language_c;
```
osandov: Maybe we can just leave it unset if we couldn't find main, since `drgn_program_get_lang()` already has the concept of a fallback? Also, `drgn_object_deinit()` should be called here, and the error needs to be handled/destroyed. Something like

```c
err = drgn_program_find_object(...);
if (err) {
	if (err->code != DRGN_ERROR_LOOKUP)
		return err;
	/* We couldn't find "main". Don't set the language. */
	drgn_error_destroy(err);
} else {
	prog->lang = drgn_type_lang(res.type);
}
drgn_object_deinit(&res);
return NULL;
```

libdrgn/program.c
```diff
@@ -529,6 +550,14 @@ struct drgn_error *drgn_program_get_dwfl(struct drgn_program *prog, Dwfl **ret)
 	return NULL;
 }
 
+const struct drgn_language *drgn_program_get_lang(struct drgn_program *prog) {
+	if (prog->lang) {
+		return prog->lang;
+	}
+	// TODO handle fallback better.
```
osandov: Same as `drgn_dwarf_to_language()`, I think for now falling back to C is fine, and in the future we can have a default unknown language.

libdrgn/program.c
```diff
@@ -50,6 +50,27 @@ drgn_program_platform(struct drgn_program *prog)
 	return prog->has_platform ? &prog->platform : NULL;
 }
 
+void drgn_program_init_language(struct drgn_program *prog)
```
osandov: This is called every time we load debug info, so it needs an `if (prog->lang) return;`.

Also, this can be static.

libdrgn/program.c
```diff
@@ -529,6 +550,14 @@ struct drgn_error *drgn_program_get_dwfl(struct drgn_program *prog, Dwfl **ret)
 	return NULL;
 }
 
+const struct drgn_language *drgn_program_get_lang(struct drgn_program *prog) {
+	if (prog->lang) {
```
osandov: Coding style nit: no curly braces around a single statement.

libdrgn/program.c
```diff
@@ -50,6 +50,27 @@ drgn_program_platform(struct drgn_program *prog)
...
+void drgn_program_init_language(struct drgn_program *prog)
+{
+	struct drgn_error *err;
+	struct drgn_qualified_type main_type;
+	struct drgn_object res;
+
+	drgn_object_init(&res, prog);
+
+	// Use the CU from the "main" object as the lang for the whole program.
+	err = drgn_program_find_object(prog, "main", NULL,
```
jgkamat: @osandov Actually, one more thing I forgot to ask. I *think* as it is right now, this line will populate the dwarf info cache which (I think) is currently meant to be loaded lazily. If that's the case, I could try to load this lazily as well. What do you think?

osandov: At this point, the debugging information has already been loaded and the DWARF index populated, so it's fair game to look things up.

jgkamat:
I'm still working on tests for this, but the base infrastructure should mostly be there. It might need a little more cleanup though, so I'll give this another look-over myself once I'm done with tests (and then I'll remove the WIP tag).

libdrgn/dwarf_info_cache.c
```diff
@@ -976,7 +980,8 @@ drgn_typedef_type_from_dwarf(struct drgn_dwarf_info_cache *dicache,
 		return err;
 	}
 
-	drgn_typedef_type_init(type, name, aliased_type);
+	drgn_typedef_type_init(type, name, aliased_type, lang);
+
```
osandov: Stray whitespace change.

libdrgn/language.h
```diff
@@ -198,6 +200,42 @@ drgn_element_format_object_flags(enum drgn_format_object_flags flags)
...
+		return &drgn_default_language;
+	}
+}
+
+/**
+ * Deduce language which corresponds with the CU of the given dwarf die.
+ *
+ * @param[in] die Dwarf_Die used to retrieve the current language.
+ */
+static const struct drgn_language *drgn_language_from_dwarf(Dwarf_Die *die)
```
osandov: This is only used in `dwarf_info_cache.c`, correct? Let's just move it there.

Also, naming-wise, `drgn_language_from_die` would be a little clearer.

libdrgn/language.h
```diff
@@ -198,6 +200,42 @@ drgn_element_format_object_flags(enum drgn_format_object_flags flags)
...
+{
+	Dwarf_Die cudie;
+	if (dwarf_cu_info(die->cu, NULL, NULL, &cudie, NULL, NULL, NULL, NULL) == 0) {
+		// TODO don't fail silently on failure
```
osandov: Yeah if we can't get the CU info, something's pretty busted, so this should return a `struct drgn_error *` (and return the language via a return parameter).

libdrgn/language.h
```diff
@@ -198,6 +200,42 @@ drgn_element_format_object_flags(enum drgn_format_object_flags flags)
 		(flags & DRGN_FORMAT_OBJECT_ELEMENT_TYPE_NAMES) >> 2);
 }
 
+/** Return the drgn language spec corresponding to the given dwfl language */
+static const struct drgn_language *drgn_dwarf_to_language(int dwfl_lang)
```
osandov: `drgn_language_from_dw_lang` would be a more descriptive name here.

libdrgn/language.h
```diff
@@ -198,6 +200,42 @@ drgn_element_format_object_flags(enum drgn_format_object_flags flags)
...
+{
+	Dwarf_Die cudie;
+	if (dwarf_cu_info(die->cu, NULL, NULL, &cudie, NULL, NULL, NULL, NULL) == 0) {
+		// TODO don't fail silently on failure
+		return drgn_dwarf_to_language(dwarf_srclang(&cudie));
+	}
+	return &drgn_default_language;
+}
+
+static inline const struct drgn_language *drgn_language_default(
```
osandov: Maybe `drgn_language_or_default`?

Style-wise, I'd write it as

```c
static inline const struct drgn_language *
drgn_language_or_default(const struct drgn_language *lang)
{
	return lang ? lang : &drgn_default_language;
}
```

libdrgn/object.c
```diff
@@ -1075,7 +1075,7 @@ drgn_object_cast(struct drgn_object *res,
 		 struct drgn_qualified_type qualified_type,
 		 const struct drgn_object *obj)
 {
-	const struct drgn_language *lang = &drgn_language_c;
+	const struct drgn_language *lang = drgn_object_lang(obj);
```
osandov: I believe we want `drgn_type_lang(qualified_type.type)` here, i.e., the language of the type we're casting to.

libdrgn/object.c
```diff
@@ -1135,31 +1135,31 @@ drgn_object_reinterpret(struct drgn_object *res,
 LIBDRGN_PUBLIC struct drgn_error *
 drgn_object_integer_literal(struct drgn_object *res, uint64_t uvalue)
 {
-	const struct drgn_language *lang = &drgn_language_c;
+	const struct drgn_language *lang = drgn_object_lang(res);
```
osandov: `drgn_object_*_literal()` don't operate on the existing object but rather replace it, so this should be `drgn_program_lang(obj->prog)` here and in the other literal helpers.

libdrgn/type_index.c
```diff
@@ -161,49 +161,49 @@ static void default_primitive_types_init(void)
 
 	drgn_int_type_init(&default_primitive_types[DRGN_C_TYPE_CHAR],
 			   drgn_primitive_type_spellings[DRGN_C_TYPE_CHAR][0],
-			   1, true);
+			   1, true, NULL);
```
osandov: These primitive types are all for C, so let's explicitly pass `&drgn_language_c`.

libdrgn/type_index.c
```diff
@@ -343,7 +343,7 @@ drgn_type_index_find_primitive(struct drgn_type_index *tindex,
 					&tindex->default_size_t :
 					&tindex->default_ptrdiff_t);
 				drgn_typedef_type_init(*ret, spellings[0],
-						       qualified_type);
+						       qualified_type, NULL);
```
osandov: Same, this is specifically for C.

libdrgn/type_index.c
```diff
@@ -411,7 +411,7 @@ drgn_type_index_pointer_type(struct drgn_type_index *tindex,
 	type = malloc(sizeof(*type));
 	if (!type)
 		return &drgn_enomem;
-	drgn_pointer_type_init(type, tindex->word_size, referenced_type);
+	drgn_pointer_type_init(type, tindex->word_size, referenced_type, NULL);
```
osandov: This should use the same language as the referenced type.

libdrgn/type_index.c
```diff
@@ -448,7 +448,7 @@ drgn_type_index_array_type(struct drgn_type_index *tindex, uint64_t length,
 	type = malloc(sizeof(*type));
 	if (!type)
 		return &drgn_enomem;
-	drgn_array_type_init(type, length, element_type);
+	drgn_array_type_init(type, length, element_type, NULL);
```
osandov: Likewise, same language as `element_type`.

libdrgn/type_index.c
```diff
@@ -484,7 +484,7 @@ drgn_type_index_incomplete_array_type(struct drgn_type_index *tindex,
 	type = malloc(sizeof(*type));
 	if (!type)
 		return &drgn_enomem;
-	drgn_array_type_init_incomplete(type, element_type);
+	drgn_array_type_init_incomplete(type, element_type, NULL);
```
osandov: Ditto.

libdrgn/python/program.c
```diff
@@ -839,6 +839,17 @@ static PyObject *Program_get_platform(Program *self, void *arg)
...
 
+static PyObject *Program_get_language(Program *self)
+{
+	const struct drgn_language *lang;
+
+	lang = drgn_program_lang(&self->prog);
+	if (lang)
+		return Language_wrap(lang);
+	else
+		Py_RETURN_NONE;
```
osandov: This won't ever be true since we fallback to the default language, correct? Which is the behavior I think we want.

libdrgn/python/type.c
```diff
@@ -404,6 +404,17 @@ static PyObject *DrgnType_get_is_variadic(DrgnType *self)
...
 
+static PyObject *DrgnType_get_language(DrgnType *self)
+{
+	const struct drgn_language *lang;
+
+	lang = drgn_type_lang(self->type);
+	if (lang)
+		return Language_wrap(lang);
+	else
+		Py_RETURN_NONE;
```
osandov: Same here, we should never have a `NULL` language for a type.

libdrgn/python/program.c
jgkamat: I think this shouldn't ever happen, but I wanted to have this fallback just in case there was some bug. But I suppose failing hard is better so we catch issues early. I'll remove this.

tests/test_language.py
```diff
@@ -0,0 +1,66 @@
+import unittest
```
jgkamat: This file is a little trivial, but it (afaik) executes all the main paths through the type-detection code, even though the result is the same.

libdrgn/python/language.c
```diff
@@ -0,0 +1,64 @@
...
+	.tp_doc = drgn_Language_DOC,
+	.tp_getset = Language_getset,
+};
+
+static PyObject *languages_py[DRGN_NUM_LANGUAGES];
+
+PyObject *Language_wrap(const struct drgn_language *language)
+{
+	PyObject *obj = languages_py[language - drgn_languages];
+	Py_INCREF(obj);
```
jgkamat: Heads up, I changed this from your commit. Without this, I was getting crashes in one of the tests I added (but I wasn't able to repro in normal debugging). Increasing the ref once when adding it to the array below didn't seem to help.

I'm not sure if this is correct though, I'm still not sure how lifecycle is handled (especially on init and handing out like this).

libdrgn/dwarf_info_cache.c
```diff
@@ -88,6 +88,26 @@ static int dwarf_flag(Dwarf_Die *die, unsigned int name, bool *ret)
...
+{
+	Dwarf_Die cudie;
+	if (dwarf_cu_info(die->cu, NULL, NULL, &cudie, NULL, NULL, NULL, NULL) == 0) {
+		// TODO don't fail silently on failure
+		*lang = drgn_language_from_dw(dwarf_srclang(&cudie));
+		return NULL;
+	}
+	*lang = &drgn_default_language;
+	return drgn_error_format(DRGN_ERROR_LOOKUP,
+				 "Binary is missing CU information.");
```
osandov: This is a fatal error, so we don't need to return anything in `*lang`. Also, we should return the error from libdw with `drgn_error_libdw()`.

libdrgn/python/program.c
```diff
@@ -839,6 +839,13 @@ static PyObject *Program_get_platform(Program *self, void *arg)
 		Py_RETURN_NONE;
 }
 
+static PyObject *Program_get_language(Program *self)
+{
+	const struct drgn_language *lang;
+	lang = drgn_program_lang(&self->prog);
+	return Language_wrap(lang);
```
osandov: This can simply be `return Language_wrap(drgn_program_lang(&self->prog));` now.

libdrgn/python/type.c
```diff
@@ -404,6 +404,13 @@ static PyObject *DrgnType_get_is_variadic(DrgnType *self)
 	return PyBool_FromLong(drgn_type_is_variadic(self->type));
 }
 
+static PyObject *DrgnType_get_language(DrgnType *self)
+{
+	const struct drgn_language *lang;
+	lang = drgn_type_lang(self->type);
+	return Language_wrap(lang);
```
osandov: `return Language_wrap(drgn_type_lang(self->type));`

docs/api_reference.rst
```diff
@@ -35,6 +35,13 @@ Programs
 
         :vartype: Platform or None
 
+    .. attribute:: language
+
+        Primary language that this program uses, or ``None`` if it has not
+        been determined yet.
+
+        :vartype: Language or None
```
osandov: Documentation needs updating since this can no longer be `None`. (Note that the work I just pushed moved this documentation to `_drgn.pyi`).

docs/api_reference.rst
```diff
@@ -1172,6 +1179,13 @@ Types
 
         :vartype: bool
 
+    .. attribute:: language
+
+        Language that this type is created from, or ``None`` if it has not
+        been determined yet.
+
+        :vartype: Language or None
```
osandov: Ditto.

libdrgn/python/language.c
osandov: Yup this is correct, good catch. Whoever gets this is going to drop a reference when it's done with it, so we need to account for that. This fix should be folded into my commit.

_drgn.pyi
```diff
@@ -495,6 +498,19 @@ class Register:
...
+    """
+    A ``Language`` represents a programming language supported by drgn.
+
+    This class cannot be constructed; there are singletons for the supported
+    languages.
+    """
+    name: str
+    """Name of the programming language."""
+
+    C: "Language"
```
jgkamat: Sorry for the (continued) issues, but this dosen't seem to be propogated through properly (the docstring at least).

From the shell:

```
class Language(builtins.object)
 |  A ``Language`` represents a programming language supported by drgn.
 |  
 |  This class cannot be constructed; there are singletons for the supported
 |  languages.
 |  
 |  Methods defined here:
 |  
 |  __repr__(self, /)
 |      Return repr(self).
 |  
 |  ----------------------------------------------------------------------
 |  Data descriptors defined here:
 |  
 |  name
 |      str
 |      
 |      Name of the programming language.
 |  
 |  ----------------------------------------------------------------------
 |  Data and other attributes defined here:
 |  
 |  C = Language.C
```

Any idea why this might be? If not, I can try to triage how the help text is propogated tomorrow :)

osandov:
That's expected. Python doesn't have a way to add docstrings to attributes in general. `Language.name` works because it comes from a descriptor in the C extension, but `Language.C` is populated by the equivalent of `Language.C = Language(...)`, so there's no way to give the runtime the docstring.

This looks good overall. I'm going to add a `language` argument to all of the type constructors in the Python bindings and clean up some style nits before merging.

osandov:
Just pushed some changes:

* I noticed that we should probably have a void_type per language, so I added that (which required some refactoring)
* I added a language parameter to the type constructors in the Python bindings
* I reorganized the series so all of the prep work is first, then a commit adding the language to drgn_type, then a commit adding the language to drgn_program, with the Python bindings and tests folded into those commits
* I changed the heuristic based on the language of "main" to only apply to userspace programs and set the kernel to always be C.

Sorry about all of the shuffling, normally I would've asked you to do it but I thought it'd be a quick style cleanup and one thing led to another. You did the hard part, so the main commits are still attributed to you.

Would you mind taking a look that everything still looks sane?

osandov:
Ignore that Travis failure, I think I confused it by pushing to my repo rather than yours at first.

libdrgn/kdump.c
```diff
@@ -129,6 +129,8 @@ struct drgn_error *drgn_program_set_kdump(struct drgn_program *prog)
 					     prog);
 	if (err)
 		goto err;
+	if (!prog->lang)
+		prog->lang = &drgn_language_c;
```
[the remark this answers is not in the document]
osandov: My thinking here is that in the future, we'll have an interface for manually setting the program language, which should always take precedence over what we determine automatically.

PR #50, opened 2020-03-03
Add cpp language stub

jgkamat:
This is a pretty small and mostly independent change, but I wanted to make sure this is ironed out before moving forward, as there are a couple rough edges I wanted to discuss.

After this, I want to start by tackling the problem of storing template types in `struct drgn_type`, and working on the template support itself (which will probably be another relatively large change).

### DIFF
--- a/libdrgn/program.c
+++ b/libdrgn/program.c
@@ -544,29 +544,6 @@ struct drgn_error *drgn_program_get_dwfl(struct drgn_program *prog, Dwfl **ret)
 	return NULL;
 }
 
-/* Set the default language from the language of "main". */
-static struct drgn_error *
-drgn_program_set_language_from_main(struct drgn_program *prog)
-{
-	struct drgn_error *err;
-	struct drgn_object res;
-
-	drgn_object_init(&res, prog);
-	err = drgn_program_find_object(prog, "main", NULL, DRGN_FIND_OBJECT_ANY,
-				       &res);
-	if (err) {
-		if (err->code == DRGN_ERROR_LOOKUP) {
-			/* We couldn't find "main". Don't set the language. */
-			drgn_error_destroy(err);
-			err = NULL;
-		}
-	} else {
-		prog->lang = drgn_type_language(res.type);
-	}
-	drgn_object_deinit(&res);
-	return err;
-}
-
 static struct drgn_error *
 userspace_report_debug_info(struct drgn_program *prog,
 			    struct drgn_dwarf_index *dindex,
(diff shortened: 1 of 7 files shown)


libdrgn/program.c
```diff
@@ -673,14 +669,19 @@ drgn_program_load_debug_info(struct drgn_program *prog, const char **paths,
...
+	report_from_dwfl = (!is_kernel && load_main);
 	err = drgn_dwarf_index_report_end(dindex, report_from_dwfl);
 	if ((!err || err->code == DRGN_ERROR_MISSING_DEBUG_INFO) &&
 	    !prog->has_platform) {
 		dwfl_getdwarf(prog->_dicache->dindex.dwfl,
 			      drgn_set_platform_from_dwarf, prog, 0);
 	}
+
+	if (!is_kernel && !prog->lang) {
+		err = drgn_program_set_language_from_main(prog);
```
jgkamat: I think there was a bug where this was running too soon (even if main was available, it would not be found). Moving the invocation here fixes it, but I'm not sure if I kept the semantics of the original call.

tests/test_dwarf.py
```diff
@@ -2059,3 +2059,13 @@ def test_language(self):
         self.assertEqual(
             dwarf_program(dies, lang=DW_LANG.BLISS).language, DEFAULT_LANGUAGE
         )
+        self.assertEqual(
+            dwarf_program(dies, lang=DW_LANG.C_plus_plus_14).language,
+            Language.CPP
+        )
+
+        tmp = dwarf_program(dies, lang=DW_LANG.C_plus_plus)
```
jgkamat: Removing this temp variable causes a segfault, and I'm not sure why. Could that be another bug? I'm not sure where to begin looking for that. It seems like the parent object must be deleted while it's children is alive.

libdrgn/language_cpp.c
```diff
@@ -0,0 +1,9 @@
+// Copyright 2018-2020 - Omar Sandoval
+// SPDX-License-Identifier: GPL-3.0+
+
+/*
+ * This file provides functions derived from the language_c.c file.
+ *
+ * Whenever cpp and c share the same functionality in a function, the c
+ * definition is used directly. Otherwise, it will be wrapped here.
```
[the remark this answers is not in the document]
osandov: I see two approaches. One is to add a `struct drgn_language *` parameter to all of the callbacks, reuse the callbacks for both C and C++, and have them change behavior as necessary. Something like

`language_c.c`:
```c
struct drgn_error *
c_family_format_type_name(const struct drgn_language *lang,
			  struct drgn_qualified_type qualified_type,
			  char **ret)
{
	...
	if (lang == &drgn_language_cpp)
		do C++ thing
	else
		do C thing
	...
}
```

`language.c`:
```c
const struct drgn_language drgn_languages[] = {
	[DRGN_LANGUAGE_C] = { .format_type_name = c_family_format_type_name },
	[DRGN_LANGUAGE_CPP] = { .format_type_name = c_family_format_type_name },
};
```

Another option is a shared implementation that takes a `bool cpp`, with the actual callbacks being wrappers around the shared implementation. Something along the lines of

`language_c.c`:
```c
struct drgn_error *
c_family_format_type_name(struct drgn_qualified_type qualified_type, bool cpp,
			  char **ret)
{
	...
	if (cpp)
		do C++ thing
	else
		do C thing
	...
}

struct drgn_error *c_format_type_name(struct drgn_qualified_type qualified_type,
				      char **ret)
{
	return c_family_format_type_name(qualified_type, false, ret);
}
```

`language_cpp.c`:
```c
struct drgn_error *
cpp_format_type_name(struct drgn_qualified_type qualified_type, char **ret)
{
	return c_family_format_type_name(qualified_type, true, ret);
}
```

`language.c`:
```c
const struct drgn_language drgn_languages[] = {
	[DRGN_LANGUAGE_C] = { .format_type_name = c_format_type_name },
	[DRGN_LANGUAGE_CPP] = { .format_type_name = cpp_format_type_name },
};
```

The first approach is more flexible and future-proof. The second approach is more explicit and easier to understand. What do you think?

tests/test_dwarf.py
osandov: https://github.com/osandov/drgn/commit/79f973007b2df7ecaacca40312a1b46befa7d53a

libdrgn/program.c
jgkamat: Yup, this seems to work, thanks so much for fixing that! I'll rebase minus my first commit on that branch.

libdrgn/language_cpp.c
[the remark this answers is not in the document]
jgkamat: I think both of these solutions are fine. I like the first one personally, as it makes it very easy to support another c-based language, but I agree the second is more explicit. 

The main downside of the second method is we would need to define many different boilerplate wrappers, which I think could get very repetitive (although we could make them with a macro or similar). 

The main downside of the first is that the language definitions for c/c++ will look almost identical, but I think that can be explained with more detailed comments above the language definition. If you like it, I can implement 1 in the future! :)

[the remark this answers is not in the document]
osandov: Okay, let's go with the first approach. I'll merge the first two commits now, and we can add this file if/when we have something to put in it.

PR #54, opened 2020-04-01
Migrate inline allocation to variable length allocation in drgn_type

jgkamat:
In anticipation of adding template type pointers in an array to `drgn_type`, I wanted  a way to add additional arrays to this structure. However, currently all variable length arrays are stored by reallocing the drgn_type object to be larger and appended to the object. This works for a single array, but for multiple arrays it won't. 

Since we don't want two methods of storing arrays in drgn_type, and because this was pretty confusing, I refactored the existing arrays into a dynamically allocated array. 

I'm still not 100% sure about the memory allocations (and I'm not sure how to properly run valgrind and other memory debuggers against python/c extensions), so that's the part that will probably need the most review.

After this and #50, I'm going to (hopefully) get template paramter types accessible through drgn!

I'll add more comments below.

### DIFF
--- a/libdrgn/dwarf_info_cache.c
+++ b/libdrgn/dwarf_info_cache.c
@@ -16,6 +16,9 @@
 
 DEFINE_HASH_TABLE_FUNCTIONS(dwarf_type_map, hash_pair_ptr_type,
 			    hash_table_scalar_eq)
+DEFINE_VECTOR(drgn_type_member_vector, struct drgn_type_member)
+DEFINE_VECTOR(drgn_type_enumerator_vector, struct drgn_type_enumerator)
+DEFINE_VECTOR(drgn_type_parameter_vector, struct drgn_type_parameter)
 
 struct drgn_type_from_dwarf_thunk {
 	struct drgn_type_thunk thunk;
@@ -24,43 +27,33 @@ struct drgn_type_from_dwarf_thunk {
 	bool can_be_incomplete_array;
 };
 
-static bool drgn_type_realloc(struct drgn_type **type, size_t capacity,
-			      size_t element_size)
-{
-	struct drgn_type *tmp;
-	size_t size;
-
-	if (__builtin_mul_overflow(capacity, element_size, &size) ||
-	    __builtin_add_overflow(size, sizeof(**type), &size))
-		return false;
-
-	tmp = realloc(*type, size);
-	if (!tmp)
-		return false;
-
-	*type = tmp;
-	return true;
-}
-
 static void drgn_dwarf_type_free(struct drgn_dwarf_type *dwarf_type)
 {
 	if (dwarf_type->should_free) {
 		struct drgn_type *type = dwarf_type->type;
 
 		if (drgn_type_has_members(type)) {
+			struct drgn_type_member *members;
 			size_t num_members, i;
 
+			members = drgn_type_members(type);
 			num_members = drgn_type_num_members(type);
 			for (i = 0; i < num_members; i++)
-				drgn_type_member_deinit(type, i);
+				drgn_type_member_deinit(&members[i]);
+			free(members);
 		}
 		if (drgn_type_has_parameters(type)) {
+			struct drgn_type_parameter *parameters;
 			size_t num_parameters, i;
 
+			parameters = drgn_type_parameters(type);
 			num_parameters = drgn_type_num_parameters(type);
 			for (i = 0; i < num_parameters; i++)
-				drgn_type_parameter_deinit(type, i);
+				drgn_type_parameter_deinit(&parameters[i]);
+			free(parameters);
 		}
+		if (drgn_type_has_enumerators(type))
+			free(drgn_type_enumerators(type));
 		free(type);
 	}
 }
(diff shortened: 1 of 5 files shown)


libdrgn/python/type.c
```diff
@@ -1578,13 +1586,16 @@ static DrgnType *compound_type(PyObject *tag_obj, PyObject *size_obj,
 		if (!cached_members_obj)
 			return NULL;
 		num_members = PyTuple_GET_SIZE(cached_members_obj);
+		members = malloc(sizeof(struct drgn_type_member) * num_members);
```
jgkamat, line 1589: This malloc is in a bit of a weird place, it is created and built up here, but it is actually used below. Since the condition is the same as here, I think it should be safe but I was wondering if there was a better/cleaner way to do this.

libdrgn/python/type.c
```diff
@@ -1578,13 +1582,16 @@ static DrgnType *compound_type(PyObject *tag_obj, PyObject *size_obj,
...
+		if (!members)
+			goto err;
 
 		type_obj = DrgnType_new(qualifiers, num_members,
 					sizeof(struct drgn_type_member));
```
osandov, line 1594: This should be `DrgnType_new(qualifiers, 0, 0)` now, since the members array isn't embedded anymore.

libdrgn/python/type.c
osandov: I'm okay with having it here. However, please use `malloc_array(num_members, sizeof(*members))`.

libdrgn/type.h
```diff
@@ -255,32 +253,30 @@ static inline void drgn_type_member_init(struct drgn_type *type, size_t i,
  *
  * This only frees @ref drgn_type_member::type.
  *
- * @param[out] type Type containing member to free.
+ * @param[out] members Array containing member to free.
  * @param[in] i Index of member to free.
  */
-static inline void drgn_type_member_deinit(struct drgn_type *type, size_t i)
+static inline void drgn_type_member_deinit(struct drgn_type_member *members, size_t i)
```
osandov, line 259: This should also take a single `struct drgn_type_member *`.

libdrgn/dwarf_info_cache.c
```diff
@@ -63,6 +64,9 @@ static void drgn_dwarf_type_free(struct drgn_dwarf_type *dwarf_type)
 			for (i = 0; i < num_parameters; i++)
 				drgn_type_parameter_deinit(type, i);
 		}
+		if (drgn_type_has_enumerators(type)) {
+			free(drgn_type_enumerators(type));
+		}
```
osandov: Style nit: no braces around this one-line statement.

libdrgn/python/type.c
```diff
@@ -564,6 +564,9 @@ static void DrgnType_dealloc(DrgnType *self)
 			for (i = 0; i < num_parameters; i++)
 				drgn_lazy_type_deinit(&parameters[i].type);
 		}
+		if (drgn_type_has_enumerators(self->type)) {
+			free(drgn_type_enumerators(self->type));
+		}
```
osandov: Ditto, no braces.

libdrgn/dwarf_info_cache.c
```diff
@@ -16,6 +16,7 @@
 
 DEFINE_HASH_TABLE_FUNCTIONS(dwarf_type_map, hash_pair_ptr_type,
 			    hash_table_scalar_eq)
+DEFINE_VECTOR(type_member_vector, struct drgn_type_member)
```
osandov, line 19: It's more verbose, but I'd prefer to spell out `drgn_type_member_vector` to be explicit.

libdrgn/python/type.c
```diff
@@ -552,6 +553,7 @@ static void DrgnType_dealloc(DrgnType *self)
 			num_members = drgn_type_num_members(self->type);
 			for (i = 0; i < num_members; i++)
 				drgn_lazy_type_deinit(&members[i].type);
```
osandov, line 555: While you're here, could you make this use `drgn_type_member_deinit()`?

libdrgn/dwarf_info_cache.c
```diff
@@ -676,29 +669,26 @@ drgn_compound_type_from_dwarf(struct drgn_dwarf_info_cache *dicache,
 					"libdw could not parse DIE children");
 		goto err;
 	}
-	if (capacity != num_members) {
-		/* We don't care if this fails. */
-		drgn_type_realloc(&type, num_members,
-				  sizeof(struct drgn_type_member));
-	}
+	type_member_vector_shrink_to_fit(&members);
+
```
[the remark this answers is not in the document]
jgkamat: I think the malloc has to stay earlier because of the 'incomplete type' declarations above:

```
	if (declaration) {
		switch (kind) {
		case DRGN_TYPE_STRUCT:
			drgn_struct_type_init_incomplete(type, tag, lang);
```

let me know if I misunderstand though!

libdrgn/python/type.c
jgkamat: I decided to do this on the leading 3 commits, and in the cleanup commit I just removed those parameters entirely (as they weren't used).

libdrgn/dwarf_info_cache.c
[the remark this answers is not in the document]
osandov: I missed that, you're right!

osandov:
Merged, thanks!

PR #57, opened 2020-04-08
libdrgn: allow to build without openmp

jeffmahoney:
The configure script allows the user to not use any openmp implementation
but dwarf_index.c uses the locking APIs unconditionally.

Adding simple stubs for the locking API allows it to build properly.
This is useful when debugging crashes in dwarf indexing during development.


osandov:
Hm, what compiler and version are you using, and what error message are you getting? I tried both GCC 7.5.0 and GCC 9.3.0 and `--disable-openmp` worked just fine since we're including `omp.h` either way.

jeffmahoney:
gcc 7.4.1

osandov:
Oh, for me it builds but doesn't run because it's not linked against the runtime. So you're right, this is necessary.

jeffmahoney:
Yep, that's what I'm seeing here. Sorry for the mixup.

osandov:
How about something like:

```diff
diff --git a/libdrgn/dwarf_index.h b/libdrgn/dwarf_index.h
index 52843795a..8b312618c 100644
--- a/libdrgn/dwarf_index.h
+++ b/libdrgn/dwarf_index.h
@@ -14,10 +14,19 @@
 
 #include <elfutils/libdwfl.h>
 #include <libelf.h>
-#include <omp.h>
 #include <stddef.h>
 #include <stdint.h>
 
+#ifdef _OPENMP
+#include <omp.h>
+#else
+typedef struct {} omp_lock_t;
+#define omp_init_lock(lock) do {} while(0)
+#define omp_set_lock(lock) do {} while(0)
+#define omp_unset_lock(lock) do {} while(0)
+#define omp_destroy_lock(lock) do {} while(0)
+#endif
+
 #include "drgn.h"
 #include "hash_table.h"
 #include "string_builder.h"
```

The compiler defines `_OPENMP` when compiling with OpenMP, so we don't need to define our own macro. Making the `#include <omp.h>` conditional makes sure that any new OpenMP dependencies fail at build time.

jeffmahoney:
Yep, that works and is cleaner.

osandov:
I managed to confuse GitHub, but this is merged as https://github.com/osandov/drgn/commit/bf05d9bf3ffa424b538499638b199d7bb00242ae.

PR #58, opened 2020-04-10
Enable DWARF indexing to work with partial units

jeffmahoney:
Loading debuginfo from a kernel with separate or re-combined debuginfo
results in the following failure:

  File "/home/jeffm/.local/bin/drgn", line 11, in <module>
    load_entry_point('drgn==0.0.3+39.gbf05d9bf', 'console_scripts', 'drgn')()
  File "/home/jeffm/.local/lib/python3.6/site-packages/drgn-0.0.3+39.gbf05d9bf-py3.6-linux-x86_64.egg/drgn/internal/cli.py", line 121, in main
    prog.load_debug_info(args.symbols, **args.default_symbols)
Exception: invalid DW_AT_decl_file 10

A typical kernel build results in DWARF information with
DW_TAG_compile_unit tags but the objcopy --only-keep-debug and eu-unstrip
commands use DW_TAG_partial_unit tags.  The DWARF indexer only handles
the former, so the file table isn't populated and we get the exception.

Fortunately, the two are more or less interchangeable. (See
http://dwarfstd.org/doc/Dwarf3.pdf, section E.2.3).  Accepting both for
indexing compile units results in a working session.

The RPMs that SUSE ships contain stripped binaries with optional packages available for download that contain only the debuginfo. This change is mandatory for working with our kernel packages.

### DIFF
--- a/libdrgn/dwarf_index.c
+++ b/libdrgn/dwarf_index.c
@@ -1410,7 +1410,8 @@ static struct drgn_error *read_abbrev_decl(const char **ptr, const char *end,
 		break;
 	}
 
-	if (should_index || tag == DW_TAG_compile_unit)
+	if (should_index || tag == DW_TAG_compile_unit ||
+	    tag == DW_TAG_partial_unit)
 		die_flags = tag;
 	else
 		die_flags = 0;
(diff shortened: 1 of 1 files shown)


osandov:
Thanks! I wasn't able to create any partial units with `objcopy` or `eu-unstrip`, but I remember running into these when I tried [`dwz`](https://sourceware.org/dwz/). `dwz` also needs #59, but if this is enough for SUSE kernels, I'll go ahead and merge this part.

PR #64, opened 2020-06-19
Stack vars poc3

jeffmahoney:
Hi Omar -

I don't expect this pull request to be accepted but I wanted to post what I have so far for discussion.  I've gotten stack variables and parameters working, hooked into the python code, and run through some testing (e.g. for_each_task print all variables and params in all frames). I've rebased it on your stack-trace-variables branch -- at least everything until the stack variable commit itself. I had introduced a block cache tree which had the benefit of delaying resolution of scopes until needed, but at the cost of indexing a bunch of new tags and building a tree subject to a single lock. I like attaching the scopes to the frame a lot more.

I haven't modified the stack frame formatter to incorporate the parameters but it will probably make sense to do that.  If we go that route, that'll change how I handle parameters entirely.  The code right now enumerates parameters and locals when the value for one or the count are requested. It keeps a vector of each as symbols (name, Dwarf_Die). When an object is required, it creates one to represent a stack object with type, size, etc but no value or address. It's a normal object but has a needs_stack_evaluation flag set that causes temporary object to be used to evaluate the location when the value is requested and uses the result to populate the new object. If we include parameters in the formatter, I'd likely just keep a vector of named objects instead of the symbols.

### DIFF
--- a/_drgn.pyi
+++ b/_drgn.pyi
@@ -1046,6 +1046,15 @@ class StackFrame:
     """
     A ``StackFrame`` represents a single *frame* (i.e., function call) in a
     thread's call stack.
+
+    :class:`str() <str>` returns a pretty-printed stack frame:
+
+    >>> prog.stack_trace(1)[0]
+    #0 at 0xffffffff9a883039 (__schedule+0x2c9/0x75d)
+
+    This includes more information than when printing the full stack trace. The
+    drgn CLI is set up so that stack frames are displayed with ``str()`` by
+    default.
     """
 
     pc: int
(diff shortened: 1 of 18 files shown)


osandov:
This functionality was finally added in the work culminating in https://github.com/osandov/drgn/commit/bc85767e5fde24ceefb45b6b123092363c10a8e0. The interface is a little different from the one here, and in order to support more DWARF operations I had to implement DWARF expressions rather than reusing libdwfl's code, but this was really useful as a jumping off point. Thanks!

PR #66, opened 2020-07-31
helpers: add full type information to linux/mm

brenns10:
I noticed some missing type information in the mm helpers and went ahead and added them. Ran the test suite for sanity but no code really changed.

Related, I have a question regarding the contributing guide:

> Type hints should be provided for all public interfaces other than helpers (including the C extension) and most private interfaces.

Is there any reason helpers are excluded here (other than time/effort)? I find them nice to have and would be happy to contribute more.

### DIFF
--- a/drgn/helpers/linux/mm.py
+++ b/drgn/helpers/linux/mm.py
@@ -10,10 +10,10 @@ Linux memory management (MM) subsystem. Only x86-64 support is currently
 implemented.
 """
 
-from typing import List
+from typing import List, Iterator, Union
 
 from _drgn import _linux_helper_read_vm
-from drgn import Object, cast
+from drgn import Object, Program, cast
 
 
 __all__ = (
@@ -31,7 +31,7 @@ __all__ = (
 )
 
 
-def for_each_page(prog):
+def for_each_page(prog: Program) -> Iterator[Object]:
     """
     Iterate over all pages in the system.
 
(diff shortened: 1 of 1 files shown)


osandov:
Hi, so sorry that I missed this before! There are a few reasons that I haven't added type hints to the helpers yet:

1. I've been wanting to add type checking at the helper level (e.g., if a helper expects a `struct task_struct *`, that should be checked), and I thought it might be possible to combine that with the Python-level type checking in some way. After some testing I've mostly given up on this idea, but it was holding me back from adding annotations earlier.
2. More importantly, several helpers need overloads with `@typing.overload`, and my [documentation generation extension](https://github.com/osandov/drgn/tree/master/docs/exts/drgndoc) doesn't support that yet. I have a work in progress for that, but I'm not sure when I'll get back to it.
3. Sphinx doesn't really like the C signatures in the docstring, and some helpers don't have an equivalent C signature, so I wanted to move away from that to just use `:param:` for everything and do that at the same time as annotating helpers.

So there are a couple of blockers here, but I'd be happy to revisit this once those are unblocked.

brenns10:
That all makes sense, it's pretty reasonable to table this! Thanks for taking the time. Feel free to close if you want to keep the PR list clean :)

osandov:
The work-in-progress I had previously was more complete than I remembered, so I finished it up and pushed it. Helpers are now fully annotated, thanks for giving me the nudge to finish it :)

PR #67, opened 2020-08-14
libdrgn: Add support for storing and returning namespace tags

jgkamat:
Sorry in advance for how rough this is. Right now this is a first-pass so there's a lot of things that are unaddressed (tests, missing/stale/debug comments). I'll fix all of these as we go along, but for a first pass I'm more interested in just the architecture (I'll clean up style issues later on). For that, I have these comments: 

- Right now the namespace object is a memory leak, and I'm not sure how to free it when the dindex (?) gets freed. The easiest way I can think of is having a vector of the allocated ns objects in the dindex, but I think we prefer avoiding vector types in structs.


libdrgn/dwarf_index.c
```diff
@@ -2351,10 +2423,25 @@ static struct drgn_error *index_cu_second_pass(struct drgn_dwarf_index *dindex,
...
+struct drgn_error *drgn_dwarf_index_namespace_resolve(
+	struct drgn_dwarf_index_namespace *ns) {
+	struct drgn_error *err = NULL;
+	while (ns->pending_cus.size > 0) {
+		struct drgn_dwarf_index_pending_cu* cu =
+			drgn_dwarf_index_pending_cu_vector_pop(&ns->pending_cus);
+		if (err = index_cu_second_pass(ns, cu->cu, cu->ptr)) {
+			return err;
+		}
+	}
```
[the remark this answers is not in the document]
jgkamat: I'm not 100% sure, but I think (from my limited tests) that the first pass already goes into namespaces. It seems to use only `(insn & DIE_FLAG_CHILDREN)` for deciding whether to go deeper on the node (as opposed to the second pass).

libdrgn/dwarf_index.c
```diff
@@ -2357,11 +2413,28 @@ static struct drgn_error *index_cu_second_pass(struct drgn_dwarf_index *dindex,
...
+					     tag,
 					     file_name_hash, module,
-					     die_offset)))
+					     die_offset, &new_ns)))
 				return err;
+
+			if (new_ns) {
+				struct drgn_dwarf_index_pending_cu *pending;
+				#pragma omp critical(pending_cu_append)
+				{
```
osandov, line 2427: I believe the `{}` block is unnecessary. The `#pragma omp critical` can apply directly to the `if` statement.

libdrgn/dwarf_index.c
```diff
@@ -2357,11 +2413,28 @@ static struct drgn_error *index_cu_second_pass(struct drgn_dwarf_index *dindex,
...
+					if (pending = drgn_dwarf_index_pending_cu_vector_append_entry(
```
osandov, line 2428: Add extra parentheses around the assignment expression to make it clear that it's intentional and not a typo for `==`: `if ((pending = ...))`.

libdrgn/dwarf_index.c
```diff
@@ -2381,10 +2455,31 @@ static struct drgn_error *index_cu_second_pass(struct drgn_dwarf_index *dindex,
...
 }
 
+
+struct drgn_error *drgn_dwarf_index_namespace_resolve(
+	struct drgn_dwarf_index_namespace *ns) {
+	struct drgn_error *err = NULL;
+
+	#pragma omp parallel
+	{
+		#pragma omp for schedule(dynamic)
```
osandov, line 2465: These can be collapsed into

```c
#pragma omp parallel for schedule(dynamic)
for (...) {
	...
}
```

libdrgn/dwarf_index.c
```diff
@@ -2381,10 +2455,31 @@ static struct drgn_error *index_cu_second_pass(struct drgn_dwarf_index *dindex,
...
+		for (size_t i = 0; i < ns->pending_cus.size; i++) {
+			if (!err) {
```
osandov, line 2467: Missing actual error handling, I'm guessing you're leaving that for later :)

libdrgn/dwarf_index.c
```diff
@@ -1834,10 +1872,10 @@ index_specification(struct drgn_dwarf_index *dindex, uintptr_t declaration,
  * DW_AT_specification.
  */
 static struct drgn_error *index_cu_first_pass(struct drgn_dwarf_index *dindex,
-					      struct drgn_dwarf_index_cu *cu)
+					      struct drgn_dwarf_index_cu *cu,
+					      const char* ptr)
```
osandov, line 1876: Since you changed it so that `index_cu_first_pass()` is only done for the initial indexing, you don't need the `ptr` parameter anymore.

libdrgn/dwarf_index.c
jgkamat: Oh, this is actually sort of a micro-optimization, I didn't want to acquire the lock (I'm assuming this is a lock) if new_ns is false. I could move this outside though if you think that won't matter, what do you think?

osandov: I meant it could apply directly to the `if ((pending = ...))` statement. But now that I'm looking at it again, maybe it'd be better to pass in the `cu` to `index_die()`, and have `index_die()` add the `pending_cu` within the shard lock so we don't need an extra critical section and we don't need to return the namespace at all?

libdrgn/dwarf_index.c
```diff
@@ -2501,21 +2508,25 @@ drgn_dwarf_index_report_end_internal(struct drgn_dwarf_index *dindex,
...
 	 * After this point, if we hit an error, then we have to roll back the
 	 * index.
 	 */
-	err = index_cus(dindex, cus.data, cus.size);
+	err = index_cus(dindex, old_cus_size);
 	if (err) {
+		for (size_t i = old_cus_size; i < dindex->cus.size; i++) {
+			drgn_dwarf_index_cu_deinit(&dindex->cus.data[i]);
+		}
+		dindex->cus.size = old_cus_size;
```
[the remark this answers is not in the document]
jgkamat: I'll put it on the err label instead, that way I won't have to check for err being non-null (and it makes more sense there in my opinion).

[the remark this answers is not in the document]
osandov: :+1: good point, that's better.

libdrgn/dwarf_index.c
jgkamat: sounds good, that's a lot nicer!

libdrgn/dwarf_index.c
```diff
@@ -492,22 +516,65 @@ struct drgn_error *drgn_dwarf_index_init(struct drgn_dwarf_index *dindex,
...
+	struct drgn_error *err;
+	struct drgn_dwarf_index_iterator_vector it_vector;
+	struct drgn_dwarf_index_iterator *it;
+	struct drgn_dwarf_index_die *die;
+
+	drgn_dwarf_index_iterator_vector_init(&it_vector);
+
+	it = drgn_dwarf_index_iterator_vector_append_entry(&it_vector);
+	if (!it)
+		// We can't do much on memory failures here.
```
jgkamat, line 536: Just flagging this - in order to free this completely, we have to allocate some memory (either through recursion or like this). Right now this is a leak because I don't have a better idea (other than making the deallocation something that can fail...). Hopefully this will work out as we just freed a large chunk of the index just before this.

libdrgn/dwarf_index.c
```diff
@@ -2346,10 +2444,14 @@ static struct drgn_error *index_cu_second_pass(struct drgn_dwarf_index *dindex,
...
-			if ((err = index_die(dindex, name,
-					     insn & DIE_FLAG_TAG_MASK,
+
+			struct drgn_dwarf_index_namespace *new_ns = NULL;
+			if ((err = index_die(ns, name,
+					     tag,
 					     file_name_hash, module,
-					     die_offset)))
+					     die_offset,
+					     die_offset + debug_info_buffer,
```
jgkamat, line 2451: This is a little inelegant, I could just store the offset instead of the ptr, but then I would need to have logic to detect if it's being passed in again (debug_info_buffer isn't available outside of this fn).

libdrgn/dwarf_index.c
```diff
@@ -2125,23 +2199,34 @@ static struct drgn_error *index_die(struct drgn_dwarf_index *dindex,
...
-	if (!append_die_entry(shard, tag, file_name_hash, module, offset)) {
+	if (!append_die_entry(shard, tag, file_name_hash, module, offset, ns->dindex)) {
 		err = &drgn_enomem;
 		goto out;
 	}
+	new_die = &shard->dies.data[shard->dies.size - 1];
 	shard->dies.data[index].next = shard->dies.size - 1;
 	err = NULL;
 out:
+	if (!err && new_die && new_die->tag == DW_TAG_namespace) {
```
osandov, line 2210: When will `new_die` be `NULL` here?

libdrgn/dwarf_index.c
osandov: Hm, if we store the offset in the `pending_cu`, can't we get the pointer as `section_ptr(pending_cu->cu->userdata->debug_info, 0) + pending_cu->offset`?

libdrgn/dwarf_index.c
```diff
@@ -2381,10 +2471,41 @@ static struct drgn_error *index_cu_second_pass(struct drgn_dwarf_index *dindex,
...
+	struct drgn_dwarf_index_namespace *ns) {
+	if (ns->saved_err)
+		return drgn_error_copy(ns->saved_err);
+
+	struct drgn_error *err = NULL;
+
+	#pragma omp for schedule(dynamic)
+	for (size_t i = 0; i < ns->pending_cus.size; i++) {
+		if (!err) {
+			struct drgn_dwarf_index_pending_cu* cu = &ns->pending_cus.data[i];
```
osandov, line 2485: Style nit: space before asterisk, not after.

libdrgn/dwarf_index.c
```diff
@@ -2533,6 +2651,9 @@ drgn_dwarf_index_report_end_internal(struct drgn_dwarf_index *dindex,
 err:
 	drgn_dwarf_index_free_modules(dindex, false, false);
 	drgn_dwarf_index_reset_errors(dindex);
+	for (size_t i = old_cus_size; i < dindex->cus.size; i++)
+		drgn_dwarf_index_cu_deinit(&dindex->cus.data[i]);
+	dindex->cus.size = old_cus_size;
```
osandov, line 2656: This hunk should've been folded into the previous commit.

libdrgn/dwarf_index.c
```diff
@@ -2677,6 +2798,17 @@ drgn_dwarf_index_iterator_next(struct drgn_dwarf_index_iterator *it,
 		}
 	}
 
+	if (resolve_ns && die->tag == DW_TAG_namespace)
```
osandov, line 2801: I think we should be indexing the namespace when it is first passed to `drgn_dwarf_index_iterator_init()` (i.e., the first time we actually want to access it) rather than here. `drgn_dwarf_index_iterator_init()` would need to return `struct drgn_error *`, but that's an easy change.

libdrgn/dwarf_index.c
```diff
@@ -2686,3 +2818,21 @@ drgn_dwarf_index_iterator_next(struct drgn_dwarf_index_iterator *it,
...
+			       Dwarf_Die *die_ret, uint64_t *bias_ret)
+{
+	struct drgn_dwarf_index_die *die;
+	die = drgn_dwarf_index_iterator_next_die(it, true);
+	if (!die)
+		return &drgn_stop;
+	return drgn_dwarf_index_convert_die(die, die_ret, bias_ret);
+}
+
+inline struct drgn_dwarf_index_namespace *
```
osandov, line 2833: There's no point in making this inline since it can't be seen outside of `dwarf_index.c`. I think we might as well just move the definition of `struct drgn_dwarf_index_die` to `dwarf_index.h`.

libdrgn/dwarf_index.c
osandov: I'm on the fence about this vs. recursion. I'm leaning towards recursion so we don't have to do the explicit memory allocation. If we're paranoid, we can limit the depth we're willing to free (or limit it earlier before we create a namespace of that depth), but practically I think we can get by without it.

libdrgn/dwarf_index.c
```diff
@@ -513,14 +570,11 @@ void drgn_dwarf_index_deinit(struct drgn_dwarf_index *dindex)
...
-		drgn_dwarf_index_die_map_deinit(&dindex->shards[i].map);
-		omp_destroy_lock(&dindex->shards[i].lock);
-	}
-	for (size_t i = 0; i < dindex->cus.size; i++) {
+	for (size_t i = 0; i < dindex->cus.size; i++)
 		drgn_dwarf_index_cu_deinit(&dindex->cus.data[i]);
-	}
+	drgn_dwarf_index_deinit_namespaces(&dindex->ns);
+	drgn_dwarf_index_shards_deinit(dindex->ns.shards, ARRAY_SIZE(dindex->ns.shards));
+	drgn_dwarf_index_pending_cu_vector_deinit(&dindex->ns.pending_cus);
```
osandov, line 577: This should just be done inside of `drgn_dwarf_index_deinit_namespaces()`.

libdrgn/dwarf_info_cache.c
```diff
@@ -1475,6 +1478,50 @@ drgn_dwarf_object_find(const char *name, size_t name_len, const char *filename,
...
+		const int ns_len = ns_ptr - name;
+		const size_t ns_colon_len = ns_len + 2;
+		struct drgn_dwarf_index_iterator it;
+		uint64_t ns_tag = DW_TAG_namespace;
+		struct drgn_dwarf_index_die *die;
+
+		drgn_dwarf_index_iterator_init(&it, ns, name, ns_len, &ns_tag, 1);
+		// Namespaces should be de-duplicated so we only check for the first one.
+		while ((die = drgn_dwarf_index_iterator_next_die(&it, true))) {
+			if (drgn_dwarf_index_die_namespace(die)) {
```
osandov, line 1505: Since we're only iterating over namespace DIEs here, we don't need to check this, right?

libdrgn/dwarf_index.c
jgkamat: I was worried about any corner cases causing a segfault here, but I think when `!err` this should be non-null. I'll change it back.

libdrgn/dwarf_info_cache.c
jgkamat: In theory this should never be false but I was worried that if there was some bug where we returned a non-namespace type, this would explode, I'll remove it.

libdrgn/program.c
```diff
@@ -635,8 +635,10 @@ static void drgn_program_set_language_from_main(struct drgn_program *prog,
 	struct drgn_dwarf_index_iterator it;
 	static const uint64_t tags[] = { DW_TAG_subprogram };
 
-	drgn_dwarf_index_iterator_init(&it, dindex, "main", strlen("main"),
-				       tags, ARRAY_SIZE(tags));
+	err = drgn_dwarf_index_iterator_init(&it, &dindex->ns, "main", strlen("main"),
+					     tags, ARRAY_SIZE(tags), true);
+	if (err)
+		drgn_error_destroy(err);
```
jgkamat, line 641: I don't know how to handle this - just making sure this is ok

libdrgn/dwarf_index.c
jgkamat: I guess that would work, I was worried it would just be a little messy at that point then, I'll switch it over.

libdrgn/dwarf_index.h
```diff
@@ -193,6 +217,45 @@ DEFINE_HASH_TABLE_TYPE(drgn_dwarf_module_table, struct drgn_dwarf_module *,
 
 DEFINE_HASH_SET_TYPE(c_string_set, const char *)
 
+struct drgn_dwarf_index_cu;
+DEFINE_VECTOR_TYPE(drgn_dwarf_index_cu_vector, struct drgn_dwarf_index_cu)
+
+struct drgn_dwarf_index_pending_cu {
```
osandov: I guess `struct drgn_dwarf_index_pending_die` would be a more accurate name.

libdrgn/dwarf_index.h
```diff
@@ -205,11 +268,9 @@ DEFINE_HASH_SET_TYPE(c_string_set, const char *)
  */
 struct drgn_dwarf_index {
 	/**
-	 * Index shards.
-	 *
-	 * This is sharded to reduce lock contention.
+	 * Container of default, top-level shards for this index.
 	 */
-	struct drgn_dwarf_index_shard shards[1 << DRGN_DWARF_INDEX_SHARD_BITS];
+	struct drgn_dwarf_index_namespace ns;
```
osandov: This is the global namespace, so naming this `global` would be more clear.

libdrgn/dwarf_index.h
```diff
@@ -424,6 +492,32 @@ struct drgn_error *
...
+
+/**
+ * Get the next matching drgn_dwarf_die from a DWARF index iterator.
+ *
+ * This is a more bare-bones version of drgn_dwarf_index_iterator_next, not doing the conversion to a Dwarf DIE.
+ *
+ * @param[in] it DWARF index iterator.
+ */
+struct drgn_dwarf_index_die *
+drgn_dwarf_index_iterator_next_die(struct drgn_dwarf_index_iterator *it);
```
osandov: I'd rather get rid of the old `drgn_dwarf_index_iterator_next()` and make everywhere do the conversion manually.

libdrgn/dwarf_info_cache.c
```diff
@@ -1475,6 +1485,52 @@ drgn_dwarf_object_find(const char *name, size_t name_len, const char *filename,
...
+				       void *arg, struct drgn_object *ret) {
+	struct drgn_error *err;
+	struct drgn_dwarf_info_cache *dicache = arg;
+	char* ns_ptr;
+
+	if (!ns)
+		ns = &dicache->dindex.ns;
+
+find_ns:
+	ns_ptr = strstr(name, "::");
```
osandov: `name` has a given length and isn't necessarily null-terminated, so this needs to use `memmem()`.

libdrgn/dwarf_info_cache.c
```diff
@@ -1475,6 +1485,52 @@ drgn_dwarf_object_find(const char *name, size_t name_len, const char *filename,
...
+		err = drgn_dwarf_index_iterator_init(&it, ns, name, ns_len, &ns_tag, 1, true);
+		if (err)
+			return err;
+		// Namespaces should be de-duplicated so we only check for the first one.
+		while ((die = drgn_dwarf_index_iterator_next_die(&it))) {
+			assert(die->tag == DW_TAG_namespace);
+			name = name + ns_colon_len;
+			ns = die->namespace;
+			// Dive into this namespace
+			goto find_ns;
```
osandov: Not a fan of this `goto`. This can be a loop while we find a `::` string.

libdrgn/dwarf_info_cache.c
```diff
@@ -1475,6 +1485,52 @@ drgn_dwarf_object_find(const char *name, size_t name_len, const char *filename,
...
+	}
+
+	return drgn_dwarf_object_find_internal(name, name_len, filename, flags, ns, arg, ret);
+}
+
+struct drgn_error *
+drgn_dwarf_object_find(const char *name, size_t name_len, const char *filename,
+		       enum drgn_find_object_flags flags, void *arg,
+		       struct drgn_object *ret) {
+	return drgn_dwarf_object_find_with_namespaces(name, name_len, filename, flags, NULL, arg, ret);
```
osandov: I'd just combine `drgn_dwarf_object_find()`, `drgn_dwarf_object_find_with_namespaces()`, and `drgn_dwarf_object_find_internal()`.

libdrgn/program.c
osandov: This is fine, but we need to return, not march forward with an invalid iterator.

libdrgn/dwarf_index.c
```diff
@@ -466,19 +446,35 @@ static void drgn_dwarf_index_free_modules(struct drgn_dwarf_index *dindex,
...
+		omp_init_lock(&shard->lock);
+		drgn_dwarf_index_die_map_init(&shard->map);
+		drgn_dwarf_index_die_vector_init(&shard->dies);
+	}
+	ns->saved_err = NULL;
+	ns->dindex = dindex;
+	drgn_dwarf_index_pending_cu_vector_init(&ns->pending_cus);
+}
+
+static void drgn_dwarf_index_shards_deinit(struct drgn_dwarf_index_shard* shards, size_t num_shards) {
```
osandov: This should just be in `drgn_dwarf_index_deinit_namespaces()`.

libdrgn/dwarf_index.c
```diff
@@ -492,6 +488,35 @@ struct drgn_error *drgn_dwarf_index_init(struct drgn_dwarf_index *dindex,
...
+}
+
+static void drgn_dwarf_index_deinit_namespaces(struct drgn_dwarf_index_namespace *parent_ns) {
+	const uint64_t ns_tag = DW_TAG_namespace;
+	struct drgn_error *err;
+	struct drgn_dwarf_index_iterator it;
+	struct drgn_dwarf_index_die *die;
+
+
+	err = drgn_dwarf_index_iterator_init(&it, parent_ns, NULL, 0, &ns_tag, 1, false);
```
osandov: I don't like exposing `resolve_ns` as part of the iterator interface. This can just walk the shards manually so we don't need to use the iterator.

libdrgn/dwarf_index.c
```diff
@@ -2346,10 +2398,13 @@ static struct drgn_error *index_cu_second_pass(struct drgn_dwarf_index *dindex,
 				file_name_hash = cu->file_name_hashes[decl_file - 1];
 			else
 				file_name_hash = 0;
-			if ((err = index_die(dindex, name,
-					     insn & DIE_FLAG_TAG_MASK,
+
+			struct drgn_dwarf_index_namespace *new_ns = NULL;
```
osandov: This isn't used anymore.

osandov (commented):
I had a few last comments, but I went ahead and fixed those up myself. Would you mind looking it over and testing it? If it all looks good, please add your `Signed-off-by` to the commits (see [here](https://github.com/osandov/drgn/blob/master/CONTRIBUTING.rst#submitting-prs)), push them, and I'll merge it. Thanks so much!

osandov:
(I also squashed the testing commit into the top commit, and expanded the commit messages.)

Issue #68, opened 2020-08-17
manylinux wheels


brenns10:
Hello again,

I've written some scripts which generate manylinux2010 wheels for drgn. These wheels require no compilation to install, and they bundle the dependencies (such as libkdumpfile) so that it is quite easy to install on any reasonably recent Linux distro (tested on RHEL/OL6+, Ubuntu 20.04).

Among my coworkers, the difficulty to build and install drgn plus libkdumpfile hinders adoption. Especially on older systems such as RHEL/OL 6/7, due to older compiler versions, etc. I think that distributing wheels might simplify installation enough to encourage new users to try out drgn and lower the bar to entry from Crash. Both libkdumpfile and drgn are GPLv3, so there shouldn't be any license concerns with wheel distribution.

Would you be interested in a PR containing these scripts?

osandov:
We now provide manylinux2010 wheels: https://pypi.org/project/drgn/#files.

PR #69, opened 2020-08-17
Add scripts to build manylinux2010 wheels

brenns10:
Here are the scripts mentioned in #68, along with a short readme. For the most part, they are standard build steps (which could be part of a packaging script), but a few slight hacks are included, most to do with libkdumpfile. Below is what I ran for verification, I can provide the wheels if necessary (github doesn't support attaching wheels in PRs).

    ./build_drgn.sh "36 37 38"

I'm not sure how this will best fit into the repository (e.g. as a travis job?) but am sure something can be worked out.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,5 @@
 /drgn.egg-info
 /htmlcov
 __pycache__
+/wheels/*.whl
+/wheels/*.tar.*
--- a/wheels/README.md
+++ b/wheels/README.md
@@ -1,0 +1,20 @@
+Build Wheels
+============
+
+This directory contains scripts for creating manylinux wheels. The following are
+required:
+
+* docker
+* curl
+* bash
+
+To build wheels:
+
+    cd wheels  # your working directory MUST be wheels
+    ./build_drgn.sh "36 37 38"
+
+The above builds wheels for Python 3.6, 3.7, and 3.8. The quotes are required.
+Resulting wheels will be placed in the current directory.
+
+If your user account is not authorized to use Docker commands, but you have
+sudo, you can use `DOCKER="sudo docker" ./build_drgn.sh ...` instead.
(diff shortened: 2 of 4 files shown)


brenns10:
Hi @osandov just wanted to check back in on this. Let me know if I can do anything to make these scripts easier to review.

osandov:
Hi, sorry it took me awhile to get to this. Since f83bb7c71bb8f01e90e2d29b57c0ce4f2022aad6, libdrgn uses the OpenMP taskloop directive. Apparently, libgomp in CentOS 6 doesn't support taskloop. There are a few options:
1. Build a newer libgomp in the container.
2. Stop using taskloop in libdrgn.
3. Use manylinux2014 instead of manylinux2010.

Given that RHEL/CentOS 6 will be EOL at the end of next month, I'm not very excited about options 1 or 2. I'm leaning towards option 3, which would also allow us to get rid of the hacks in the build script. However, [this page](https://packaging.python.org/specifications/platform-compatibility-tags/) notes that "the manylinux2014 specification is relatively new and is not yet widely recognised by install tools". I'm not sure whether that's still true.

I pushed some scripts for building on manylinux2014 (based on your PR and the official [example](https://github.com/pypa/python-manylinux-demo)) to the [manylinux2014](https://github.com/osandov/drgn/tree/manylinux2014) branch.

Do you have any opinions/suggestions here? Thanks!

brenns10:
For my use cases, RHEL/CentOS 6 & its derivatives will be living on for quite a lot longer, as many people pay for extended support. As a result, I'd personally use option 1 as I don't think it would be too difficult to script up another library installation. However, installing a non-standard libgomp might break other manylinux2010 wheels which use libgomp.

That said, I don't think that's a problem for upstream projects to deal with. If it makes most sense and is maintainable to do manylinux2014, then do that. I can continue to build 2010 wheels where necessary, and the 2014 wheels would be much better than nothing else.

osandov:
libdrgn doesn't use taskloop anymore, so I was able to get manylinux2010 to work. I pushed some scripts adapted from yours, and I'll start building manylinux2010 wheels starting with the next release. Sorry for the delay, and thanks for kickstarting this!

PR #72, opened 2020-10-01
Add support for reading methods via "members" array

jgkamat:
This is the first step towards real "method" support (ie: reading methods like `Class::methodname`). It also adds support for `instance.method` right now, and adds the methods to the members array of object types. 

Here are the biggest red flags (that I would like input on):
- Right now the python side is completely unaware that parameters can be objects - this works because objects can be automagically evaluated to types. Is this fine for now, or would you like that to be implemented here as well. 
- Anywhere outside the core - parameters are converted to types for compatibility. Is there anywhere a dual pathway makes sense.
- I am using the `lazy_type` and `lazy_object` variable names to signify the parameter must be a type/object at that point. I could split things a little more to remove this distinction if you wish.
- I still need to work on adding tests. 

If you have high-level suggestions for the architecture, I'm also happy to hear those as well.


libdrgn/debug_info.c
```diff
@@ -1535,28 +1581,41 @@ parse_member(struct drgn_debug_info *dbinfo, Dwarf_Die *die,
...
+		// Creating objects
+		err = drgn_lazy_object_from_dwarf(dbinfo, die, bias, name, &member);
+		if (err)
+			return err;
+
+		err = drgn_compound_type_builder_add_member(builder, member, name,
+							    0, bit_field_size);
+		if (err)
+			goto err;
+		return NULL;
```
osandov, line 1595: Could you refactor this so you only need to call `drgn_compound_type_builder_add_member()` in one place?

libdrgn/dwarf_index.h
```diff
@@ -319,6 +319,8 @@ struct drgn_error *drgn_dwarf_index_get_die(struct drgn_dwarf_index_die *die,
 					    Dwarf_Die *die_ret,
 					    uint64_t *bias_ret);
 
+bool find_definition(struct drgn_dwarf_index *dindex, uintptr_t die_addr,
```
osandov, line 322: Since this is being made public, it should be renamed `drgn_dwarf_index_find_definition()` and given a comment.

libdrgn/type.c
```diff
@@ -479,10 +479,13 @@ drgn_compound_type_builder_add_member(struct drgn_compound_type_builder *builder
 				      const char *name, uint64_t bit_offset,
 				      uint64_t bit_field_size)
 {
-	struct drgn_error *err = drgn_lazy_type_check_prog(&parameter,
-							   builder->prog);
-	if (err)
-		return err;
+	struct drgn_error *err = NULL;
+	if (!parameter.is_object) {
+		err = drgn_lazy_type_check_prog(&parameter, builder->prog);
```
osandov, line 484: I think we need to check this in the `is_object` case, too, right?

osandov (changes requested):
The architecture all looks good here. There are a couple of bugs I spotted, but my comments are mostly around naming and commit organization.

libdrgn/debug_info.c
```diff
@@ -1201,7 +1212,52 @@ drgn_lazy_type_from_dwarf(struct drgn_debug_info *dbinfo, Dwarf_Die *parent_die,
...
+	thunk->thunk.prog = dbinfo->prog;
+	thunk->thunk.evaluate_fn = drgn_object_from_dwarf_thunk_evaluate_fn;
+	thunk->thunk.free_fn = drgn_object_from_dwarf_thunk_free_fn;
+	thunk->name = name;
+
+	bool decl;
+	if (!dwarf_flag(die, DW_AT_declaration, &decl) && decl) {
+		Dwfl_Module *module;
+		Dwarf_Off offset;
+		if (drgn_dwarf_index_find_definition(&dbinfo->dindex, (uintptr_t)die->addr, &module, &offset)) {
```
jgkamat, line 1249: Note to self - if we aren't able to find a declaration here, the object returned is pretty odd and can probably cause a crash if mishandled. Not sure if we should worry about that at the moment - but I have been able to generate class functions with no declarations (usually if the function was not used or defined)

libdrgn/lazy_parameter.c
```diff
@@ -0,0 +1,47 @@
...
+
+#include "lazy_parameter.h"
+
+void drgn_lazy_parameter_deinit(struct drgn_lazy_parameter *lazy_parameter)
+{
+	if (!drgn_lazy_parameter_is_evaluated(lazy_parameter))
+		drgn_type_thunk_free(lazy_parameter->thunk);
+}
+
+inline struct drgn_error *
```
osandov, line 18: I'd just drop the `inline` now that this has been moved to a `.c` file.

libdrgn/lazy_parameter.c
```diff
@@ -0,0 +1,47 @@
...
+			return drgn_error_create(DRGN_ERROR_INVALID_ARGUMENT,
+						 "type is from different program");
+		}
+		drgn_lazy_parameter_init_type(lazy_parameter, ret->type,
+					      ret->qualifiers);
+		thunk.free_fn(thunk_ptr);
+	}
+	return NULL;
+}
+
```
osandov, line 76: Stray blank line.

libdrgn/lazy_parameter.h
```diff
@@ -0,0 +1,113 @@
...
+ * @return @c NULL on success, non-@c NULL on error.
+ */
+struct drgn_error *drgn_lazy_parameter_get_type(struct drgn_lazy_parameter *lazy_parameter,
+						struct drgn_qualified_type *ret);
+
+/**
+ * Validate a @ref drgn_program matches the one in a @ref drgn_lazy_parameter.
+ *
+ * @param[in] lazy_parameter Lazy parameter to validate.
+ * @param[in] prog The program to compare against..
```
osandov: Extra "."

libdrgn/lazy_parameter.c
```diff
@@ -5,31 +5,60 @@
 
 void drgn_lazy_parameter_deinit(struct drgn_lazy_parameter *lazy_parameter)
 {
-	if (!drgn_lazy_parameter_is_evaluated(lazy_parameter))
-		drgn_type_thunk_free(lazy_parameter->thunk);
+	if (lazy_parameter->is_object) {
+		if (!drgn_lazy_parameter_is_evaluated(lazy_parameter))
+			drgn_object_thunk_free(lazy_parameter->object_thunk);
+		else
+			free(lazy_parameter->object);
```
osandov, line 12: You also need to call `drgn_object_deinit()` first.

libdrgn/lazy_parameter.c
```diff
@@ -5,31 +5,60 @@
...
+	} else
+		if (!drgn_lazy_parameter_is_evaluated(lazy_parameter))
```
osandov, line 14: Squash these to `else if`.

libdrgn/lazy_parameter.c
```diff
@@ -5,31 +5,60 @@
...
 			       struct drgn_program *prog)
 {
-	if ((drgn_lazy_parameter_is_evaluated(lazy_parameter) ?
-	     drgn_type_program(lazy_parameter->type) :
-	     lazy_parameter->thunk->prog) != prog) {
-		return drgn_error_create(DRGN_ERROR_INVALID_ARGUMENT,
-					 "type is from different program");
+	if (lazy_parameter->is_object) {
+		if ((drgn_lazy_parameter_is_evaluated(lazy_parameter) ?
+		     drgn_type_program(lazy_parameter->object->type) :
```
osandov, line 24: `drgn_object_program(lazy_parameter->object)`

libdrgn/lazy_parameter.c
```diff
@@ -5,31 +5,60 @@
...
+			return drgn_error_create(DRGN_ERROR_INVALID_ARGUMENT,
+						 "type is from different program");
+		}
 	}
 	return NULL;
 }
 
+struct drgn_error
+*drgn_lazy_parameter_evaluate_type(struct drgn_lazy_parameter *parameter,
+				   struct drgn_qualified_type *ret) {
```
osandov, line 42: It's confusing having two functions that get the type of a lazy parameter. Can we make `drgn_lazy_parameter_get_type()` handle both objects and types?

libdrgn/lazy_parameter.c
```diff
@@ -45,3 +74,21 @@ struct drgn_error *drgn_lazy_parameter_get_type(struct drgn_lazy_parameter *lazy
 	return NULL;
 }
 
+struct drgn_error *drgn_lazy_parameter_get_object(struct drgn_lazy_parameter *lazy_parameter,
+						  struct drgn_object *ret) {
```
osandov, line 78: Style nit: new line before the `{`

libdrgn/lazy_parameter.c
```diff
@@ -45,3 +74,21 @@ struct drgn_error *drgn_lazy_parameter_get_type(struct drgn_lazy_parameter *lazy
 	return NULL;
 }
 
+struct drgn_error *drgn_lazy_parameter_get_object(struct drgn_lazy_parameter *lazy_parameter,
```
osandov, line 77: This should probably return an error if the parameter is not an object. Right now it'll just crash.

libdrgn/lazy_parameter.c
```diff
@@ -45,3 +74,21 @@ struct drgn_error *drgn_lazy_parameter_get_type(struct drgn_lazy_parameter *lazy
...
+		struct drgn_object_thunk *thunk_ptr = lazy_parameter->object_thunk;
+		struct drgn_object_thunk thunk = *thunk_ptr;
+		drgn_object_init(&obj, thunk.prog);
+		struct drgn_error *err = thunk.evaluate_fn(thunk_ptr, &obj);
+		if (err)
+			return err;
+		drgn_lazy_parameter_init_object(lazy_parameter, &obj);
+		thunk.free_fn(thunk_ptr);
+	}
+	drgn_object_init(ret, drgn_object_program(lazy_parameter->object));
```
osandov, line 90: The convention in libdrgn is to assume that a passed object is already initialized. Rather than calling `drgn_object_init()` here, the caller should do it.

libdrgn/lazy_parameter.c
```diff
@@ -45,3 +74,21 @@ struct drgn_error *drgn_lazy_parameter_get_type(struct drgn_lazy_parameter *lazy
...
+	drgn_object_copy(ret, lazy_parameter->object);
+	*ret = *lazy_parameter->object;
```
osandov, line 92: `drgn_object_copy()` already effectively did this, so this shouldn't be necessary and isn't safe since the object may have an allocated buffer.

libdrgn/lazy_parameter.c
```diff
@@ -45,3 +74,21 @@ struct drgn_error *drgn_lazy_parameter_get_type(struct drgn_lazy_parameter *lazy
...
+						  struct drgn_object *ret) {
+	if (!drgn_lazy_parameter_is_evaluated(lazy_parameter)) {
+		struct drgn_object obj;
+		struct drgn_object_thunk *thunk_ptr = lazy_parameter->object_thunk;
+		struct drgn_object_thunk thunk = *thunk_ptr;
+		drgn_object_init(&obj, thunk.prog);
+		struct drgn_error *err = thunk.evaluate_fn(thunk_ptr, &obj);
+		if (err)
+			return err;
+		drgn_lazy_parameter_init_object(lazy_parameter, &obj);
```
osandov, line 87: Using `drgn_lazy_parameter_init_object()` means that you're creating a temporary object only to copy it and then discard it. I'd just `malloc()` the new object directly and reinitialize the lazy parameter here.

libdrgn/lazy_parameter.h
```diff
@@ -70,6 +96,31 @@ drgn_lazy_parameter_init_type(struct drgn_lazy_parameter *lazy_parameter,
...
+static inline struct drgn_error
+*drgn_lazy_parameter_init_object(struct drgn_lazy_parameter *lazy_object,
+				 struct drgn_object *object)
+{
+	// Freed by drgn_program_deinit_types
+	struct drgn_object *new_obj = malloc(sizeof(*new_obj));
+	if (!new_obj)
+		return &drgn_enomem;
+	drgn_object_init(new_obj, drgn_object_program(object));
+	drgn_object_copy(new_obj, object);
```
osandov, line 119: This can fail, you need to handle errors.

libdrgn/debug_info.c
```diff
@@ -1030,6 +1030,13 @@ struct drgn_type_from_dwarf_thunk {
 	uint64_t bias;
 };
 
+struct drgn_object_from_dwarf_thunk {
+	struct drgn_object_thunk thunk;
+	Dwarf_Die die;
+	uint64_t bias;
+	const char* name;
```
osandov, line 1037: `const char *name`

osandov (commented):
Didn't look at the final commit in much detail yet, but wanted to send you the comments on the lazy object handling.

PR #76, opened 2020-10-20
libkdumpfile API usage regression - fix kdump_vmcoreinfo_raw leak

### LINKED ISSUE #75: Recent libkdumpfile changes breaks drgn for kdumps
The latest successful SDB nightly was 2 days ago: https://github.com/delphix/sdb/actions/runs/307613188
Yesterday the nightly failed (https://github.com/delphix/sdb/runs/1265111694?check_suite_focus=true) because when `drgn` tried to open our regression kdump dump to run the tests we got back `VMCOREINFO does not contain valid OSRELEASE`.

Since it seems like libkdumpfile was the only repo that had changes during that time I've started looking into some of these changes that look suspicious. This could be a regression on the `drgn` kdump code if the API changed but it could also be a bug in the new libkdumpfile code. In any case I wanted to create this issue here for reference.

sdimitro:
As of b4b3c94cc24dc88d08701ebe35f72a1f1785046a , libkdumpfile
changed some of its API infrastructure and ended up breaking
`kdump_vmcoreinfo_raw` temporarily - The regression was later
fixed with fedbe7f66561754782165be2512bca0ed1b775e9 but the
call now allocates a buffer in `kdump_vmcoreinfo_raw` that
is later up to the caller to free it. This patch updates our
usage of this function so we free our vmcoreinfo string buffer
after we copy the contents that we need from it.

Closes #75.

Signed-off-by: Serapheim Dimitropoulos <serapheim@delphix.com>

### DIFF
--- a/libdrgn/kdump.c
+++ b/libdrgn/kdump.c
@@ -70,8 +70,8 @@ struct drgn_error *drgn_program_set_kdump(struct drgn_program *prog)
 	struct drgn_error *err;
 	kdump_ctx_t *ctx;
 	kdump_status ks;
-	const char *vmcoreinfo;
 	struct drgn_platform platform;
+	char *vmcoreinfo = NULL;
 
 	ctx = kdump_new();
 	if (!ctx) {
@@ -135,6 +135,7 @@ struct drgn_error *drgn_program_set_kdump(struct drgn_program *prog)
 	return NULL;
 
 err:
+	free(vmcoreinfo);
 	kdump_free(ctx);
 	return err;
 }


osandov:
Thanks for tracking this down! There is a problem -- on older versions of libkdumpfile, we shouldn't be doing this `free()`. It could crash or corrupt memory. Is there a version check we can do? If not, I think we have no choice but to leak it (we should at least add a comment).

osandov:
I emailed Petr. Hopefully he doesn't mind tagging a new release.

PR #79, opened 2021-01-08
Add basic support for parsing DWARF type units

jgkamat:
This ended up being more straightforward than expected - although there are still a couple rough edges (and no tests right now). The biggest change is removing all usages of offsets from the index. 

I'll highlight the biggest questionable things in comments below - hopefully I'll be able to add tests and clean this up a bit more next week :).

### DIFF
--- a/libdrgn/debug_info.c
+++ b/libdrgn/debug_info.c
@@ -2986,8 +2986,19 @@ drgn_type_from_dwarf_internal(struct drgn_debug_info *dbinfo,
 					 "maximum DWARF type parsing depth exceeded");
 	}
 
-	/* If we got a declaration, try to find the definition. */
+	/* If the DIE has a type unit signature, follow it. */
 	Dwarf_Die definition_die;
+	{
+		Dwarf_Attribute attr_mem, *attr;
+		if ((attr = dwarf_attr_integrate(die, DW_AT_signature,
+						 &attr_mem))) {
+			if (!dwarf_formref_die(attr, &definition_die))
+				return drgn_error_libdw();
+			die = &definition_die;
+		}
+	}
+
+	/* If we got a declaration, try to find the definition. */
 	bool declaration;
 	if (dwarf_flag(die, DW_AT_declaration, &declaration))
 		return drgn_error_libdw();
--- a/libdrgn/debug_info.c
+++ b/libdrgn/debug_info.c
@@ -97,6 +97,7 @@ static const char *dwarf_tag_str(Dwarf_Die *die, char buf[DW_TAG_BUF_LEN])
 
 static const char * const drgn_debug_scn_names[] = {
 	[DRGN_SCN_DEBUG_INFO] = ".debug_info",
+	[DRGN_SCN_DEBUG_TYPES] = ".debug_types",
 	[DRGN_SCN_DEBUG_ABBREV] = ".debug_abbrev",
 	[DRGN_SCN_DEBUG_STR] = ".debug_str",
 	[DRGN_SCN_DEBUG_LINE] = ".debug_line",
@@ -3002,9 +3003,19 @@ drgn_type_from_dwarf_internal(struct drgn_debug_info *dbinfo,
 				return drgn_error_libdwfl();
 			uintptr_t start =
 				(uintptr_t)module->scn_data[DRGN_SCN_DEBUG_INFO]->d_buf;
-			if (!dwarf_offdie(dwarf, die_addr - start,
-					  &definition_die))
-				return drgn_error_libdw();
+			size_t size =
+				module->scn_data[DRGN_SCN_DEBUG_INFO]->d_size;
+			if (die_addr >= start && die_addr < start + size) {
+				if (!dwarf_offdie(dwarf, die_addr - start,
+						  &definition_die))
+					return drgn_error_libdw();
+			} else {
+				start = (uintptr_t)module->scn_data[DRGN_SCN_DEBUG_TYPES]->d_buf;
+				/* Assume .debug_types */
+				if (!dwarf_offdie_types(dwarf, die_addr - start,
+							&definition_die))
+					return drgn_error_libdw();
+			}
 			die = &definition_die;
 		}
 	}
--- a/libdrgn/debug_info.c
+++ b/libdrgn/debug_info.c
@@ -2986,6 +2986,7 @@ drgn_type_from_dwarf_internal(struct drgn_debug_info *dbinfo,
 	}
 
 	/* If we got a declaration, try to find the definition. */
+	Dwarf_Die definition_die;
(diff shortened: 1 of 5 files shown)


libdrgn/debug_info.c
```diff
@@ -2127,6 +2128,19 @@ drgn_type_from_dwarf_internal(struct drgn_debug_info *dbinfo, Dwarf_Die *die,
 					 "maximum DWARF type parsing depth exceeded");
 	}
 
+	// Follow signature tags on die
```
jgkamat, line 2131: Right now the positioning of this is a little off - If the signature points to a incomplete declaration, it won't be resolved like normal. To fix that, the simplest solution is to do this translation when lazy objects are created (as I think that's where the declarations are resolved). However, since that area of code is likely changing a lot, I wanted to avoid touching it as much as possible so I put this here.

libdrgn/dwarf_index.c
```diff
@@ -583,6 +584,13 @@ static struct drgn_error *read_cu(struct drgn_dwarf_index_cu_buffer *buffer)
 					 &buffer->cu->address_size)))
 		return err;
 
+	// Skip rest of TU to properly place buffer pos
```
jgkamat, line 587: I don't like splitting this TU specific logic both here and in `index_cu_header_offset`, but I think this is the cleanest way to do it.

libdrgn/dwarf_index.c
```diff
@@ -1032,6 +1051,7 @@ void drgn_dwarf_index_read_module(struct drgn_dwarf_index_update_state *state,
 		{
 			struct drgn_dwarf_index_cu cu = {
 				.module = module,
+				.scn = scn,
```
jgkamat, line 1054: I hate to add fields to structs that are used a lot, but I think I will need some way to tell if a CU is a TU or not, and this is the simplest way I could think of doing that. If you have ideas for alternatives, let me know!

osandov (changes requested):
Thanks, this looks great overall!

I have a couple of requests for commit organization. First, it'd be more logical for "libdrgn: Switch to addresses for indexing into definition map" to come before "libdrgn: Add support for parsing type units", since without the former, the latter doesn't handle specifications correctly for `.debug_types`.  Second, "libdrgn: Add function to look up dies by addr" should be combined into "libdrgn: Add support for parsing type units", again because it's required for proper handling of `.debug_types`.

The commit messages are also missing some additional explanation and your Signed-off-by.

Other than that, a few small comments here and there, and like you, I'm not sure yet about where exactly is best to resolve the signatures.

libdrgn/debug_info.c
jgkamat: Ah, previously, the methods work added a find_definition as part of the lazy object port, as method members are frequently pointed to as incomplete types. Since (on that previous patch) this lookup is done on lazy type creation, it wasn't done in this case.

https://github.com/jgkamat/drgn/blob/07185513740e33d79d95420c26b6aad5de39c7c9/libdrgn/debug_info.c#L1249

 I'm not sure if this is an issue for templates on their own, but it is a concern here - especially on gcc, a signature tag often points to an incomplete type (which must use `find_definition` to resolve properly). However, right now this lookup (on master) only occurs on indexing, which leaves some types as incomplete. 

Would you prefer me to just use find_definition right here for now? I *think* this could also be moved to lazy type creation (along with a declaration lookup) - up to you.

jgkamat:
Also one more question about this, right now I'm not skipping over the skeleton declarations in dwarf_index first pass and second pass, mainly because I'm not sure how to get that flag in that code. Should I do something like DIE_FLAG_DECLARATION and set a flag in the instruction to follow, or is there another way to do it? 
That said, this does seem to work even without skipping over those blocks.

libdrgn/debug_info.c
osandov: Even without type units, I think you're right that if `drgn_type_from_dwarf_internal()` gets a DIE with `DW_AT_declaration`, it should try to find the definition. Would you mind opening a separate PR with just "libdrgn: Switch to addresses for indexing into definition map", and a commit on top of that adding the appropriate `find_definition()` to `drgn_type_from_dwarf_internal()`? I can merge that right away while we figure out the remaining nuances of signatures.

There's a minor question of whether to follow the declaration to its definition before or after checking the DWARF type cache (`drgn_debug_info::types`). The cache is a mapping from DIE address to the parsed `drgn_type`. Should we check the cache for the declaration first, then follow the declaration, and cache the type both for the declaration and the definition? This is faster for subsequent accesses to the declaration, since we won't need to follow the definition again, but it adds more entries to the cache.

For this PR, I think we do want the `DW_AT_signature` lookup you're doing here.

libdrgn/dwarf_index.c
```diff
@@ -1426,6 +1444,10 @@ index_cu_second_pass(struct drgn_dwarf_index_namespace *ns,
 		    !specification) {
 			if (insn & DIE_FLAG_DECLARATION)
 				declaration = true;
+			if (signature)
+				// TODO determine if this is needed - usually types with signature are also declarations
```
osandov, line 1448: From what I can tell, a DIE with `DW_AT_signature` but not `DW_AT_declaration` is still a definition, which is pointless at best and nonsense at worst. So, I don't think we need this.

libdrgn/debug_info.c
```diff
@@ -2443,8 +2444,17 @@ drgn_type_from_dwarf_internal(struct drgn_debug_info *dbinfo,
 			if (!dwarf)
 				return drgn_error_libdwfl();
 			uintptr_t start = (uintptr_t)module->scns[DRGN_SCN_DEBUG_INFO]->d_buf;
-			if (!dwarf_offdie(dwarf, die_addr - start, die))
-				return drgn_error_libdw();
+			uint64_t size = module->scns[DRGN_SCN_DEBUG_INFO]->d_size;
+			if (die_addr >= start &&
```
jgkamat, line 2461: I added a bit of duplication here, but I can make a helper in dwarf_index.h if you'd like to combine them.

tests/test_dwarf.py
```diff
@@ -4102,3 +4102,200 @@ def test_reference_counting_type_parameter(self):
...
+                    TypeMember(prog.int_type("int", 4, True), "y", 32),
+                ),
+            ),
+        )
+        self.assertIdentical(
+            prog.type("TEST").type,
+            prog.type("struct point"),
+        )
+
+    def test_type_unit_specification(self):
```
jgkamat, line 4229: This test actually dosen't do what I originally intended (test following specifications/declarations in type units) because it is also indexed and caught by find_complete. I might add another which more reliably tests it with nested classes or similar...

osandov (approved):

PR #87, opened 2021-01-11
WIP for adding key listing functionality to drgn

alastor-erinyes:
This is far from complete (it doesn't even compile yet), but I wanted to show where I was heading with this.

Primarily what I have so far is most of the libdrgn interface created and that interface at least partially hooked up to python.  The main logic for the listing operation as well has how it's handled in conversion to a frozenset and exposure to the user in the python code still needs to be implemented.

I'm considering using a linked list for accumulation of the keys in libdrgn, this would be easy to merge with items from multiple listing functions, then the deduplication occurs when adding to the frozenset in the Python c-api with a single iteration through the combined list.

### DIFF
--- a/_drgn.pyi
+++ b/_drgn.pyi
@@ -14,6 +14,7 @@ from typing import (
     Any,
     Callable,
     Dict,
+    FrozenSet,
     Iterable,
     Iterator,
     Mapping,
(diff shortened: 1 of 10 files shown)


alastor-erinyes:
I'm also considering just finishing hooking all the piping up in this PR and leaving the real libdrgn function logic for a subsequent PR that will be easier to read :)

osandov:
After giving it more thought, it occurred to me that long term, we're going to want a way to iterate not just over names, but over the objects themselves. And for C++ support, we're going to want to do that per namespace, not just for the whole program. I'm not sure yet how that will look, but I know it would require redoing a lot of this work.

So, I think we can keep it simple for this PR and implement a short term solution: drop the libdrgn API and have `Program.keys()` go straight to the DWARF index. Something like:

```c
PyObject *Program_keys(Program *self)
{
	struct drgn_error *err;
	PyObject *frozenset = ...;

	if (self->prog._dbinfo) {
		static const uint64_t tags[] = ...;
		struct drgn_dwarf_index_iterator it;
		err = drgn_dwarf_index_iterator_init(&it,
						     &self->prog._dbinfo->dindex.global,
						     NULL, 0, tags, 1);
		if (err)
			return err;
		/* iterate over dwarf index and populate frozenset */
	}
	return frozenset;
}
```

It's a bit of a layering violation, but there's no need to make it generic yet if it's going to be reworked in the near future. Sorry for sending you in the wrong direction!

osandov:
Closing for now since this will need a different approach.

PR #89, opened 2021-02-22
libdrgn: Fix lookup of indexed incomplete classes

jgkamat:
Previously, if a class, struct, or enum was both indexed and
incomplete, the check to see if the class was a declaration or
specification during type lookup would always return as a declaration.
This would cause the "complete type" to be looked up again, which
triggers a recursion loop.

One note - I applied this to enums too although I haven't tested that case - as it does the same declaration lookup.

### DIFF
--- a/libdrgn/debug_info.c
+++ b/libdrgn/debug_info.c
@@ -1144,6 +1144,18 @@ static int dwarf_flag(Dwarf_Die *die, unsigned int name, bool *ret)
 	Dwarf_Attribute attr_mem;
 	Dwarf_Attribute *attr;
 
+	if (!(attr = dwarf_attr(die, name, &attr_mem))) {
+		*ret = false;
+		return 0;
+	}
+	return dwarf_formflag(attr, ret);
+}
+
+static int dwarf_flag_integrate(Dwarf_Die *die, unsigned int name, bool *ret)
+{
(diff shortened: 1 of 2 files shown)


osandov:
Thanks, the fix looks correct. The commit message focuses on the indexing aspect, but the real bug is how we're checking `DW_AT_declaration`; the infinite loop with the index is a knock-on effect. How about changing it to something like this:

```
libdrgn: debug_info: fix parsing specifications of declarations

drgn_compound_type_from_dwarf() and drgn_enum_type_from_dwarf() check
the DW_AT_declaration flag to decide whether the type is a declaration
of an incomplete type or a definition of a complete type. However, they
check DW_AT_declaration with dwarf_attr_integrate(), which follows the
DW_AT_specification reference if it is present. The DIE referenced by
DW_AT_specification typically is a declaration, so this erroneously
identifies definitions as declarations. Additionally, if
drgn_debug_info_find_complete() finds the same definition, we can end up
recursing until we hit the DWARF parsing depth limit. Fix it by not
using dwarf_attr_integrate() for DW_AT_declaration.
```

osandov:
Oops sorry I didn't notice that you had already updated this a couple of days ago. Thanks for the fix!

PR #90, opened 2021-03-06
Convert offsets to addresses and add lookup for indirect dies in type lookup

jgkamat:
Continuation from https://github.com/osandov/drgn/pull/79#discussion_r554248032

I decided to leave the type cache for after the lookup,  as I don't think the extra lookup wll be too bad, and I think it will take a lot of space (and if it does seem slow, it should be an easy fix). This way seems simpler to me, so I think that's where we should start. 

The first commit should only have the minor changes requested, otherwise it should be the same.

### DIFF
--- a/libdrgn/debug_info.c
+++ b/libdrgn/debug_info.c
@@ -2433,6 +2433,26 @@ drgn_type_from_dwarf_internal(struct drgn_debug_info *dbinfo,
 					 "maximum DWARF type parsing depth exceeded");
 	}
 
+	/* If we got a declaration, try to find the definition. */
+	bool declaration;
+	if (dwarf_flag(die, DW_AT_declaration, &declaration))
+		return drgn_error_libdw();
+	if (declaration) {
+		uintptr_t die_addr;
+		if (drgn_dwarf_index_find_definition(&dbinfo->dindex,
+						     (uintptr_t)die->addr,
+						     &module, &die_addr)) {
+			Dwarf_Addr bias;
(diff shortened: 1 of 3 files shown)


libdrgn/debug_info.c
```diff
@@ -2433,6 +2433,21 @@ drgn_type_from_dwarf_internal(struct drgn_debug_info *dbinfo,
 					 "maximum DWARF type parsing depth exceeded");
 	}
 
+	/* If we get a declaration, try to follow it to a specification */
+	bool decl;
+	if (!dwarf_flag(die, DW_AT_declaration, &decl) && decl) {
```
osandov, line 2438: I think we should return `drgn_error_libdw()` if `dwarf_flag()` fails rather than ignoring it.

libdrgn/debug_info.c
```diff
@@ -2433,6 +2433,21 @@ drgn_type_from_dwarf_internal(struct drgn_debug_info *dbinfo,
...
+		size_t die_addr;
```
osandov, line 2439: This should be `uintptr_t` (it's probably the same as `size_t` on every sane platform, but better to be consistent with how `drgn_dwarf_index_find_definition()` is declared).

osandov (commented):
Thanks! Just two tiny comments.

osandov:
I cleaned up some line wrapping and clarified the comments/commit messages a bit. I'll merge it as soon as tests pass, thanks again!

PR #91, opened 2021-03-06
Add support for DW_FORM_indirect

jgkamat:
This isn't completely done, unfortunately, but I wanted to get it out in the open at least for now. The main thing missing is real tests, right now I'm using your patch to convert every test into an indirect one (so right now the tests don't test normal dwarf). Do you think we should just duplicate every test into a indirect and non-indirect variant? 

Most of this is unfortunately just moving code around, the main strategy is to record the "context" of an attribute when we see an indirect (in the instruction), and use that to re-create the "real" instruction when later parsing. To do this I (unfortunately) needed to make several additional instructions (to encode this context), but I don't know if there's a cleaner way to do this. 

Let me know what you think - hopefully I will have time to clean up the tests soon...

### DIFF
--- a/libdrgn/dwarf_index.c
+++ b/libdrgn/dwarf_index.c
@@ -231,6 +231,244 @@ void drgn_dwarf_index_update_cancel(struct drgn_dwarf_index_update_state *state,
 		state->err = err;
 }
 
+static struct drgn_error *dw_form_to_insn(struct drgn_dwarf_index_cu *cu,
+					  struct binary_buffer *bb,
+					  uint64_t form, uint8_t *insn_ret)
+{
+	switch (form) {
+	case DW_FORM_addr:
+		*insn_ret = cu->address_size;
+		return NULL;
+	case DW_FORM_data1:
+	case DW_FORM_ref1:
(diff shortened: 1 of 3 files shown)


osandov:
I had a cleanup lying around related to this, so I rebased this on that change and cleaned it up. I also just sent the required [elfutils fix](https://sourceware.org/pipermail/elfutils-devel/2021q2/003739.html) upstream. I'll probably wait to get some feedback on the elfutils patch before merging this, because this functionality won't be useful if elfutils can't handle DW_FORM_indirect anyways.

As for testing, we probably need to gate it behind an elfutils version check or something like that.

osandov:
To elaborate on testing: it's probably overkill to run every test with and without DW_FORM_indirect. We can pick a handful of tests that exercise the attributes used by the DWARF index and run those in both ways. But as I mentioned, we'll either need to check the runtime version of libdw or just catch errors caused by missing libdw support.

osandov:
The elfutils patch was merged upstream, so I went ahead and added a few test cases and merged this. Thanks!

PR #93, opened 2021-03-19
libdrgn: stretch minimum supported version of libelf to 0.170

sdimitro:
Currently libdrgn requires libelf to be of version 0.175 or
later. This patch allows the library to be compiled with libelf
0.170 (the newest version supported by Ubuntu 18.04 LTS).

Signed-off-by: Serapheim Dimitropoulos <serapheim@delphix.com>

### DIFF
--- a/libdrgn/debug_info.c
+++ b/libdrgn/debug_info.c
@@ -7,6 +7,7 @@
 #include <elfutils/known-dwarf.h>
 #include <elfutils/libdw.h>
 #include <elfutils/libdwelf.h>
+#include <elfutils/version.h>
 #include <errno.h>
 #include <fcntl.h>
 #include <gelf.h>
@@ -2466,6 +2467,17 @@ err:
 	drgn_compound_type_builder_deinit(&builder);
 	return err;
 }
+
+#if !_ELFUTILS_PREREQ(0, 171)
+#define DW_FORM_implicit_const 0x21
+#endif
+
+#if !_ELFUTILS_PREREQ(0, 175)
+static Elf *dwelf_elf_begin(int fd)
+{
+	return elf_begin(fd, ELF_C_READ_MMAP_PRIVATE, NULL);
+}
+#endif
 
 static struct drgn_error *
 parse_enumerator(Dwarf_Die *die, struct drgn_enum_type_builder *builder,
--- a/libdrgn/language.c
+++ b/libdrgn/language.c
@@ -67,7 +67,7 @@ struct drgn_error *drgn_language_from_die(Dwarf_Die *die, bool fall_back,
 					  const struct drgn_language **ret)
 {
 	Dwarf_Die cudie;
-	if (dwarf_cu_info(die->cu, NULL, NULL, &cudie, NULL, NULL, NULL, NULL))
+	if (!dwarf_cu_die(die->cu, &cudie, NULL, NULL, NULL, NULL, NULL, NULL))
 		return drgn_error_libdw();
 	switch (dwarf_srclang(&cudie)) {
 	case DW_LANG_C:
--- a/libdrgn/program.c
+++ b/libdrgn/program.c
@@ -6,6 +6,7 @@
 #include <dwarf.h>
 #include <elf.h>
 #include <elfutils/libdw.h>
+#include <elfutils/version.h>
 #include <errno.h>
 #include <fcntl.h>
 #include <gelf.h>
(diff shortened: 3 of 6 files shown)


libdrgn/language.c
```diff
@@ -67,7 +67,8 @@ struct drgn_error *drgn_language_from_die(Dwarf_Die *die, bool fall_back,
 					  const struct drgn_language **ret)
 {
 	Dwarf_Die cudie;
-	if (dwarf_cu_info(die->cu, NULL, NULL, &cudie, NULL, NULL, NULL, NULL))
+	// if (dwarf_cu_info(die->cu, NULL, NULL, &cudie, NULL, NULL, NULL, NULL))
+	if (!dwarf_cu_die(die->cu, &cudie, NULL, NULL, NULL, NULL, NULL, NULL))
```
osandov, line 71: From what I can tell, `dwarf_cu_info()` and `dwarf_cu_die()` return the same CU DIE, so we can probably make this change unconditionally, right?

libdrgn/debug_info.c
```diff
@@ -4324,7 +4325,8 @@ struct drgn_error *open_elf_file(const char *path, int *fd_ret, Elf **elf_ret)
 	*fd_ret = open(path, O_RDONLY);
 	if (*fd_ret == -1)
 		return drgn_error_create_os("open", errno, path);
-	*elf_ret = dwelf_elf_begin(*fd_ret);
+	// *elf_ret = dwelf_elf_begin(*fd_ret);
+	*elf_ret = elf_begin(*fd_ret, ELF_C_READ_MMAP_PRIVATE, NULL);
```
osandov, line 4329: Looks like `dwelf_elf_begin()` was added in 0.175. We can probably define a stub like:

```c
#if !_ELFUTILS_PREREQ(0, 175)
static Elf *dwelf_elf_begin(int fd)
{
	return elf_begin(fd, ELF_C_READ_MMAP_PRIVATE, NULL);
}
#endif
```

(With a comment about what the real `dwelf_elf_begin()` does that `elf_begin()` doesn't.)

This might need an `#include <elfutils/version.h>`. It can probably go in `libdrgn/debug_info.c`. A little ugly but not the worst.

libdrgn/program.c
```diff
@@ -33,8 +33,8 @@ DEFINE_HASH_TABLE_FUNCTIONS(drgn_prstatus_map, int_key_hash_pair, scalar_key_eq)
 
 static Elf_Type note_header_type(GElf_Phdr *phdr)
 {
-	if (phdr->p_align == 8)
-		return ELF_T_NHDR8;
+	// if (phdr->p_align == 8)
+	// 	return ELF_T_NHDR8;
```
osandov, line 37: I suppose an `#if` here wouldn't be too bad, either (again, elfutils 0.175):

```c
static Elf_Type note_header_type(GElf_Phdr *phdr)
{
#if _ELFUTILS_PREREQ(0, 175)
	if (phdr->p_align == 8)
		return ELF_T_NHDR8;
#endif
	return ELF_T_NHDR;
}
```

libdrgn/debug_info.c
```diff
@@ -2483,8 +2483,9 @@ parse_enumerator(Dwarf_Die *die, struct drgn_enum_type_builder *builder,
 					 "DW_TAG_enumerator is missing DW_AT_const_value");
 	}
 	struct drgn_error *err;
-	if (attr->form == DW_FORM_sdata ||
-	    attr->form == DW_FORM_implicit_const) {
+	// if (attr->form == DW_FORM_sdata ||
+	//     attr->form == DW_FORM_implicit_const) {
+	if (attr->form == DW_FORM_sdata) {
```
osandov, line 2488: This one I'm not as sure about. It feels wrong to ignore newer DWARF types even if the library doesn't support them. Maybe we should include copies of the latest `dwarf.h` and `known-dwarf.h` for drgn to use (since those just define constants) instead of the system ones?

osandov (commented):
I left my thoughts on each of the necessary changes. Let me know what you think.

libdrgn/debug_info.c
sdimitro: I agree that we should support newer DWARF types if we can.

As for the proposed solution, we don't have a lot of alternatives right? Maybe instead of a full copy of the header files we could only define that one constant for now? That said having the whole file may make more sense - we'd need to care less about whether new code may break old libelf users and pay closer attention to what constants drgn needs in general.

osandov: Sure, for now you can just do

```c
#if !_ELFUTILS_PREREQ(0, 171)
#define DW_FORM_implicit_const 0x21
#endif
```

Probably put that next to the `dwelf_elf_begin()` stub. If we end up needing more of these, we can add the copy at that point.

libdrgn/language.c
```diff
@@ -63,11 +63,12 @@ const struct drgn_language drgn_languages[] = {
 	},
 };
 
+
```
osandov, line 66: Stray newline.

osandov (commented):
We also need to update the documentation in a couple of places. I'll do that and fix the stray newline and merge, thanks!

osandov:
I just pushed e5bc41f16c4886e67c242deb17e72aa0180c63f2 which extends this even further to 0.165 by adding local copies of `elf.h` and `dwarf.h`.

PR #100, opened 2021-04-06
add COPYING to sdist

davide125:


osandov:
Looks like the bleeding edge version of setuptools does this by default (pypa/setuptools@749b97499), but even Arch Linux doesn't have that quite yet so it makes sense to do this manually, thanks.

PR #107, opened 2021-07-15
helpers: add kernel nodemask helpers

Arch-Zheng:
Sometimes we want to traverse numa nodes in the system,
so add kernel nodemask helpers to support this.

Signed-off-by: Qi Zheng <zhengqi.arch@bytedance.com>

### DIFF
--- a/drgn/helpers/linux/nodemask.py
+++ b/drgn/helpers/linux/nodemask.py
@@ -1,0 +1,55 @@
+# Copyright (c) ByteDance, Inc. and its affiliates.
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+"""
+NUMA Node Masks
+---------------
+
+The ``drgn.helpers.linux.nodemask`` module provides helpers for working with
+NUMA node masks from :linux:`include/linux/nodemask.h`.
+"""
+
+from typing import Iterator
+
+from drgn import IntegerLike, Object, Program
+from drgn.helpers.linux.cpumask import _for_each_set_bit
+
+__all__ = (
+    "for_each_node",
+    "for_each_node_mask",
+    "for_each_node_state",
+    "for_each_online_node",
+)
+
+
+def for_each_node_mask(mask: Object) -> Iterator[int]:
+    """
+    Iterate over all of the NUMA nodes in the given mask.
+
+    :param mask: ``nodemask_t``
+    """
+    try:
+        nr_node_ids = mask.prog_["nr_node_ids"].value_()
+    except KeyError:
+        nr_node_ids = 1
+    return _for_each_set_bit(mask.bits, nr_node_ids)
+
+
+def for_each_node_state(prog: Program, state: IntegerLike) -> Iterator[int]:
+    """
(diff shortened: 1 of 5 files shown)


osandov (changes requested):
Sorry for the delay, I just got back from vacation. Thanks for this! I have a few requests, mostly nitpicks. Other than that, it'd also be nice to have some test cases similar to the ones in `tests/helpers/linux/test_cpumask.py`. It looks like you can use the `online`, `possible`, `memory`, etc. files under `/sys/devices/system/node`. They use the same format as the CPU mask lists.

drgn/helpers/linux/nodemask.py
```diff
@@ -0,0 +1,48 @@
...
+from typing import Iterator
+
+from drgn import Object, Program, sizeof
+
+__all__ = (
+    "for_each_node",
+    "for_each_online_node",
+)
+
+def for_each_node_mask(mask: Object) -> Iterator[int]:
```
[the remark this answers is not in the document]
Arch-Zheng: Maybe we can add a ``drgn.helpers.linux.bitops`` module to support bitmap-related operation?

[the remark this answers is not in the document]
osandov: I think that's a good idea for a follow-up PR. For this PR we can just keep it as an internal helper.

drgn/helpers/linux/nodemask.py
```diff
@@ -0,0 +1,51 @@
+# Copyright (c) ByteDance, Inc. and its affiliates.
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+"""
+NUMA Node Masks
+---------
```
osandov, line 6: Nit: these hyphens should go to the end of the title above.

tests/helpers/linux/test_nodemask.py
```diff
@@ -0,0 +1,35 @@
...
+    nodes = set()
+    for node_range in nodelist.split(","):
+        first, sep, last = node_range.partition("-")
+        if sep:
+            nodes.update(range(int(first), int(last) + 1))
+        else:
+            nodes.add(int(first))
+    return nodes
+
+class TestNodeMask(LinuxHelperTestCase):
```
osandov, line 24: Looks like this needs a `@unittest.skipUnless(os.path.exists("/sys/devices/system/node"), "kernel does not support NUMA")`.

osandov (changes requested):

tests/helpers/linux/test_nodemask.py
Arch-Zheng: Yes, I will add this and resend the patch. BTW, Is NUMA node not enabled on your CI test machine?

osandov: No, it uses a minimal kernel configuration so it appears that it left out NUMA. I'll enable it for future builds, but we should still have this check.

osandov (approved):
There were a few isort and black errors. I went ahead and cleaned them up. I'll merge this once tests pass, thanks again!

PR #110, opened 2021-08-11
helpers: Add netdev_get_by_index()

ypl-coffee:
Add a helper to find the "struct net_device *" object given the
corresponding interface index.  As an example:

	>>> netdev = find_netdev_by_index(prog.object("init_net"), 1)
	>>> netdev.name.string_().decode()
	'lo'

Currently looking up by name is non-trivial to do in drgn, see kernel
function net/core/dev.c:netdev_name_node_lookup() for reference.  Use
socket.if_nametoindex() instead:

	>>> dummy = find_netdev_by_index(prog.object("init_net"), \
	...                              socket.if_nametoindex("dummy0"))
	>>> dummy.name.string_().decode()
	'dummy0'

### DIFF
--- a/drgn/helpers/linux/net.py
+++ b/drgn/helpers/linux/net.py
@@ -9,16 +9,49 @@ The ``drgn.helpers.linux.net`` module provides helpers for working with the
 Linux kernel networking subsystem.
 """
 
-from typing import Iterator
+import operator
+from typing import Iterator, Union
 
-from drgn import Object
+from drgn import NULL, IntegerLike, Object, Program
+from drgn.helpers.linux.list import hlist_for_each_entry
 from drgn.helpers.linux.list_nulls import hlist_nulls_for_each_entry
 from drgn.helpers.linux.tcp import sk_tcpstate
 
 __all__ = (
+    "netdev_get_by_index",
     "sk_fullsock",
     "sk_nulls_for_each",
 )
+
+
+_NETDEV_HASHBITS = 8
+_NETDEV_HASHENTRIES = 1 << _NETDEV_HASHBITS
+
+
+def netdev_get_by_index(
+    prog_or_net: Union[Program, Object], ifindex: IntegerLike
+) -> Object:
+    """
+    Get the network device with the given interface index number.
+
+    :param prog_or_net: ``struct net *`` containing the device, or
+        :class:`Program` to use the initial network namespace.
+    :param ifindex: Network interface index number.
+    :return: ``struct net_device *`` (``NULL`` if not found)
+    """
+    if isinstance(prog_or_net, Program):
+        prog_or_net = prog_or_net["init_net"]
+    if isinstance(ifindex, Object):
+        ifindex = ifindex.read_()
+
+    head = prog_or_net.dev_index_head[
+        operator.index(ifindex) & (_NETDEV_HASHENTRIES - 1)
+    ]
+    for netdev in hlist_for_each_entry("struct net_device", head, "index_hlist"):
+        if netdev.ifindex == ifindex:
+            return netdev
+
+    return NULL(prog_or_net.prog_, "struct net_device *")
 
 
 def sk_fullsock(sk: Object) -> bool:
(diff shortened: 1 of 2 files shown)


drgn/helpers/linux/net.py
```diff
@@ -45,3 +43,25 @@ def sk_nulls_for_each(head: Object) -> Iterator[Object]:
         "struct sock", head, "__sk_common.skc_nulls_node"
     ):
         yield sk
+
+
+_NETDEV_HASHBITS = 8
+_NETDEV_HASHENTRIES = 1 << _NETDEV_HASHBITS
+
+
+def find_netdev_by_index(net: Object, ifindex: int) -> Object:
```
osandov, line 52: The equivalent kernel function is called `dev_get_by_index()`, so we should either give it the same name or maybe name it `netdev_get_by_index()`.

drgn/helpers/linux/net.py
```diff
@@ -45,3 +43,25 @@ def sk_nulls_for_each(head: Object) -> Iterator[Object]:
...
+    """
+    Return the ``struct net_device`` object of the given interface index
+    number, or ``NULL`` if not found.
+
+    :param net: ``struct net *``
```
osandov, line 57: We should also allow the first argument to be `prog` and automatically translate that to `init_net`. See [`path_lookup`](https://github.com/osandov/drgn/blob/51f63bb53b2fe8adb5b685f9215773aaaf7fb55b/drgn/helpers/linux/fs.py#L102).

drgn/helpers/linux/net.py
osandov, line 52: `ifindex` should be `drgn.IntegerLike` instead of `int`, because it's also valid to pass an integer-type `drgn.Object`.

osandov (changes requested):
Thanks! drgn definitely needs more networking helpers, but being a filesystem developer, I have a blind spot there. Regarding name lookups, we can just do a linear search over the hash table. It'll be relatively slow, but still useful. Also, thank you for adding a test case! I have a few minor comments.

drgn/helpers/linux/net.py
```diff
@@ -45,3 +44,33 @@ def sk_nulls_for_each(head: Object) -> Iterator[Object]:
...
+                        :class:`Program` to use the initial network namespace
+    :param ifindex: Network interface index number of the target device
+    :return: ``struct net_device *``
+    """
+    if isinstance(prog_or_net, Program):
+        prog_or_net = prog_or_net["init_net"]
+
+    head = prog_or_net.dev_index_head[
+        operator.index(ifindex) & (_NETDEV_HASHENTRIES - 1)
+    ]
```
osandov, line 70: I don't think we need to call `operator.index()` here; a `drgn.Object` can be indexed by another `drgn.Object`. But, we probably want to read the ifindex once (otherwise, if it is a reference, we will read it from `/proc/kcore` or the core dump on every comparison):

```suggestion
    if isinstance(ifindex, Object):
        ifindex = ifindex.read_()

    head = prog_or_net.dev_index_head[ifindex & (_NETDEV_HASHENTRIES - 1)]
```

ypl-coffee: Ah, I see; I thought `&` doesn't work on `IntegerLike`.

osandov:
You're right, `&` works for `Object`, but `IntegerLike` is more generic, so mypy doesn't like it. I'll add it back along with a couple of cosmetic changes and merge it once tests pass, thanks!

Would you like to submit a separate PR for `netdev_get_by_name()`? If not, I can add that.

osandov (approved):

ypl-coffee:
> You're right, `&` works for `Object`, but `IntegerLike` is more generic, so mypy doesn't like it. I'll add it back along with a couple of cosmetic changes and merge it once tests pass, thanks!

I see, thanks for the confirmation!

> Would you like to submit a separate PR for `netdev_get_by_name()`? If not, I can add that.

Yeah sure, I will do that soon later.

PR #111, opened 2021-08-11
helpers: Add netdev_get_by_name()

ypl-coffee:
Hi Omar,

Here's netdev_get_by_name() and a test, as we discussed.  Currently the `name` parameter is `str` since I didn't find the `str` version of `IntegerLike`; should we use something like `Path`?

---
Add a helper to get the network device ("struct net_device *") given an
interface name.  As an example:

    >>> netdev = netdev_get_by_name(prog["init_net"], "lo")
    >>> netdev.ifindex.value_()
    1

Or pass a "Program" as the first argument, and let the helper find in
the initial network namespace (i.e. "init_net"):

    >>> netdev = netdev_get_by_index(prog, "dummy0")
    >>> netdev.ifindex.value_()
    2

This helper simply does a linear search over the name hash table of the
network namespace, since implementing hashing in drgn is non-trivial.
It is obviously slower than net/core/dev.c:netdev_name_node_lookup() in
the kernel, but still useful.

### DIFF
--- a/drgn/helpers/linux/net.py
+++ b/drgn/helpers/linux/net.py
@@ -18,6 +18,7 @@ from drgn.helpers.linux.list_nulls import hlist_nulls_for_each_entry
 
 __all__ = (
     "netdev_get_by_index",
+    "netdev_get_by_name",
     "sk_fullsock",
     "sk_nulls_for_each",
 )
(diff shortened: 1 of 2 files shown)


drgn/helpers/linux/net.py
```diff
@@ -53,6 +54,27 @@ def netdev_get_by_index(
     return NULL(prog_or_net.prog_, "struct net_device *")
 
 
+def netdev_get_by_name(prog_or_net: Union[Program, Object], name: str) -> Object:
```
osandov, line 57:
```suggestion
def netdev_get_by_name(prog_or_net: Union[Program, Object], name: Union[str, bytes]) -> Object:
```

drgn/helpers/linux/net.py
```diff
@@ -53,6 +54,27 @@ def netdev_get_by_index(
...
+    Get the network device with the given interface name.
+
+    :param prog_or_net: ``struct net *`` containing the device, or
+        :class:`Program` to use the initial network namespace.
+    :param name: Network interface name.
+    :return: ``struct net_device *`` (``NULL`` if not found)
+    """
+    if isinstance(prog_or_net, Program):
+        prog_or_net = prog_or_net["init_net"]
+
```
osandov, line 68:
```suggestion
if isinstance(name, str):
    name = name.encode()

```

drgn/helpers/linux/net.py
```diff
@@ -53,6 +54,27 @@ def netdev_get_by_index(
...
+    for i in range(_NETDEV_HASHENTRIES):
+        head = prog_or_net.dev_name_head[i]
+        for name_node in hlist_for_each_entry("struct netdev_name_node", head, "hlist"):
+            if name_node.name.string_().decode() == name:
```
osandov, line 72:
```suggestion
            if name_node.name.string_() == name:
```

ypl-coffee:
Oops, let me see what is happening to `struct netdev_name_node`...

osandov (changes requested):
Thanks, this looks great! There's no "StringLike" in drgn, but we should make this support both `str` and `bytes`.

osandov:
> Oops, let me see what is happening to `struct netdev_name_node`...

Ah, it must've changed in between 5.4 and 5.10.

osandov:
I'd do something like:

```python
try:
    entry_type = prog.type("struct netdev_name_node")
    is_name_node = True
except LookupError:
    entry_type = prog.type("struct net_device")
    is_name_node = False
...
for entry in hlist_for_each_entry(entry_type, head, "hlist"):
    if entry.name.string_() == name:
        if is_name_node:
            return entry.dev
        else:
            return entry
```

I just pushed 980d1c64188211b7b4ca57c440637aa345a15921 so mypy won't complain about that.

ypl-coffee:
Hi Omar,

Thanks for the hint, turns out that the `hlist_for_each_entry()` `member` name also changed since that commit.  CI is happy on my fork (I will run it myself first before future PR's).

Added comment for the compat issue as other files under `helpers/linux` do; reworded commit log.

Please fix any style issues as you see fit.  Thanks!

osandov (approved):
I added the major kernel version containing the commit and reworded a bit. Thanks again!

PR #113, opened 2021-08-16
helpers: Add for_each_netns()

ypl-coffee:
Hi Omar,

Another simple helper to iterate over all `struct net`s.  Looking up by name is impractical since Linux kernel doesn't maintain name for netns.  One way would be let drgn read the corresponding nsfs node directly e.g. `/var/run/netns/foo`, but different userspace tools bind-mount `/proc/$PID/ns/net` to different places (e.g. `/var/run/netns/*` for `ip`, `/var/run/docker/*` for Docker)...I think a `for_each_` helper would be sufficient for now.

thanks,
ypl

---
Add a helper to iterate over all network namespaces in the system.  As
an example:

	>>> for net in for_each_netns(prog):
	...     if netdev_get_by_name(net, "enp0s3"):
	...         print(net.ipv4.sysctl_ip_early_demux.value_())
	...
	1

### DIFF
--- a/drgn/helpers/linux/net.py
+++ b/drgn/helpers/linux/net.py
@@ -13,15 +13,28 @@ import operator
 from typing import Iterator, Union
 
 from drgn import NULL, IntegerLike, Object, Program
-from drgn.helpers.linux.list import hlist_for_each_entry
+from drgn.helpers.linux.list import hlist_for_each_entry, list_for_each_entry
 from drgn.helpers.linux.list_nulls import hlist_nulls_for_each_entry
 
 __all__ = (
+    "for_each_net",
     "netdev_get_by_index",
     "netdev_get_by_name",
     "sk_fullsock",
     "sk_nulls_for_each",
 )
+
+
+def for_each_net(prog: Program) -> Iterator[Object]:
+    """
+    Iterate over all network namespaces in the system.
+
+    :return: Iterator of ``struct net *`` objects.
+    """
+    for net in list_for_each_entry(
+        "struct net", prog["net_namespace_list"].address_of_(), "list"
+    ):
+        yield net
 
 
 _NETDEV_HASHBITS = 8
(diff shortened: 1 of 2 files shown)


drgn/helpers/linux/net.py
```diff
@@ -13,17 +13,30 @@
...
 __all__ = (
+    "for_each_netns",
     "netdev_get_by_index",
     "netdev_get_by_name",
     "sk_fullsock",
     "sk_nulls_for_each",
 )
 
 
+def for_each_netns(prog: Program) -> Iterator[Object]:
```
osandov, line 28: It looks like in the kernel, this is called `for_each_net`. I try to stick to the kernel naming (except when it's confusing or too generic, like `dev_get_by_{index,name}`), so would you mind renaming it to `for_each_net`?

tests/helpers/linux/test_net.py
```diff
@@ -27,3 +32,11 @@ def test_netdev_get_by_name(self):
...
             netdev = netdev_get_by_name(self.prog, name)
             self.assertEqual(netdev.ifindex, index)
+
+    def test_for_each_netns(self):
+        self.assertTrue(
+            any(
+                netns.value_() == self.prog["init_net"].address_
+                for netns in for_each_netns(self.prog)
+            )
+        )
```
osandov, line 42: This can be written more succinctly as:

```suggestion
        self.assertIn(self.prog["init_net"].address_of_(), for_each_netns(self.prog))
```

osandov (changes requested):
Very nice, thank you. Just a couple of nitpicks.

tests/helpers/linux/test_net.py
ypl-coffee: Wow, that's so much cleaner, thanks

osandov (approved):
I updated the commit message to say `for_each_net` as well. This looks ready to go, thanks!

PR #114, opened 2021-08-18
helpers: Add get_net_ns_by_{inode,fd}()

ypl-coffee:
Hi Omar,

This is an "RFC" PR to add yet another helper to get `struct net` from a netns NSFS `struct file`.

As I mentioned earlier, there's no easy way to lookup netns by name, since the kernel simply doesn't maintain netns names.  So, I came up with this solution, reading the corresponding NSFS file (`/var/run/netns/$NAME`, `/var/run/docker/netns/$HASH`, etc.) directly.

AFAICT this is only useful when working on the current kernel, but I think it does provide an easy way to e.g. get `net` for a container.  What do you think?

About naming: see kernel `net/core/net_namespace.c`:`get_net_ns_by_*()` functions.

thanks,
ypl

---
Get the network namespace ("netns") descriptor ("struct net *") given an
netns NSFS pseudo-file ("struct file *") e.g. "/proc/$PID/ns/net" or
"/var/run/netns/$NAME".  As an example:

	>>> import os
	>>> task = find_task(prog, os.getpid())
	>>> file = open("/var/run/netns/foo")
	>>> net = get_net_ns_by_file(fget(task, file.fileno()))
	>>> netdev = netdev_get_by_name(net, "eth3")
	>>> netdev.ifindex.value_()
	5

"foo" was created by "ip netns add foo".  Similarly, Docker NSFS
files can be found under "/var/run/docker/netns/".

Obviously reading NSFS files on the fly may not work if we are not
working on the currently running kernel, but when we _are_, this should
still serve as an easy way to look up netns "by name".

The "file" parameter must point to a valid netns (i.e. of type
CLONE_NEWNET) NSFS file.

### DIFF
--- a/drgn/helpers/linux/net.py
+++ b/drgn/helpers/linux/net.py
@@ -12,12 +12,15 @@ Linux kernel networking subsystem.
 import operator
 from typing import Iterator, Union
 
-from drgn import NULL, IntegerLike, Object, Program
+from drgn import NULL, IntegerLike, Object, Program, cast, container_of
+from drgn.helpers.linux.fs import fget
 from drgn.helpers.linux.list import hlist_for_each_entry, list_for_each_entry
 from drgn.helpers.linux.list_nulls import hlist_nulls_for_each_entry
 
 __all__ = (
     "for_each_net",
+    "get_net_ns_by_inode",
+    "get_net_ns_by_fd",
     "netdev_get_by_index",
     "netdev_get_by_name",
     "sk_fullsock",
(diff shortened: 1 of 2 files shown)


drgn/helpers/linux/net.py
```diff
@@ -37,6 +39,27 @@ def for_each_net(prog: Program) -> Iterator[Object]:
...
+def get_net_ns_by_file(file: Object) -> Object:
+    """
+    Get the network namespace corresponding to the given network namespace (NSFS)
+    pseudo-file, e.g. '/proc/$PID/ns/net' or '/var/run/netns/$NAME'.
+
+    :param file: ``struct file *`` network namespace pseudo-file.
+    :return: ``struct net *`` (``NULL`` if invalid file)
+    """
+    if file.f_op != file.prog_["ns_file_operations"].address_of_():
+        return NULL(file.prog_, "struct net *")
```
osandov, line 54: I think we should raise an exception instead. Maybe:

```suggestion
        raise ValueError("not a net namespace inode")
```

drgn/helpers/linux/net.py
```diff
@@ -37,6 +39,27 @@ def for_each_net(prog: Program) -> Iterator[Object]:
...
+
+    ns = cast("struct ns_common *", file.f_inode.i_private)
+    if ns.ops.type != _CLONE_NEWNET:
+        return NULL(file.prog_, "struct net *")
```
osandov, line 58: Same here.

osandov (changes requested):
Thanks for this! This doesn't have to be limited to live kernel: with `path_lookup()`, you can get `struct path` from a file path, then get the inode from that:

```
[root@vmtest /]# ip netns add test
[root@vmtest /]# drgn
>>> inode = path_lookup(prog, "/run/netns/test").dentry.d_inode                                                                                                                                                                                                                                                                
>>> inode.i_fop
*(const struct file_operations *)(ns_file_operations+0x0 = 0xffffffff9e218380) = { ... }
```
(I had to use `/run/netns` instead of `/var/run/netns` because on my system, `/var/run` is a symlink to `/run` and `path_lookup()` doesn't know how to resolve symlinks.)

If this becomes `get_net_ns_by_inode()` and takes a `struct inode *` instead, we can use it with both `fget(...).f_inode` and `path_lookup(...).dentry.d_inode`.

For convenience, it might also be useful to have a `get_net_ns_by_fd(task, fd)` wrapper (in the same commit would be fine).

ypl-coffee:
> (I had to use `/run/netns` instead of `/var/run/netns` because on my system, `/var/run` is a symlink to `/run` and `path_lookup()` doesn't know how to resolve symlinks.)

Thanks a lot for the hint, now I see... :-)

osandov (approved):
I edited the comments a bit and changed the test case to use `get_net_ns_by_fd()` so that we'll get coverage of the wrapper, too. Thanks!

PR #116, opened 2021-08-24
libdrgn: reads from segment mapping

chenyingk:
We added an mmap interface for kcore (https://lore.kernel.org/patchwork/patch/1453867/).
It can reduce system calls significantly when accessing the kernel data structures.
The patch will be merged into next kernel version.

Co-developed-by:Feng Zhou <zhoufeng.zf@bytedance.com>
Signed-off-by: chenying <chenying.kernel@bytedance.com>

### DIFF
--- a/libdrgn/memory_reader.c
+++ b/libdrgn/memory_reader.c
@@ -284,3 +284,33 @@ struct drgn_error *drgn_read_memory_file(void *buf, uint64_t address,
 	memset(p, 0, count);
 	return NULL;
 }
+
+struct drgn_error *drgn_read_memory_mmap(void *buf, uint64_t address,
+					 size_t count, uint64_t offset,
+					 void *arg, bool physical)
+{
+	struct drgn_memory_file_segment *file_segment = arg;
+	char *p = buf;
+	char *map_address = file_segment->map_address;
+	size_t file_count;
+
+	if (!map_address)
+		return drgn_read_memory_file(buf, address, count,
+					     offset, arg, physical);
+
+	if (offset < file_segment->file_size) {
+		file_count = min((uint64_t)count,
+				 file_segment->file_size - offset);
+		count -= file_count;
+	} else {
+		file_count = 0;
+	}
+
+	if (file_count) {
+		memcpy(p, map_address + offset, file_count);
+		p += file_count;
+	}
+	memset(p, 0, count);
+
+	return NULL;
+}
--- a/libdrgn/program.c
+++ b/libdrgn/program.c
@@ -16,6 +16,7 @@
 #include <string.h>
 #include <sys/statfs.h>
 #include <unistd.h>
+#include <sys/mman.h>
 
 #include "array.h"
 #include "debug_info.h"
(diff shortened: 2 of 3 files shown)


libdrgn/memory_reader.c
```diff
@@ -284,3 +284,33 @@ struct drgn_error *drgn_read_memory_file(void *buf, uint64_t address,
...
+struct drgn_error *drgn_read_memory_mmap(void *buf, uint64_t address,
+					 size_t count, uint64_t offset,
+					 void *arg, bool physical)
+{
+	struct drgn_memory_file_segment *file_segment = arg;
+	char *p = buf;
+	char *map_address = file_segment->map_address;
+	size_t file_count;
+
+	if (!map_address)
```
osandov, line 297: We only use this callback if we successfully mapped this segment, right? So we should either remove this check or get rid of `drgn_read_memory_mmap()` and have `drgn_read_memory_file()` handle both the mmap and read cases.

libdrgn/program.c
```diff
@@ -16,6 +16,7 @@
 #include <string.h>
 #include <sys/statfs.h>
 #include <unistd.h>
+#include <sys/mman.h>
```
osandov, line 19: Alphabetical order for this include, please.

```suggestion
#include <string.h>
#include <sys/mman.h>
#include <sys/statfs.h>
#include <unistd.h>
```

libdrgn/program.c
```diff
@@ -86,6 +87,37 @@ void drgn_program_init(struct drgn_program *prog,
...
 }
 
+static int drgn_program_segments_deinit(struct drgn_program *prog)
+{
+	GElf_Phdr phdr_mem, *phdr;
+	size_t phnum, i, j;
+	uint64_t file_size;
+	void *map_address;
+
+	if (elf_getphdrnum(prog->core, &phnum) != 0)
```
osandov, line 97: Rather than digging through the core file again, let's just save `num_file_segments` in `struct drgn_program`.

libdrgn/program.c
```diff
@@ -381,9 +416,18 @@ drgn_program_set_core_dump(struct drgn_program *prog, const char *path)
 		prog->file_segments[j].file_size = phdr->p_filesz;
 		prog->file_segments[j].fd = prog->core_fd;
 		prog->file_segments[j].eio_is_fault = false;
+		prog->file_segments[j].map_address = NULL;
+
+		map_address = mmap(NULL, phdr->p_filesz, PROT_READ,
+				MAP_PRIVATE, prog->core_fd, phdr->p_offset);
+		if (map_address && map_address != MAP_FAILED) {
```
osandov, line 423: According to [`mmap(2)`](https://man7.org/linux/man-pages/man2/mmap.2.html), mmap won't return `NULL` unless `MAP_FIXED` is given.

```suggestion
		if (map_address != MAP_FAILED) {
```

osandov (changes requested):
Sorry for the delay, I was on vacation last week and am catching up.

This is amazing! I wasn't sure that it'd be possible to implement `mmap` for `/proc/kcore` in the kernel.

I have a few small comments on the code, and a couple of more general questions:
1. How well will this work for normal core dump files? My concern is that core dump segments might not be always be aligned to the page size, but at least on the core dumps that I have lying around, they are properly aligned.
2. Should we skip the mmap path on 32-bit systems? Virtual address space is much tighter there, but I suppose that 32-bit systems will usually be debugging 32-bit core dumps that will probably still fit in the address space.

chenyingk:
> We added an mmap interface for kcore (https://lore.kernel.org/patchwork/patch/1453867/).
> It can reduce system calls significantly when accessing the kernel data structures.
> The patch will be merged into next kernel version.
> 
> Co-developed-by:Feng Zhou [zhoufeng.zf@bytedance.com](mailto:zhoufeng.zf@bytedance.com)
> Signed-off-by: chenying [chenying.kernel@bytedance.com](mailto:chenying.kernel@bytedance.com)

chenyingk:
> Sorry for the delay, I was on vacation last week and am catching up.
> 
> This is amazing! I wasn't sure that it'd be possible to implement `mmap` for `/proc/kcore` in the kernel.
> 
> I have a few small comments on the code, and a couple of more general questions:
> 
> 1. How well will this work for normal core dump files? My concern is that core dump segments might not be always be aligned to the page size, but at least on the core dumps that I have lying around, they are properly aligned.
> 2. Should we skip the mmap path on 32-bit systems? Virtual address space is much tighter there, but I suppose that 32-bit systems will usually be debugging 32-bit core dumps that will probably still fit in the address space.

Thanks for your comments. I will revision addresses your comments in the next few days. But，I'm sorry I don't fully understand your concern about the mmap for 32-bit systems.

chenyingk:
All comments have been addressed.

PR #118, opened 2021-09-05
libdrgn: reads from segment mapping

chenyingk:
We added an mmap interface for kcore (https://lore.kernel.org/patchwork/patch/1453867/).
It can reduce system calls significantly when accessing the kernel data structures.
The patch will be merged into next kernel version.

Co-developed-by: Feng Zhou <zhoufeng.zf@bytedance.com>
Signed-off-by: chenying <chenying.kernel@bytedance.com>


chenyingk:
The kernel patch "fs/proc/kcore.c: add mmap interface" was dropped，so I closed this pull request .

PR #119, opened 2021-09-24
dwarf_index.c: Shrink abbrev tables before saving them in CUs

jgkamat:
Firstly, there's no rush on the pulls I'm submitting at all - I'm just putting it on the queue for discussion later on.

In larger binaries, there can be a large number of CUs, and since we
store an abbrev table for each CU the extra space starts to add up.
The simplest way to mitigate this is to shrink the vectors before
saving them.

On a large binary, I noticed a memory reduction from 20.4G RES to 18.6G
RES (on initial load-in).

I'm not sure if this will cause a CPU regression, but at worst this would be an extra copy (and hopefully a realloc is cheaper than that), but I think the memory savings are worth it. I don't think a length is stored anywhere for abbrev tables, so I don't think we can pre-calculate the size here.

### DIFF
--- a/libdrgn/dwarf_index.c
+++ b/libdrgn/dwarf_index.c
@@ -838,6 +838,8 @@ static struct drgn_error *read_abbrev_table(struct drgn_dwarf_index_cu *cu,
 			return err;
 		}
 	}
+	uint8_vector_shrink_to_fit(&insns);
+	uint32_vector_shrink_to_fit(&decls);
 	cu->abbrev_decls = decls.data;
 	cu->num_abbrev_decls = decls.size;
 	cu->abbrev_insns = insns.data;


javierhonduco:
This is a great find!! ❤️

osandov:
Agreed, great find, thanks! I have a similar patch sitting in a refactoring branch:

```diff
commit b1cb7831f5d4c5b110700606ef9817614c8a29ab
Author: Omar Sandoval <osandov@osandov.com>
Date:   Tue Aug 24 13:57:28 2021 -0700

    libdrgn: dwarf_index: free pending DIEs after indexing namespace
    
    Once we've cleared the pending DIEs vector, we won't use the vector
    again (unless we load more debugging information), so we can free it.
    
    Signed-off-by: Omar Sandoval <osandov@osandov.com>

diff --git a/libdrgn/dwarf_index.c b/libdrgn/dwarf_index.c
index b347f22..88cc7d1 100644
--- a/libdrgn/dwarf_index.c
+++ b/libdrgn/dwarf_index.c
@@ -2666,6 +2666,7 @@ static struct drgn_error *index_namespace(struct drgn_dwarf_index_namespace *ns)
 		return drgn_error_copy(ns->saved_err);
 	}
 	ns->pending_dies.size = 0;
+	drgn_dwarf_index_pending_die_vector_shrink_to_fit(&ns->pending_dies);
 	return err;
 }
 
```

I'm curious to see if you can measure any savings from that one.

jgkamat:
Also, if you want to automate dco checks, it seems like there is a publicly hosted service to do it on github: https://probot.github.io/apps/dco/

or a very simple script we can run through GH Actions: https://github.com/sifive/pydevicetree/blob/master/scripts/check-dco.sh

osandov:
Thanks for those pointers, that script was a good starting point. I just pushed b4a82c30abf9abec00b667b10c829b1cd4b04a47 to do the DCO check (without pulling the file contents for the whole history of the repository).

osandov:
> Ah, sorry about the missing sign-off! I applied the patch above, and I didn't see any noticeable difference - but I'm testing memory only on initial load for now - it's possible that that patch would save more memory when actively loading types (which is what I'll probably test next) - it seems like a good idea to me :)

Yeah, it should help after you've resolved a type inside of a namespace (and therefore indexed the namespace). I'll go ahead and merge it, I'm pretty sure that it's the right thing to do.

PR #120, opened 2021-09-24
splay_tree.c: Rename splay_tree to avoid conflicts with splay-tree.h

jgkamat:
When linking libdrgn as a static library, the name 'splay_tree' can
conflict with splay-tree.h in gcc (namely splay_tree_splay).

Rename relevant functions to have a 'drgn_' prefix

### DIFF
--- a/libdrgn/binary_search_tree.h
+++ b/libdrgn/binary_search_tree.h
@@ -234,41 +234,41 @@ struct binary_tree_search_result {
 /*
  * Binary search tree variants need to define three functions:
  *
- * variant##_tree_insert_fixup(root, node, parent) is called after a node is
- * inserted (as *root, parent->left, or parent->right). It must set the node's
- * parent pointer and rebalance the tree.
+ * drgn_##variant##_tree_insert_fixup(root, node, parent) is called after a node
+ * is inserted (as *root, parent->left, or parent->right). It must set the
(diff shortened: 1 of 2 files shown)


osandov:
I changed this so we still pass "splay" as the variant to the macro, realigned the whitespace, and updated the commit message to mention libiberty.

This is also missing a signed-off-by. @JakeHillion do I have your permission to add your signed-off-by (see [here](https://github.com/osandov/drgn/blob/main/CONTRIBUTING.rst#signing-off)]?  If we don't hear back, since Jake's internship is over, Jay can I add yours? (I guess #119 was missing a signed-off-by, too, but I missed that before merging. I should add a GitHub actions check for that.)

JakeHillion:
> I changed this so we still pass "splay" as the variant to the macro, realigned the whitespace, and updated the commit message to mention libiberty.
> 
> This is also missing a signed-off-by. @JakeHillion do I have your permission to add your signed-off-by (see [here](https://github.com/osandov/drgn/blob/main/CONTRIBUTING.rst#signing-off)]?  If we don't hear back, since Jake's internship is over, Jay can I add yours? (I guess #119 was missing a signed-off-by, too, but I missed that before merging. I should add a GitHub actions check for that.)

Absolutely! Thanks for checking, apologies for not putting it there originally.

PR #121, opened 2021-10-07
Allow name lookup to fallback to local symbols

brenns10:
As discussed in #108, this commit allows the `Program.symbol()` method (or `drgn_program_find_symbol_by_name`) to fall back to returning a local symbol if no global is found. (Sorry for the delay)

The naive implementation would search once for globals, and then again for locals. But this would cause two lookups in the case of failure. So my implementation does a single search. When a global symbol is found, the search immediately returns. When a local symbol is found, it is stored for later use, unless another local symbol was already found.

If the search returned early, either an error occurred or a global was found, so free the fallback if necessary and return. If the search finished without aborting early, then return a fallback symbol if it w as found. Otherwise, raise a lookup error.

Testing below with drgn on a vmcore:

```
$ drgn -c vmcore -s vmlinux
could not get debugging information for:
kernel modules (could not read depmod: open: /lib/modules/5.4.17-2102.203.6.el8uek.x86_64/modules.dep.bin: No such file or directory)
drgn 0.0.14+30.g861160b.dirty (using Python 3.9.5, elfutils 0.180, with libkdumpfile)
For help, type help(drgn).
>>> import drgn
>>> prog.symbol("schedule")
Symbol(name='schedule', address=0xffffffff92fd92e0, size=0x9a, binding=<SymbolBinding.GLOBAL: 2>, kind=<SymbolKind.FUNC: 2>)
>>> prog.symbol("slab_bug")
Symbol(name='slab_bug', address=0xffffffff928a6df3, size=0xbc, binding=<SymbolBinding.LOCAL: 1>, kind=<SymbolKind.FUNC: 2>)
>>> prog.symbol("name_show")
Symbol(name='name_show', address=0xffffffff927175d0, size=0x64, binding=<SymbolBinding.LOCAL: 1>, kind=<SymbolKind.FUNC: 2>)
```

I want to add some form of test -- it seems like it would be good to create a C program with a couple compilation units, add some static and global symbols, and make it call `abort()`. Then we could use the core dump to test symbol lookup without requiring root to look at the running kernel? Let me know your thoughts on this.

### DIFF
--- a/_drgn.pyi
+++ b/_drgn.pyi
@@ -192,8 +192,12 @@ class Program:
     # address_or_name is positional-only.
     def symbol(self, address_or_name: Union[IntegerLike, str]) -> Symbol:
         """
-        Get the symbol containing the given address, or the global symbol with
-        the given name.
+        Get the symbol containing the given address, or symbol with the given
+        name.
+
+        When performing name lookups, global (i.e.
+        :class:`SymbolBinding.GLOBAL`) symbols are preferred, but a local
+        symbol may be returned if there is a local match and no global.
 
         :param address_or_name: The address or name.
         :raises LookupError: if no symbol contains the given address or matches
--- a/libdrgn/drgn.h.in
+++ b/libdrgn/drgn.h.in
@@ -809,6 +809,11 @@ drgn_program_find_symbol_by_address(struct drgn_program *prog, uint64_t address,
 /**
  * Get the symbol corresponding to the given name.
  *
+ * If there is a global symbol which matches this name, then it is returned.
+ * Otherwise, an arbitrary local symbol may be returned if that is found. If
+ * there are no global or local symbols with this name, then the lookup fails
+ * with ::DRGN_ERROR_LOOKUP.
+ *
  * @param[out] ret The returned symbol. It should be freed with @ref
  * drgn_symbol_destroy().
  * @return @c NULL on success, non-@c NULL on error.
--- a/libdrgn/program.c
+++ b/libdrgn/program.c
@@ -1143,6 +1143,7 @@ drgn_program_find_symbol_by_address(struct drgn_program *prog, uint64_t address,
 struct find_symbol_by_name_arg {
 	const char *name;
 	struct drgn_symbol **ret;
+	struct drgn_symbol *local_fallback;
 	struct drgn_error *err;
 	bool bad_symtabs;
 };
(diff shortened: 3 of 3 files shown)


brenns10:
(Added forgotten DCO...)

_drgn.pyi
```diff
@@ -192,8 +192,12 @@ class Program:
     # address_or_name is positional-only.
     def symbol(self, address_or_name: Union[IntegerLike, str]) -> Symbol:
         """
-        Get the symbol containing the given address, or the global symbol with
-        the given name.
+        Get the symbol containing the given address, or symbol with the given
+        name.
+
+        When performing name lookups, global (i.e.
+        :class:`SymbolBinding.GLOBAL`) symbols are preferred, but a local
```
osandov, line 199: I think this should be `:attr:`, not `:class:`:

```suggestion
        :attr:`SymbolBinding.GLOBAL`) symbols are preferred, but a local
```

_drgn.pyi
```diff
@@ -192,8 +192,12 @@ class Program:
...
+        symbol may be returned if there is a local match and no global.
```
osandov, line 200: The mention of `SymbolBinding` made me realize that we probably want to flesh out the rules for the other kinds of bindings. I think we want:

1. If a `GLOBAL` or `UNIQUE` symbol is found, return that. (From what I can tell, `UNIQUE` symbols are just a special case of `GLOBAL`, so we can treat them the same for name lookups.)
2. If a `WEAK` symbol is found, return that. (So `GLOBAL` symbols override `WEAK` symbols, just like when linking.)
3. Return any other matching symbol.

osandov (changes requested):
Thanks for picking this back up! That `test_stack_trace` failure indeed looks like a spurious failure, although I haven't seen it before and I wasn't able to reproduce it after a few thousand runs. I left a few comments inline.

osandov:
Hey @brenns10, I ended up needing this functionality. If you don't mind, I'm going to finish what you started and merge it tomorrow.

Specifically, we have a test case that uses an arbitrary global symbol: https://github.com/osandov/drgn/blob/c3f31e28f9193414e92c3b5fd1687ed2365c5739/tests/helpers/linux/test_debug_info.py#L13
However, that global symbol was removed in Linux 5.16, and I couldn't find any other global symbols lying around in the modules that we use for testing, so I'd like to change the test to use a local symbol.

brenns10:
Hi Omar I'm so sorry for going AWOL on this branch, glad you got the functionality in where needed. I'll check the tests you added for next time as well.

PR #122, opened 2021-10-12
dwarf_index.c: lazily allocate shards to avoid memory use on empy ns

jgkamat:
Previously shards were allocated as soon as a namespace was
encountered, which means that we had a large array sitting around for
every ns we saw. By allocating them lazily, we can reduce this usage.

I'm not sure how much of a regression this will cause, but to me it seems pretty minimal.

I could have used a pointer to a fixed size array here instead of a normal pointer, but since this will likely become a dynamically sized array in the future, I left the type as a simple pointer.

There is one usage of shards in `drgn_dwarf_index_iterator_next` before initialization, but assuming that next is not called while the iterator is being initialized, it should be safe to my knowledge)

### DIFF
--- a/libdrgn/dwarf_index.c
+++ b/libdrgn/dwarf_index.c
@@ -19,6 +19,9 @@
 #include "platform.h"
 #include "util.h"
 
+static const size_t DRGN_DWARF_INDEX_SHARD_BITS = 8;
+static const size_t DRGN_DWARF_INDEX_NUM_SHARDS = 1 << DRGN_DWARF_INDEX_SHARD_BITS;
+
 struct drgn_dwarf_index_pending_cu {
 	struct drgn_debug_info_module *module;
 	const char *buf;
@@ -199,21 +202,35 @@ static inline size_t hash_pair_to_shard(struct hash_pair hp)
 	 */
 	return ((hp.first >>
 		 (8 * sizeof(size_t) - 8 - DRGN_DWARF_INDEX_SHARD_BITS)) &
-		(((size_t)1 << DRGN_DWARF_INDEX_SHARD_BITS) - 1));
+		(DRGN_DWARF_INDEX_NUM_SHARDS - 1));
 }
 
 static void
 drgn_dwarf_index_namespace_init(struct drgn_dwarf_index_namespace *ns,
 				struct drgn_dwarf_index *dindex)
 {
-	array_for_each(shard, ns->shards) {
+	ns->shards = NULL;
+	ns->dindex = dindex;
+	drgn_dwarf_index_pending_die_vector_init(&ns->pending_dies);
+	ns->saved_err = NULL;
+}
+
+static bool
+drgn_dwarf_index_namespace_shards_init(struct drgn_dwarf_index_namespace *ns)
+{
+	if (ns->shards)
+		return true;
+	ns->shards = malloc_array(DRGN_DWARF_INDEX_NUM_SHARDS,
+				  sizeof(*ns->shards));
+	if (!ns->shards)
+		return false;
+	for (size_t i = 0; i < DRGN_DWARF_INDEX_NUM_SHARDS; i++) {
+		struct drgn_dwarf_index_shard *shard = &ns->shards[i];
 		omp_init_lock(&shard->lock);
 		drgn_dwarf_index_die_map_init(&shard->map);
 		drgn_dwarf_index_die_vector_init(&shard->dies);
 	}
-	ns->dindex = dindex;
-	drgn_dwarf_index_pending_die_vector_init(&ns->pending_dies);
-	ns->saved_err = NULL;
+	return true;
 }
 
 void drgn_dwarf_index_init(struct drgn_dwarf_index *dindex)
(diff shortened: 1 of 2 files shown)


libdrgn/dwarf_index.c
```diff
@@ -206,16 +206,37 @@ static void
...
 	ns->saved_err = NULL;
 }
 
+static struct drgn_error *
+drgn_dwarf_index_namespace_shards_init(struct drgn_dwarf_index_namespace *ns)
+{
+	// Check once for init before acquiring to avoid lock contention once things are initialized.
+	if (ns->shards)
+		return NULL;
+	const size_t num_shards = 1 << DRGN_DWARF_INDEX_SHARD_BITS;
```
osandov, line 221: Let's `#define DRGN_DWARF_INDEX_NUM_SHARDS (1 << DRGN_DWARF_INDEX_SHARD_BITS)` in `dwarf_index.h` and use that everywhere.

libdrgn/dwarf_index.c
```diff
@@ -2074,6 +2099,8 @@ static struct drgn_error *index_die(struct drgn_dwarf_index_namespace *ns,
 	struct drgn_dwarf_index_die *die;
 
 	hp = drgn_dwarf_index_die_map_hash(&entry.key);
+	if ((err = drgn_dwarf_index_namespace_shards_init(ns)))
+		return err;
```
osandov, line 2103: I think it'd be simpler to set up the shards before we start indexing (in `drgn_dwarf_index_update()` and `index_namespace()`) rather than checking on every call to `index_die()`. I guess doing it that way would allocate unnecessary memory if the namespace is truly empty, but I don't think we need to be concerned with that for a few reasons:

1. The memory still won't be allocated until the user attempts to do a lookup in the namespace (which they probably won't do seeing as there is nothing in that namespace).
2. Neither Clang nor GCC seems to generate debug info for empty namespace definitions anyways.
3. If you really wanted to, you could have an extra check after indexing that frees the shards if they're all empty.

libdrgn/dwarf_index.h
```diff
@@ -145,7 +145,7 @@ struct drgn_dwarf_index_namespace {
 	 *
 	 * This is sharded to reduce lock contention.
 	 */
-	struct drgn_dwarf_index_shard shards[1 << DRGN_DWARF_INDEX_SHARD_BITS];
+	struct drgn_dwarf_index_shard* shards;
```
osandov, line 148:
```suggestion
	struct drgn_dwarf_index_shard *shards;
```

osandov (changes requested):
Is the extra memory usage actually from empty namespaces, or is it from namespaces that were found but haven't been indexed yet? I think it's the latter, because I can't get GCC or Clang to generate debug info for an empty namespace. If so, see my inline comment about a slightly simpler way to do this.

libdrgn/dwarf_index.c
jgkamat: Yeah, the main concern I'm trying to fix is namespaces which haven't been requested yet - setting up shards in those functions seems fine to me!

libdrgn/dwarf_index.c
```diff
@@ -2608,6 +2637,9 @@ drgn_dwarf_index_update(struct drgn_dwarf_index_update_state *state)
 	if (err)
 		goto err;
 
+	if ((err = drgn_dwarf_index_namespace_shards_init(&dindex->global)))
+		return err;
```
osandov, line 2641: Can we do this as the first thing in `drgn_dwarf_index_update()`? Then we'll never need to roll back before we have any shards.

libdrgn/dwarf_index.c
```diff
@@ -2473,7 +2498,11 @@ indirect_insn:;
 
 static void drgn_dwarf_index_rollback(struct drgn_dwarf_index *dindex)
 {
-	array_for_each(shard, dindex->global.shards) {
+	for (size_t i = 0; i < DRGN_DWARF_INDEX_NUM_SHARDS; i++) {
+		// If we don't have shards, there is nothing to roll back
+		if (!dindex->global.shards)
+			break;
+		struct drgn_dwarf_index_shard* shard = &dindex->global.shards[i];
```
osandov, line 2505:
```suggestion
		struct drgn_dwarf_index_shard *shard = &dindex->global.shards[i];
```

libdrgn/dwarf_index.c
```diff
@@ -199,23 +199,44 @@ static inline size_t hash_pair_to_shard(struct hash_pair hp)
...
-		drgn_dwarf_index_die_map_init(&shard->map);
-		drgn_dwarf_index_die_vector_init(&shard->dies);
-	}
+	ns->shards = NULL;
 	ns->dindex = dindex;
 	drgn_dwarf_index_pending_die_vector_init(&ns->pending_dies);
 	ns->saved_err = NULL;
 }
 
+static struct drgn_error *
```
osandov, line 215: This can only fail if it can't allocate memory, so can you please make it return a `bool` (`true` for success, `false` if no memory)?

libdrgn/dwarf_index.c
```diff
@@ -199,23 +199,44 @@ static inline size_t hash_pair_to_shard(struct hash_pair hp)
...
+drgn_dwarf_index_namespace_shards_init(struct drgn_dwarf_index_namespace *ns)
+{
+	// Check once for init before acquiring to avoid lock contention once things are initialized.
+	if (ns->shards)
+		return NULL;
+	#pragma omp critical(drgn_dwarf_index_namespace_shards_init)
```
osandov, line 221: This is always called from a single thread now, so you don't need this critical section.

libdrgn/dwarf_index.c
```diff
@@ -199,23 +199,44 @@ static inline size_t hash_pair_to_shard(struct hash_pair hp)
...
+	{
+		if (!ns->shards) {
+			struct drgn_dwarf_index_shard* shards = malloc_array(
```
osandov, line 224:
```suggestion
			struct drgn_dwarf_index_shard *shards = malloc_array(
```

libdrgn/dwarf_index.c
```diff
@@ -199,23 +199,44 @@ static inline size_t hash_pair_to_shard(struct hash_pair hp)
...
+				DRGN_DWARF_INDEX_NUM_SHARDS, sizeof(*ns->shards));
+			for (size_t i = 0; shards && i < DRGN_DWARF_INDEX_NUM_SHARDS; i++) {
```
osandov, line 226: I'd prefer an explicit `if (!shards) return ...;` check rather than putting it in the loop condition.

PR #124, opened 2021-10-21
libdrgn.c: Skip past DIEs that do not contain target addr

jgkamat:
When doing a scan into DIEs, drgn currently will go one DIE at a time, but if there is a sibling attribute, we might be able to skip ahead in the CU. 

The main question here is if modifying the end of the dies array is safe - it seems like it is to me, and this seems like the simplest way to skip ahead while still using the iterator.

I could also use `dwarf_siblingof` here, but I suspect it tries to do a full traversal instead of just looking at `DW_AT_sibling`, which might waste time.

### DIFF
--- a/libdrgn/debug_info.c
+++ b/libdrgn/debug_info.c
@@ -460,9 +460,18 @@ struct drgn_error *drgn_find_die_ancestors(Dwarf_Die *die, Dwarf_Die **dies_ret,
 	it.cu_end = (const char *)cu_die->addr - cu_die_offset + it.next_cu_off;
 
 	Dwarf_Die *dies;
+	Dwarf_Die sibling;
 	size_t length;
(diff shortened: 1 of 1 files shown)


osandov:
This is exactly what I had in mind initially, but I realized that this approach parses some DIEs redundantly. dwarf_attr() parses the DIE to find the attribute. If it doesn't find it, then we call dwarf_child() or dwarf_siblingof() later, which also parses the DIE to skip over it. It turns out that dwarf_attr() stores the end of the DIE when it doesn't find the attribute, which is either the sibling, child, or null terminator.

I benchmarked the original code, your version, and mine using this trick by finding the ancestors of all DIEs in the drgn Python module shared library. The original code took 11 seconds, yours took 2.3 seconds, and this extra optimization took 1.8 seconds.

I just pushed that to the main branch. Feel free to take a look and test it out.

PR #125, opened 2021-10-22
Add support for looking up incomplete types in namespaces

jgkamat:
Previously (see test) incomplete types were not found by find_complete if they were in namespaces, even though they were indexed by drgn - this pull corrects this. This does do a full traversal of a CU to get the path, but hopefully that will not be too slow since #124 (and if it is, we can disable this for C if you would like). 

I needed to add a small feature to the test suite to support this - which allows us to use ref4 to reference any die rather than just the top-level. 

I do have patches for extending this into classes as well to handle nested classes, but we (without other patches I have floating) don't support indexing within classes yet anyway. The big pitfall for that is that class dies might not have names and just have a signature - in which case we have to follow the signature to get the name, but otherwise it is simply just looking for class tags in addition to namespace tags.

I'm not 100% sure this is the way to free the die array from `drgn_find_die_ancestors`, but I think this should be fine(?)

### DIFF
--- a/libdrgn/dwarf_info.c
+++ b/libdrgn/dwarf_info.c
@@ -5473,6 +5473,60 @@ drgn_base_type_from_dwarf(struct drgn_debug_info *dbinfo,
 	}
 }
 
+static struct drgn_error *
+find_namespace_containing_die(struct drgn_debug_info *dbinfo,
+			      Dwarf_Die *die, const struct drgn_language *lang,
+			      struct drgn_namespace_dwarf_index **ret)
+{
+	struct drgn_error *err;
+
+	struct drgn_namespace_dwarf_index *ns = &dbinfo->dwarf.global;
+	if (!lang->has_namespaces) {
+		*ret = ns;
+		return NULL;
+	}
+
+	Dwarf_Die *ancestors;
+	size_t num_ancestors;
+	err = drgn_find_die_ancestors(die, &ancestors, &num_ancestors);
+	if (err)
+		return err;
+
+	for (size_t i = 0; i < num_ancestors; i++) {
+		if (dwarf_tag(&ancestors[i]) != DW_TAG_namespace)
+			continue;
+		Dwarf_Attribute attr_mem, *attr;
+		if (!(attr = dwarf_attr_integrate(&ancestors[i], DW_AT_name,
+						  &attr_mem)))
+			continue;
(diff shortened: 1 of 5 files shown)


libdrgn/debug_info.c
```diff
@@ -3446,7 +3446,42 @@ drgn_debug_info_find_complete(struct drgn_debug_info *dbinfo, uint64_t tag,
...
+		size_t ancestors_len;
+		if ((err = drgn_find_die_ancestors(incomplete_die, &ancestors, &ancestors_len)))
+			goto err;
+
+		struct drgn_dwarf_index_namespace *ns = &dbinfo->dindex.global;
+		uint64_t ns_tag = DW_TAG_namespace;
+		for (int i = 0; i < ancestors_len; i++) {
+			if (dwarf_tag(&ancestors[i]) != DW_TAG_namespace)
+				continue;
+			const char *name = dwarf_diename(&ancestors[i]);
```
osandov, line 3460: This can be `NULL` if the namespace is unnamed. I don't think we index unnamed namespaces (although we definitely should), but for now, I think you should just skip over unnamed namespaces, since their contents should technically be found in the enclosing namespace.

libdrgn/debug_info.c
```diff
@@ -3446,7 +3446,42 @@ drgn_debug_info_find_complete(struct drgn_debug_info *dbinfo, uint64_t tag,
...
+			ns = index_die->namespace;
+		}
+		err = drgn_dwarf_index_iterator_init(&it, ns, name, strlen(name), &tag, 1);
+		if (err)
+			goto err;
+		index_die = drgn_dwarf_index_iterator_next(&it);
+		if (!index_die)
+			err = &drgn_not_found;
+	err:
+		free(ancestors);
```
osandov, line 3480: Yup, this is the correct way to free the ancestors array.

osandov (changes requested):

osandov:
Hey, I'm finally getting around to catching up on PRs, and since it took me so long to get to this one, I just finished it up. I took a different approach for the test harness by changing the way we reference DIEs in tests from indices to explicit names. The test case itself was great so I kept that as is otherwise. I refactored the fix a bit as well. Once tests pass, I'll merge this.

osandov (approved):

PR #128, opened 2021-11-18
libdrgn: Follow typedefs in enum backing type lookup

jgkamat:
In C++ enums can be a typedef to an int, not just an int itself.

The one questionable part is that we store the underlying type and not the typedef in the enum - but that seems fine to me. However, I think I should be able to store the typedef if we would like as well, I just have to see if it's being used as if it was an int elsewhere.

### DIFF
--- a/libdrgn/debug_info.c
+++ b/libdrgn/debug_info.c
@@ -4194,7 +4194,7 @@ drgn_enum_type_from_dwarf(struct drgn_debug_info *dbinfo,
 					   &qualified_compatible_type);
 		if (err)
 			goto err;
-		compatible_type = qualified_compatible_type.type;
+		compatible_type = drgn_underlying_type(qualified_compatible_type.type);
 		if (drgn_type_kind(compatible_type) != DRGN_TYPE_INT) {
 			err = drgn_error_create(DRGN_ERROR_OTHER,
 						"DW_AT_type of DW_TAG_enumeration_type is not an integer type");
--- a/tests/test_dwarf.py
+++ b/tests/test_dwarf.py
@@ -1723,6 +1723,65 @@ class TestTypes(TestCase):
             ),
         )
 
+    def test_enum_typedef(self):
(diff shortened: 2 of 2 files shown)


libdrgn/debug_info.c
```diff
@@ -4195,6 +4195,7 @@ drgn_enum_type_from_dwarf(struct drgn_debug_info *dbinfo,
 		if (err)
 			goto err;
 		compatible_type = qualified_compatible_type.type;
+		compatible_type = drgn_underlying_type(compatible_type);
```
osandov, line 4198: This is just me be very picky, but could you write this as:

```suggestion
		compatible_type =
			drgn_underlying_type(qualified_compatible_type.type);
```

I prefer assigning a variable with a single expression rather than reassigning a variable multiple times.

tests/test_dwarf.py
```diff
@@ -1723,6 +1723,65 @@ def test_enum(self):
...
+                                    DwarfAttrib(DW_AT.name, DW_FORM.string, "BLUE"),
+                                    DwarfAttrib(DW_AT.const_value, DW_FORM.data1, 2),
+                                ),
+                            ),
+                        ),
+                    ),
+                    DwarfDie(
+                        DW_TAG.typedef,
+                        (
+                            DwarfAttrib(DW_AT.name, DW_FORM.string, "__uint64_t"),
```
osandov, line 1764: This doesn't matter since it's just a test case, but `__uint64_t` being an `unsigned int` doesn't sit right with me :p 

```suggestion
                            DwarfAttrib(DW_AT.name, DW_FORM.string, "__uint32_t"),
```

osandov (changes requested):
Thanks! I think it's okay to throw away the typedef here, since we basically just use it to decide the size and signedness. I left a couple of really nitpicky comments. The Packit failure seems to be an infra error, so we can ignore that.

osandov (approved):

PR #131, opened 2021-12-14
libdrgn: fix for compilation error

alakesh:
On older gcc version e.g. 7.3.1 , we get following compilation error

  CC       libdrgnimpl_la-dwarf_info.lo
../../libdrgn/dwarf_info.c:181:51: error: initializer element is not
constant
 static const size_t DRGN_DWARF_INDEX_NUM_SHARDS = 1 <<
DRGN_DWARF_INDEX_SHARD_BITS;

This fixes the compilation error on older versions of gcc

Signed-off-by: Alakesh Haloi <alakesh.haloi@gmail.com>


osandov:
Thanks! That test failure was a VM failing to start up, so I'll go ahead and merge this.

PR #135, opened 2021-12-16
Silence mypy warnings

brenns10:
With mypy 0.920, two warnings appear on current main:

```
$ mypy --strict --no-warn-return-any drgn _drgn.pyi
drgn/helpers/linux/__init__.py:36: error: Need type annotation for "__all__" (hint: "__all__: List[<type>] = ...")
drgn/helpers/linux/__init__.py:38: error: unused "type: ignore" comment
Found 2 errors in 1 file (checked 33 source files)
```

The "unused" type:ignore directive was necessary for prior versions, so
add --no-warn-unused-ignores, so that we pass on multiple versions.
Apply a List[str] annotation to the __all__ variable to silence the
other error.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,7 +27,7 @@ jobs:
       - name: Generate version.py
         run: python setup.py --version
       - name: Check with mypy
-        run: mypy --strict --no-warn-return-any drgn _drgn.pyi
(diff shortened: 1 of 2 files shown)


osandov (approved):
Thanks! I'll merge this as soon as the remaining builds pass.

PR #137, opened 2021-12-21
helpers: Add SOCKET_I() and SOCK_INODE()

ypl-coffee:
Add helpers to convert between sockets and inodes.  As an example:

	>>> file = fget(task, fd)
	>>> sock = SOCKET_I(file.f_inode)
	>>> sock.type.value_()
	2
	>>> import socket
	>>> int(socket.SOCK_DGRAM)
	2
	>>> inode = SOCK_INODE(sock)

Also add tests for the new helpers to tests/helpers/linux/test_net.py.

### DIFF
--- a/drgn/helpers/linux/net.py
+++ b/drgn/helpers/linux/net.py
@@ -18,6 +18,8 @@ from drgn.helpers.linux.list import hlist_for_each_entry, list_for_each_entry
 from drgn.helpers.linux.list_nulls import hlist_nulls_for_each_entry
 
 __all__ = (
+    "SOCK_INODE",
+    "SOCKET_I",
     "for_each_net",
     "get_net_ns_by_inode",
     "get_net_ns_by_fd",
@@ -27,6 +29,34 @@ __all__ = (
     "sk_fullsock",
     "sk_nulls_for_each",
 )
+
+
+_S_IFMT = 0o170000
+_S_IFSOCK = 0o140000
+
+
+def SOCKET_I(inode: Object) -> Object:
+    """
+    Get a socket from an inode referring to the socket.
+
+    :param inode: ``struct inode *``
+    :return: ``struct socket *``
+    :raises ValueError: If *inode* does not refer to a socket
+    """
+    if inode.i_mode & _S_IFMT != _S_IFSOCK:
+        raise ValueError("not a socket inode")
+
+    return container_of(inode, "struct socket_alloc", "vfs_inode").socket.address_of_()
+
+
+def SOCK_INODE(sock: Object) -> Object:
+    """
+    Get the inode of a socket.
+
+    :param sock: ``struct socket *``
+    :return: ``struct inode *``
+    """
+    return container_of(sock, "struct socket_alloc", "socket").vfs_inode.address_of_()
 
 
 def for_each_net(prog: Program) -> Iterator[Object]:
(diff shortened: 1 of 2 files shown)


drgn/helpers/linux/net.py
```diff
@@ -77,6 +78,24 @@ def get_net_ns_by_fd(task: Object, fd: IntegerLike) -> Object:
     return get_net_ns_by_inode(fget(task, fd).f_inode)
 
 
+_S_IFMT = 0o170000
+_S_IFSOCK = 0o140000
+
+
+def get_socket_by_inode(inode: Object) -> Object:
```
osandov, line 85: It looks like the corresponding function in the kernel is named [`SOCKET_I`](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/include/net/sock.h?h=v5.16-rc6#n1461). drgn helpers usually have the same name as the matching kernel function when possible, what do you think about renaming this to `SOCKET_I`?

There's also [`SOCK_INODE`](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/include/net/sock.h?h=v5.16-rc6#n1466) for going in the opposite direction, should we add that, too?

tests/helpers/linux/test_net.py
```diff
@@ -70,6 +72,17 @@ def test_get_net_ns_by_fd(self):
                 file.fileno(),
             )
 
+    def test_get_socket_by_inode(self):
+        with create_socket(type=socket.SOCK_DGRAM) as skt:
+            inode = fget(self.task, skt.fileno()).f_inode
+            sock = get_socket_by_inode(inode)
+            self.assertEqual(sock.type, socket.SOCK_DGRAM)
+
+        stdin = fget(self.task, sys.stdin.fileno())
```
osandov, line 81: Just in case someone runs the test cases with stdin redirected to a socket (although I see no reason to do that), could you have this test case open `/dev/null` and use that?

osandov:
Just a couple of small comments, thanks!

drgn/helpers/linux/net.py
ypl-coffee: Yes I thought about that, but `SOCKET_I` doesn't check `inode.i_mode`.  See [this](https://elixir.bootlin.com/linux/latest/source/net/netlink/af_netlink.c#L1173):
```c
	if (!S_ISSOCK(inode->i_mode))
		return ERR_PTR(-ENOTSOCK);

	sock = SOCKET_I(inode)->sk;
```
Should we implement both `S_ISSOCK` and `SOCKET_I` instead?  I'm fine with both.

> There's also SOCK_INODE for going in the opposite direction, should we add that, too?

I guess we can just do `sock.file.f_inode`, but let me double check that.

osandov: I think it's okay if our version of `SOCKET_I` checks `inode.i_mode` even if the kernel version doesn't.

The kernel's `SOCK_INODE` does this:

```c
static inline struct inode *SOCK_INODE(struct socket *socket)
{
	return &container_of(socket, struct socket_alloc, socket)->vfs_inode;
}
```

You probably know more than me about whether that's better or worse than `sock.file.f_inode`.

ypl-coffee:
Yes implementing `SOCK_INODE` is better since `sock.file` may be `NULL`.  Please re-review; thanks!

osandov:
Thank you!

PR #140, opened 2022-01-07
Kdump testing framework

Svetlitski-FB:


tests/linux_kernel/vmcore/test_vmcore.py
```diff
@@ -0,0 +1,14 @@
...
+
+VMCORE_PATH = Path("/proc/vmcore")
+
+
+@unittest.skipUnless(VMCORE_PATH.exists(), "not running in kdump")
+class TestAttachToVMCore(unittest.TestCase):
+    def test_attach_to_vmcore(self):
+        prog = Program()
+        prog.set_core_dump("/proc/vmcore")
+        prog.load_debug_info("", default=True, main=True)
```
osandov, line 14: When we add a test case that needs debug info, we should probably have something similar to `LinuxHelperTestCase` (`LinuxVmcoreTestCase`?) that caches the program and handles missing debug info. For this smoke test, I think just remove the `load_debug_info` call; we're just testing whether we can attach to `/proc/vmcore`. Maybe add assertions on `prog.flags` for `!IS_LIVE` and `IS_LINUX_KERNEL` instead.

setup.py
```diff
@@ -187,9 +187,22 @@ def _run_vm(self, kernel_dir):
...
+        command = fr"""
+set -e
+
+cd {shlex.quote(os.getcwd())}
+if "$BUSYBOX" [ -e /proc/vmcore ]; then
+    "$PYTHON" -Bm unittest discover -t . -s tests/linux_kernel/vmcore {"-v" if self.verbose else ""}
+else
+    DRGN_RUN_LINUX_HELPER_TESTS=1 "$PYTHON" -Bm \
+        unittest discover -t . -s tests/helpers/linux {"-v" if self.verbose else ""}
+    if "$BUSYBOX" [ -n "$crashkernel" ]; then
```
osandov, line 199: I'd leave off this conditional; I'd rather that we get test failures if `crashkernel` wasn't set properly.

osandov (commented):
A couple of small changes. For the DCO check, you can just squash our commits together, add a better commit message since mine was pretty sparse, and reset the author to you.

setup.py
Svetlitski-FB: This _always_ fails for kernel version `5.16.0-rc8` though (`crashkernel` seems to never be set properly for that version). You can see this both in the relevant [GitHub actions logs](https://github.com/osandov/drgn/runs/4742796728?check_suite_focus=true#step:7:742), and I'm also able to reproduce it locally.

osandov: Ah, see torvalds/linux@71d2bcec2d4d69ff109c497e6611d6c53c8926d4, which was merged in v5.16-rc7. Before that commit, the `crashkernel` command line parameter wasn't recognized as a command line parameter, so the kernel passed it to the environment of `init`. It seems like this was by accident, so we shouldn't have been looking for it anyways. The correct thing is still to remove this check.

osandov:
I clarified the commit message a bit more and added my own signed-off-by. I'll merge it now, thanks!

osandov (approved):

Issue #92, opened 2021-03-09
Add drgn.Thread API


osandov:
## Problem Statement

drgn currently has no representation of a thread. The only thread-specific operation at the moment is getting a stack trace, which takes a TID to identify the thread:

```pycon
>>> prog.stack_trace(1)
#0  __schedule+0x231/0x654
#1  schedule+0x29/0xa5
#2  schedule_hrtimeout_range_clock+0x18d/0x19c
#3  ep_poll+0x3cd/0x3ec
#4  do_epoll_wait+0xb0/0xc6
#5  __x64_sys_epoll_wait+0x1a/0x1d
#6  do_syscall_64+0x48/0x106
#7  entry_SYSCALL_64+0x7c/0x156
```

When debugging the kernel, it can also take a `struct task_struct *`:

```pycon
>>> task = find_task(prog, 1)
>>> task.type_
struct task_struct *
>>> prog.stack_trace(task)
#0  __schedule+0x231/0x654
#1  schedule+0x29/0xa5
#2  schedule_hrtimeout_range_clock+0x18d/0x19c
#3  ep_poll+0x3cd/0x3ec
#4  do_epoll_wait+0xb0/0xc6
#5  __x64_sys_epoll_wait+0x1a/0x1d
#6  do_syscall_64+0x48/0x106
#7  entry_SYSCALL_64+0x7c/0x156
```

This is fine if you know what TID you want. But, in a lot of cases you probably want/expect the debugger to figure out the interesting TID for you. For example, if you're debugging a core dump, you probably want to look at the thread that segfaulted. We can do that for the kernel with something like:

```pycon
>>> task = per_cpu(prog["runqueues"], prog["crashing_cpu"]).curr
>>> task.type_
struct task_struct *
>>> prog.stack_trace(task)
#0  panic+0x194/0x408
#1  sysrq_handle_crash+0x28/0x2c
#2  __handle_sysrq+0xd0/0x218
#3  write_sysrq_trigger+0xbc/0x108
#4  proc_reg_write+0x90/0xd8
#5  __vfs_write+0x38/0x68
#6  vfs_write+0x11c/0x298
#7  ksys_write+0x84/0x13c
#8  system_call+0x5c/0x0
```

But there's currently no way to do this for userspace core dumps, and even for kernel core dumps it'd be more reliable to get it from the core dump metadata instead.

The goal of this issue is to add generic APIs to represent and find threads in drgn.

## API

[...]

PR #142, opened 2022-01-12
Live userspace process support for the Threads API

Svetlitski-FB:
Building off of the work of #129, this PR (when completed) will implement and extend the Threads API outlined in #92 for live userspace processes. This will bring `drgn` one step closer to fulfilling the traditional role of a userspace debugger, most notably adding the ability to pause and resume threads at will during the execution of the program.

As this is still a work in progress, this PR is in **draft** mode.

### DIFF
--- a/_drgn.pyi
+++ b/_drgn.pyi
@@ -853,6 +853,21 @@ class Thread:
         :meth:`Program.stack_trace()`.
         """
         ...
+    def pause(self) -> None:
+        """
+        Suspend execution of this thread until :meth:`Thread.resume()` is called.
+
+        This function shoulld only be called when debugging a live userspace process.
+        """
+        ...
+    def resume(self) -> None:
+        """
+        Resume execution of this thread. It is an error to call this function
+        without having previously called :meth:`Thread.pause()`.
+
+        This function should only be called when debugging a live userspace process.
+        """
+        ...
 
 def filename_matches(haystack: Optional[str], needle: Optional[str]) -> bool:
     """
--- a/libdrgn/arch_x86_64.c
+++ b/libdrgn/arch_x86_64.c
@@ -341,6 +341,16 @@ pt_regs_get_initial_registers_x86_64(const struct drgn_object *obj,
 }
 
 static struct drgn_error *
+user_regs_struct_get_initial_registers_x86_64(struct drgn_program *prog,
+					      const void *user_regs_struct,
+					      struct drgn_register_state **ret)
+{
+	return get_initial_registers_from_struct_x86_64(
+		prog, user_regs_struct, arch_info_x86_64.user_regs_struct_size,
+		ret);
+}
+
+static struct drgn_error *
 prstatus_get_initial_registers_x86_64(struct drgn_program *prog,
 				      const void *prstatus, size_t size,
 				      struct drgn_register_state **ret)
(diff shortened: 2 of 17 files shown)


Svetlitski-FB:
@osandov Presently this is now a MVP of the feature. The core interface works well, has test cases, and is passing all of the existing tests. Due to the complexity and subtly of `ptrace` though (e.g. the 4 different types of `ptrace`-stop states, signal queuing, etc.), it's likely there are as-of-yet undiscovered edge-cases. Right now only x86 is supported, but adding other architectures should be relatively straightforward.

osandov:
Sorry for the delay. I've been busy with other work, but this has been in the back of my head for awhile while I though about what the interface should look like. Thanks to your experimentation and MVP, I think we uncovered a few issues that we should consider:

1. ptrace hijacks threads exiting and receiving signals by default. I can imagine use cases for drgn that want this, and use cases that want the process to run undisturbed. So, I think that we should have interfaces to explicitly attach to and detach from individual threads, attach to all current and future threads, and detach from all threads. By default, we probably shouldn't attach to anything at the library level; we may or may not want the CLI to attach to everything by default.
2. ptrace is inherently event-based. This is pretty fundamental to debugging an external program: the program is going to asynchronously hit conditions that we'd like to react to. So, I think our API should be event-based as well. Specifically, I think we can have an interface that waits for the next event on a specific thread (example use case: I just sent a signal to a thread and I want to wait until that thread stops for that signal) and an interface that waits for the next event on any thread (example use case: I set a breakpoint and now I want the program to run until any thread hits it). It'd also be nice to have non-blocking versions of those (i.e., return and event immediately if it's available, otherwise bail).
3. ptrace supports various kinds of events, some of which are optional. We might want to have a way to enable/disable those optional ones, although I'm fine leaving that out for now. I think by default we want clone/fork/vfork, exec, and exit (which I believe is everything that ptrace supports other than seccomp).

This is a rough sketch of what I'm picturing:

```python3
class Program:
    def attach_all_threads(self) -> None:
        """
        Attach to all threads in this program as well as any new threads that
        are created.
        """
        ...

    def detach_all_threads(self) -> None:
        """Detach from all threads in this program."""
        ...

    def get_thread_event(self, block: bool = True) -> ThreadEvent:
        """
        Get the next event for any thread in this program. If *block* is
        ``True``, wait for the next event. If *block* is ``False``, raises an
        exception if no event is available (TODO: which one? Non-blocking
        sockets seem to return an OSError with errno set appropriately).
        """
        ...


class Thread:
    def attach(self) -> bool:
        """
        Attach to this thread.

        :return: ``True`` on success, ``False`` if the thread was already attached
        """
        ...

    def detach(self) -> bool:
        """
        Detach from this thread.

        :return: ``True`` on success, ``False`` if the thread was not attached
        """
        ...

    def interrupt(self) -> None:
        """
        """
        ...

    def continue(self) -> bool:
        """
        If this thread is currently stopped, resume it.

        :return: ``True`` on success, ``False`` if the thread was not stopped
        """
        ...

    def get_event(self, block: bool = True) -> ThreadEvent:
        """
        Get the next event for this thread.
        """
        ...
```

Then, we can have a dumb `ThreadEvent` object for different types of events. Maybe something like:

```python
ThreadEvent = Union[ThreadEventSignal, ThreadEventExit, etc...]

class ThreadEventSignal:
    signal: int

class ThreadEventExit:
    status: int

etc...
```

The above (along with the equivalent libdrgn interfaces) is more or less what I'd consider the MVP for this design. There are obviously lots of finer details to consider, as well as complications with ptrace that I may not have considered.

There are some "extras" that we should implement eventually, possibly in follow-up PRs:

* We might still want a shortcut `Thread.pause()`, which I think would basically be:
```python
def pause(self):
    self.interrupt()
    while True:
        event = self.get_event()
        if isinstance(event, ThreadEventStop):
            return
        # TODO: should we have a distinction between "stopping" events that
        # require a continue and "non-stopping" events?
        self.continue()  
```
* ptrace allows suppressing signals that the program received. Should we have a `suppress_signal: bool = False` parameter to `continue()`?
* How would we expose `PTRACE_SYSCALL` and `PTRACE_SINGLESTEP`? More parameters for `continue()`?

This is definitely larger in scope than the basic `pause` and `resume` API from before, but I'm hoping that it makes drgn useful for a much more diverse set of tools. I'd love to hear your thoughts.

osandov:
I wrote this project up as an issue, so I'm closing this one. Thanks again for getting this started!

PR #146, opened 2022-01-21
Adding helper to get the kernel ring buffer

alexlzhu:
Helper allows user to obtain the raw contents of dmesg. This would be useful to have within drgn itself for debugging.

### DIFF
--- a/drgn/helpers/linux/dmesg.py
+++ b/drgn/helpers/linux/dmesg.py
@@ -1,0 +1,170 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+"""
+Log Buffer
+----------
+
+The ``drgn.helpers.linux.dmesg`` module provides helpers for reading the Linux
+kernel log buffer.
+"""
+
+from typing import Dict, List, NamedTuple
+
+from drgn import Object, Program, cast, sizeof
+
+__all__ = ("get_printk_records",)
+
+
+class PrintkRecord(NamedTuple):
+    """Kernel log record."""
+
+    text: bytes
+    """Message text."""
+    facility: int
+    """:manpage:`syslog(3)` facility."""
+    level: int
+    """Log level."""
+    seq: int
+    """Sequence number."""
+    timestamp: int
+    """Timestamp in nanoseconds."""
+    continuation: bool
+    """Whether this record is a continuation of a previous record."""
+    context: Dict[bytes, bytes]
+    """
+    Additional metadata for the message.
+
+    See the |/dev/kmsg documentation|_ for an explanation of the keys and
+    values.
+
+    .. |/dev/kmsg documentation| replace:: ``/dev/kmsg`` documentation
+    .. _/dev/kmsg documentation: https://www.kernel.org/doc/Documentation/ABI/testing/dev-kmsg
+    """
+
+
+def _get_printk_records_lockless(prog: Program, prb: Object) -> List[PrintkRecord]:
+    ulong_size = sizeof(prog.type("unsigned long"))
+    DESC_SV_BITS = ulong_size * 8
+    DESC_FLAGS_SHIFT = DESC_SV_BITS - 2
+    DESC_FLAGS_MASK = 3 << DESC_FLAGS_SHIFT
+    DESC_ID_MASK = DESC_FLAGS_MASK ^ ((1 << DESC_SV_BITS) - 1)
+
+    LOG_CONT = prog["LOG_CONT"].value_()
+    desc_committed = prog["desc_committed"].value_()
+    desc_finalized = prog["desc_finalized"].value_()
+
+    def record_committed(current_id: int, state_var: int) -> bool:
+        state_desc_id = state_var & DESC_ID_MASK
+        state = 3 & (state_var >> DESC_FLAGS_SHIFT)
+        return (current_id == state_desc_id) and (
+            state == desc_committed or state == desc_finalized
+        )
+
+    desc_ring = prb.desc_ring
+    desc_ring_count = 1 << desc_ring.count_bits.value_()
+    text_data_ring = prb.text_data_ring
+    text_data_ring_mask = (1 << text_data_ring.size_bits) - 1
+
+    result = []
+
+    def add_record(current_id: int) -> None:
+        idx = current_id % desc_ring_count
+        desc = desc_ring.descs[idx]
+        if not record_committed(current_id, desc.state_var.counter.value_()):
+            return
+
+        lpos_begin = desc.text_blk_lpos.begin & text_data_ring_mask
+        lpos_next = desc.text_blk_lpos.next & text_data_ring_mask
+        lpos_begin += ulong_size
+
+        if lpos_begin == lpos_next:
+            # Data-less record.
+            return
+        if lpos_begin > lpos_next:
+            # Data wrapped.
+            lpos_begin -= lpos_begin
+        info = desc_ring.infos[idx]
+        text_len = info.text_len
+        if lpos_next - lpos_begin < text_len:
+            # Truncated record.
+            text_len = lpos_next - lpos_begin
+
+        context = {}
+        subsystem = info.dev_info.subsystem.string_()
+        device = info.dev_info.device.string_()
+        if subsystem:
+            context[b"SUBSYSTEM"] = subsystem
+        if device:
+            context[b"DEVICE"] = device
+
+        result.append(
+            PrintkRecord(
+                text=prog.read(text_data_ring.data + lpos_begin, text_len),
+                facility=info.facility.value_(),
+                level=info.level.value_(),
+                seq=info.seq.value_(),
+                timestamp=info.ts_nsec.value_(),
+                continuation=bool(info.flags.value_() & LOG_CONT),
+                context=context,
+            )
+        )
+
+    head_id = desc_ring.head_id.counter.value_()
+    current_id = desc_ring.tail_id.counter.value_()
+    while current_id != head_id:
+        add_record(current_id)
+        current_id = (current_id + 1) & DESC_ID_MASK
+    add_record(current_id)
+    return result
+
+
+def _get_printk_records_structured(prog: Program) -> List[PrintkRecord]:
+    printk_logp_type = prog.type("struct printk_log *")
+    LOG_CONT = prog["LOG_CONT"].value_()
+
+    result = []
+    current_idx = prog["log_first_idx"]
+    next_idx = prog["log_next_idx"]
+    seq = prog["log_first_seq"].value_()
+    while current_idx != next_idx:
+        log = cast(printk_logp_type, prog["log_buf"] + current_idx)
+        text_len = log.text_len.value_()
+        dict_len = log.dict_len.value_()
+        text_dict = prog.read(log + 1, text_len + dict_len)
+        context = {}
+        if dict_len:
+            for elmt in text_dict[text_len:].split(b"\0"):
+                key, value = elmt.split(b"=", 1)
+                context[key] = value
+        result.append(
+            PrintkRecord(
+                text=text_dict[:text_len],
+                facility=log.facility.value_(),
+                level=log.level.value_(),
+                seq=seq,
+                timestamp=log.ts_nsec.value_(),
+                continuation=bool(log.flags.value_() & LOG_CONT),
+                context=context,
+            )
+        )
+        log_len = log.len.read_()
+        if log_len:
+            current_idx += log_len
+        else:
+            # Zero means the buffer wrapped around.
+            current_idx -= current_idx
+        seq += 1
+    return result
+
+
+def get_printk_records(prog: Program) -> List[PrintkRecord]:
+    """Get a list of records in the kernel log buffer."""
+    # Linux kernel commit 896fbe20b4e2 ("printk: use the lockless ringbuffer")
+    # changed the ring buffer structure completely.
+    try:
+        prb = prog["prb"]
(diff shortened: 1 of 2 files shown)


tests/test_dmesg.py
```diff
@@ -0,0 +1,40 @@
...
+        append = line
+        if line.find(';') != -1:
+            append = line[line.find(';') + 1:]
+        append = append.replace("â", "→")
+        filtered_kmsg_list += [append]
+    return '\n'.join(filtered_kmsg_list).encode()
+
+class TestDmesg(LinuxHelperTestCase):
+    def test_get_dmesg(self):
+        self.assertEqual(read_kmsg().decode(), get_dmesg(self.prog).decode())
```
osandov, line 40: No need to `decode()` either argument here, you can just compare the raw bytes.

osandov (changes requested):
I left some comments on the helper itself and the test case. As we discussed before, the ring buffer is inherently record-based, not just text, and we should expose that. Once your ring buffer iteration code is solid, we can generalize this to return records instead of just text. I'm imagining something like this:

```python
from typing import Dict, List, NamedTuple

class PrintkRecord(NamedTuple):
    text: bytes
    facility: int
    level: int
    seq: int
    timestamp: int
    continuation: bool
    context: Dict[bytes, bytes]


def get_printk_records(prog: Program) -> List[PrintkRecord]:
    ... generalize your current code to return records

def get_dmesg(prog: Program) -> bytes:
    ... get records with get_printk_records() and combine text
```

All of the information in `PrintkRecord` above is the information exposed by `/dev/kmsg`. See the [documentation for `/dev/kmsg`](https://github.com/torvalds/linux/blob/master/Documentation/ABI/testing/dev-kmsg) and `info_print_ext_header()` and `msg_add_dict_text()` in the kernel. Let me know what you think of this interface.

drgn/helpers/linux/dmesg.py
```diff
@@ -0,0 +1,106 @@
...
+DESC_FLAGS_MASK = 0
+DESC_ID_MASK = 0
+
+def id_incr(counter):
+    return (counter + 1) & DESC_ID_MASK
+
+def record_committed(current_id, state_var):
+    state_desc_id = state_var & DESC_ID_MASK
+    state = 3 & (state_var >> DESC_FLAGS_SHIFT)
+    return (current_id == state_desc_id) and (state == 0x1 or state == 0x2)
```
osandov, line 27: Instead of hardcoding 0x1 and 0x2, you can get the states as `prog['desc_committed']` and `prog['desc_finalized']`.

drgn/helpers/linux/dmesg.py
```diff
@@ -0,0 +1,106 @@
...
+
+def dump_record(prog: Program, desc_ring, desc_ring_count, text_data_ring, text_data_ring_size, ulong_size, current_id):
```
alexlzhu, line 29: define within dump_dmesg_lockless

drgn/helpers/linux/dmesg.py
```diff
@@ -0,0 +1,106 @@
...
+
+DESC_SV_BITS = 0
+DESC_FLAGS_SHIFT = 0
+DESC_FLAGS_MASK = 0
+DESC_ID_MASK = 0
+
+def id_incr(counter):
+    return (counter + 1) & DESC_ID_MASK
+
+def record_committed(current_id, state_var):
```
alexlzhu, line 24: define within dump_dmesg_lockless

drgn/helpers/linux/dmesg.py
```diff
@@ -0,0 +1,106 @@
...
+        lpos_begin = 0
+    if lpos_next - lpos_begin < text_len:
+        text_len = lpos_next - lpos_begin
+    return prog.read(text_data_ring.data + lpos_begin, text_len) + str.encode('\n')
+
+def dump_dmesg_lockless(prog: Program, printk_ringbuffer: Object) -> bytes:
+    global DESC_SV_BITS
+    global DESC_FLAGS_SHIFT
+    global DESC_FLAGS_MASK
+    global DESC_ID_MASK
```
osandov, line 53: Instead of making these global, you can define `dump_record()` and `record_committed()` inside of this function and they'll have these variables available in their scope.

drgn/helpers/linux/dmesg.py
```diff
@@ -0,0 +1,106 @@
...
+
+The ``drgn.helpers.linux.dmesg`` module provides helpers for reading the
+Linux kernel log.
+"""
+
+from drgn import Program, Object, cast, sizeof
+
+__all__ = ("get_dmesg",)
+
+DESC_SV_BITS = 0
```
alexlzhu, line 16: define within dump_dmesg_lockless

tests/test_dmesg.py
```diff
@@ -0,0 +1,31 @@
...
+            try:
+                record = os.read(fd, 4096)
+            except OSError as e:
+                if e.errno == errno.EAGAIN:
+                    break
+                else:
+                    raise
+            prefix, _, escaped_text = record.partition(b";")
+            result.extend(unescape_text(escaped_text))
+    return bytes(b"\n".join([line for line in result.split(b"\n") if not (line.startswith(b" ") and (line.find(b"SUBSYSTEM=") >= 0 or line.find(b"DEVICE=") >= 0))]))
```
alexlzhu, line 27: split on \n and then take the first entry to avoid the context dict.

drgn/helpers/linux/dmesg.py
```diff
@@ -0,0 +1,106 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# SPDX-License-Identifier: GPL-3.0-or-later
+
```
alexlzhu, line 3: ci has some lint errors and type checking errors.

osandov:
I haven't looked at the latest version of your code yet, but FYI you should be able to get rid of that fedora-36-armhfp failure by rebasing on the main branch.

tests/test_dmesg.py
```diff
@@ -0,0 +1,42 @@
...
+                    raise
+            prefix, _, escaped_text = record.partition(b";")
+            escaped_text = escaped_text[0 : escaped_text.find(b"\n") + 1]
+            result.extend(unescape_text(escaped_text))
+    return bytes(result)
+
+
+class TestDmesg(LinuxHelperTestCase):
+    def test_get_dmesg(self):
+        self.assertEqual(read_kmsg(), get_dmesg(self.prog))
```
osandov, line 42: We want to test that we're parsing the record metadata correctly as well, so we should have `read_kmsg()` return a list of records. I think something like this would work:

```python3
def read_kmsg():
    result = []
    with open("/dev/kmsg", "rb") as f:
        fd = f.fileno()
        os.set_blocking(fd, False)
        while True:
            try:
                record = os.read(fd, 4096)
            except OSError as e:
                if e.errno == errno.EAGAIN:
                    break
                else:
                    raise
            prefix, _, escaped = record.partition(b";")
            fields = prefix.split(b",")
            syslog = int(fields[0])
            escaped_lines = escaped.splitlines()
            context = {}
            for escaped_line in escaped_lines[1:]:
                assert escaped_line.startswith(b" ")
                key, value = escaped_line[1:].split(b"=", 1)
                context[unescape_text(key)] = unescape_text(value)
            result.append(
                PrintkRecord(
                    text=unescape_text(escaped_lines[0]),
                    facility=syslog >> 3,
                    level=syslog & 7,
                    seq=int(fields[1]),
                    timestamp=int(fields[2]) * 1000,
                    continuation=b"c" in fields[3],
                    context=context,
                )
            )
    return result
```

One problem with that is that timestamps from `get_printk_records()` have nanosecond granularity, but `/dev/kmsg` only has microsecond granularity. So we'd need something like

```suggestion
        self.assertEqual(
            read_kmsg(),
            [
                # Round timestamp down since /dev/kmsg only has microsecond
                # granularity.
                record._replace(timestamp=record.timestamp // 1000 * 1000)
                for record in get_printk_records(self.prog)
            ]
        )
```

drgn/helpers/linux/dmesg.py
```diff
@@ -0,0 +1,166 @@
...
+            current_idx -= current_idx
+        else:
+            current_idx += log.len
+    return result
+
+
+def get_dmesg(prog: Program) -> bytes:
+    """Get the contents of the kernel ring buffer."""
+    # Linux kernel commit 896fbe20b4e2 ("printk: use the lockless ringbuffer")
+    # changed the ring buffer structure completely.
```
osandov, line 165: This comment needs to go in `get_printk_records()`.

osandov (changes requested):

drgn/helpers/linux/dmesg.py
```diff
@@ -0,0 +1,166 @@
...
+    current_idx = prog["log_first_idx"]
+    next_idx = prog["log_next_idx"]
+    seq = prog["log_first_seq"]
+
+    while current_idx != next_idx:
+        log = cast(printk_logp_type, prog["log_buf"] + current_idx)
+        text_dict = prog.read(log + 1, log.text_len + log.dict_len)
+        text_len = int(log.text_len)
+        dict_len = int(log.dict_len)
+        text = text_dict[0:] + str.encode("\n")
```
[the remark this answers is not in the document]
alexlzhu: Yes, that should be.

drgn/helpers/linux/dmesg.py
```diff
@@ -0,0 +1,166 @@
...
+        text = prog.read(text_data_ring.data + lpos_begin, text_len) + str.encode("\n")
+        return [
+            PrintkRecord(
+                text=text,
+                facility=facility,
+                level=level,
+                seq=seq,
+                timestamp=timestamp,
+                continuation=text.startswith(b" "),
+                context=context,
```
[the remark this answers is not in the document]
alexlzhu: I still chose to define context above, this dictionary is empty most of the time.

drgn/helpers/linux/dmesg.py
```diff
@@ -0,0 +1,166 @@
...
+        lpos_begin += ulong_size
+
+        if lpos_begin == lpos_next:
+            return []
+        if lpos_begin > lpos_next:
+            lpos_begin -= lpos_begin
+        if lpos_next - lpos_begin < text_len:
+            text_len = lpos_next - lpos_begin
+
+        text = prog.read(text_data_ring.data + lpos_begin, text_len) + str.encode("\n")
```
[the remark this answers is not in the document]
alexlzhu: I chose to take the newline out of the text, and join on newline in get_dmesg. Otherwise it's rather awkward to have to add a newline to the text section when we read from /dev/kmsg in the test.

drgn/helpers/linux/dmesg.py
```diff
@@ -0,0 +1,158 @@
...
+"""
+
+from typing import Dict, List, NamedTuple
+
+from drgn import Object, Program, cast, sizeof
+
+__all__ = ("get_dmesg", "get_printk_records")
+
+
+LOG_CONT = 8
```
osandov, line 19: You should get this from `prog` in `dump_dmesg_lockless` and `dump_dmesg_structured`. E.g., `LOG_CONT = prog["LOG_CONT"].value_()`.

drgn/helpers/linux/dmesg.py
```diff
@@ -0,0 +1,158 @@
...
+
+        text = prog.read(text_data_ring.data + lpos_begin, text_len)
+
+        context = {}
+        subsystem = info.dev_info.subsystem.string_()
+        device = info.dev_info.device.string_()
+        if len(subsystem):
+            context[b"SUBSYSTEM"] = subsystem
+        if len(device):
+            context[b"DEVICE"] = device
```
osandov, line 97: Nitpick:

```suggestion
        if subsystem:
            context[b"SUBSYSTEM"] = subsystem
        if device:
            context[b"DEVICE"] = device
```

tests/test_dmesg.py
```diff
@@ -0,0 +1,67 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+import errno
+import fcntl
```
osandov, line 5: Unused import.

tests/test_dmesg.py
```diff
@@ -0,0 +1,67 @@
...
+import os
+import pprint
```
osandov, line 7: Unused import.

tests/test_dmesg.py
```diff
@@ -0,0 +1,67 @@
...
+import re
+import sys
```
osandov, line 9: Unused import.

osandov (changes requested):
This looks like it's almost ready, a few last comments and some mypy errors.

drgn/helpers/linux/dmesg.py
```diff
@@ -0,0 +1,158 @@
...
+    while current_idx != next_idx:
+        log = cast(printk_logp_type, prog["log_buf"] + current_idx)
+        text_dict = prog.read(log + 1, log.text_len + log.dict_len)
+        text_len = int(log.text_len)
+        dict_len = int(log.dict_len)
+        text = text_dict[:text_len]
+        context: Dict[bytes, bytes] = {}
+        if dict_len:
+            context = dict(
+                item.split(b"=", 1) for item in text_dict[text_len:].split(b"\0")
```
osandov, line 135: I guess mypy isn't smart enough to accept this code, so we can probably go back to the way you originally did this.

osandov (approved):
I pushed some minor style changes, more documentation, and a version of `get_dmesg()` that includes the timestamps like the dmesg command. As soon as tests pass, I'll merge this, thanks!

PR #151, opened 2022-01-28
Add autoconf option to enable ASAN

Svetlitski-FB:
[AddressSanitizer (a.k.a ASAN)](https://github.com/google/sanitizers/wiki/AddressSanitizer) is incredibly useful during development, especially when dealing with non-deterministic behavior where re-running the code under a debugger won't necessarily reproduce the bug each time. In order not to break any existing workflows, building with ASAN is opt-in (via `--enable-asan`).

I had originally added this just for myself to make debugging easier, but I figured I may as well upstream it given how useful it's been.

PR #153, opened 2022-02-02
Fix double-free of `crashed_thread`

Svetlitski-FB:
Fixes the heap-use-after-free error uncovered by ASAN noted by @osandov in the comments on #151.

Running the test suite with ASAN enabled revealed that when
the current target was a userspace core dump, the `crashed_thread`
member of `struct drgn_program` was being freed twice – once indirectly
via `drgn_thread_set_deinit`, and once explicitly in `drgn_prog_deinit`.

### DIFF
--- a/libdrgn/program.c
+++ b/libdrgn/program.c
@@ -99,7 +99,11 @@ void drgn_program_deinit(struct drgn_program *prog)
 		else
 			drgn_thread_set_deinit(&prog->thread_set);
 	}
-	drgn_thread_destroy(prog->crashed_thread);
+	// When the target is a userspace core dump, `crashed_thread` will have
+	// been included in `prog->thread_set`, and thus freed by the above
+	// call to `drgn_thread_set_deinit`.
+	if (prog->flags & DRGN_PROGRAM_IS_LINUX_KERNEL)
+		drgn_thread_destroy(prog->crashed_thread);
 	free(prog->pgtable_it);
 
 	drgn_object_deinit(&prog->vmemmap);


libdrgn/program.c
```diff
@@ -99,7 +99,10 @@ void drgn_program_deinit(struct drgn_program *prog)
 		else
 			drgn_thread_set_deinit(&prog->thread_set);
 	}
-	drgn_thread_destroy(prog->crashed_thread);
+	// When the target is a userspace core dump, `crashed_thread` will have been
+	// included in `prog->thread_set`, and thus freed by the above call to `drgn_thread_set_deinit`
+	if (!(prog->core_dump_notes_cached && !(prog->flags & DRGN_PROGRAM_IS_LINUX_KERNEL)))
```
osandov, line 104: What is the `prog->core_dump_notes_cached` check for? Is it not sufficient to check `if (prog->flags & DRGN_PROGRAM_IS_LINUX_KERNEL)`?

libdrgn/program.c
```diff
@@ -99,7 +99,10 @@ void drgn_program_deinit(struct drgn_program *prog)
 		else
 			drgn_thread_set_deinit(&prog->thread_set);
 	}
-	drgn_thread_destroy(prog->crashed_thread);
+	// When the target is a userspace core dump, `crashed_thread` will have been
+	// included in `prog->thread_set`, and thus freed by the above call to `drgn_thread_set_deinit`
```
osandov, line 103: I was going to nitpick the `//`-style comments and tell you to use `/* */`, but honestly that's a rule I inherited from the Linux kernel style, which only has that rule because it targets C89, and C89 doesn't have `//`-style comments (although GCC seems to allow it in GNU89 mode now). So I think that's a pointless rule these days, and this is me giving up on it. You can leave it as `//`.

However, I would still like this comment to be wrapped to 80 characters.

libdrgn/program.c
Svetlitski-FB: I forgot that `drgn_thread_destroy` actually will behave just fine when given `NULL`, will change.

osandov:
Thanks!

PR #154, opened 2022-02-05
Work around Fedora Rawhide ARM crash

osandov:
* Add `gen_strswitch.py`.
* Use it to generate a static lookup function for C keywords rather than the runtime hash table that is triggering a compiler or linker bug on Fedora Rawhide for ARM.


osandov:
The Packit failures are due to the Fedora mirrors being down: https://pagure.io/fedora-infrastructure/issue/10529. Will need to rerun this once that's resolved.

PR #157, opened 2022-02-08
Add drgn.Program.main_thread()

### LINKED ISSUE #144: Add drgn.Program.main_thread()
Follow-up to #92. It'd be nice to have a way to get the `drgn.Thread` object for the main thread of a program. This probably only makes sense for userspace programs. (You could make a case that for the Linux kernel, some special kernel thread, e.g. kthreadd, is the "main thread", but I think asking for the main thread of the kernel should just be an error. This is up for debate.) So, the API is simply something like this:

```python
class Program:
    def main_thread(self) -> Thread:
        """
        Get the main thread of the program.

        :raises ValueError: if the program is the Linux kernel
        """
        ...
```

* For core dumps, the main thread can be identified by the the `pid` field of the `PRPSINFO` note, which will match the `pid` field of the `PRSTATUS` note for the main thread:

```console
$ eu-readelf -n crashme_multithreaded_core | grep -A 3 PRPSINFO
  CORE                 136  PRPSINFO
    state: 2, sname: D, zomb: 0, nice: 0, flag: 0x0000000040400404
    uid: 1000, gid: 985, pid: 39143, ppid: 39053, pgrp: 39143, sid: 39053
    fname: crashme_multith, psargs: ./crashme_multithreaded 
$ eu-readelf -n crashme_multithreaded_core | grep -A 4 PRSTATUS
  CORE                 336  PRSTATUS
    info.si_signo: 11, info.si_code: 0, info.si_errno: 0, cursig: 11
    sigpend: <>
    sighold: <>
    pid: 39144, ppid: 39053, pgrp: 39143, sid: 39053
--
  CORE                 336  PRSTATUS
    info.si_signo: 11, info.si_code: 0, info.si_errno: 0, cursig: 11
    sigpend: <>
    sighold: <>
    pid: 39143, ppid: 39053, pgrp: 39143, sid: 39053
```
* For live userspace processes, the main thread has a TID equal to the PID of the process (`struct drgn_program::pid`). Note that this part is blocked on #142.

The implementation needs to:
1. Add a `struct drgn_thread *drgn_program_main_thread(struct drgn_program *prog)` function to libdrgn.
2. Get the main thread as described above. This could either be cached (e.g., as part of `drgn_program_cache_core_dump_notes()` for core dumps and on the first call to `drgn_program_main_thread()` for live processes) or recreated every time.
3. Add the `drgn.Program.main_thread()` method to the Python bindings.

mykola-lysenko:
Currently only supported for user-space crash dumps. E.g. no support
for live user-space application debugging of kernel debugging

Closes #144

Signed-off-by: Mykola Lysenko <mykolal@fb.com>

### DIFF
--- a/libdrgn/program.c
+++ b/libdrgn/program.c
@@ -678,6 +678,31 @@ static struct drgn_error *get_prstatus_pid(struct drgn_program *prog, const char
 	return NULL;
 }
 
+static struct drgn_error *get_prpsinfo_pid(struct drgn_program *prog,
+					   const char *data, size_t size,
+					   uint32_t *ret)
+{
+	bool is_64_bit, bswap;
+	struct drgn_error *err = drgn_program_is_64_bit(prog, &is_64_bit);
+	if (err)
+		return err;
+	err = drgn_program_bswap(prog, &bswap);
+	if (err)
+		return err;
+
+	size_t offset = is_64_bit ? 24 : 12;
+	uint32_t pr_pid;
+	if (size < offset + sizeof(pr_pid)) {
+		return drgn_error_create(DRGN_ERROR_OTHER,
+					 "NT_PRPSINFO is truncated");
+	}
+	memcpy(&pr_pid, data + offset, sizeof(pr_pid));
+	if (bswap)
+		pr_pid = bswap_32(pr_pid);
+	*ret = pr_pid;
+	return NULL;
+}
+
 struct drgn_error *drgn_thread_dup_internal(const struct drgn_thread *thread,
 					    struct drgn_thread *ret)
 {
@@ -766,6 +791,8 @@ drgn_program_cache_core_dump_notes(struct drgn_program *prog)
 	size_t phnum, i;
 	bool found_prstatus = false;
 	uint32_t first_prstatus_tid;
+	bool found_prpsinfo = false;
+	uint32_t prpsinfo_pid;
 
 	if (prog->core_dump_notes_cached)
 		return NULL;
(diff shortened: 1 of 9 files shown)


libdrgn/program.c
```diff
@@ -678,6 +678,30 @@ static struct drgn_error *get_prstatus_pid(struct drgn_program *prog, const char
 	return NULL;
 }
 
+static struct drgn_error *get_prstatus_pgrp(struct drgn_program *prog, const char *data,
+					   size_t size, uint32_t *ret)
```
osandov, line 682: `pgrp` isn't quite what we want here. That's the process group ID: see [Wikipedia](https://en.wikipedia.org/wiki/Process_group) and [credentials(7)](https://man7.org/linux/man-pages/man7/credentials.7.html). Process groups are usually used for shell pipelines. For example, if I do:

```
$ python3 -c 'import os, sys; print("Pipe PID:", os.getpid(), file=sys.stderr)' | ./crashme_multithreaded
Main TID: 3886869
Second TID: 3886870
Pipe PID: 3886868
zsh: segmentation fault (core dumped)  ./crashme_multithreaded
```

Then `pgrp` is the PID of the first process in the pipeline:

```
$ eu-readelf -n core.3886869 | grep PRSTATUS -A 4
  CORE                 336  PRSTATUS
    info.si_signo: 11, info.si_code: 0, info.si_errno: 0, cursig: 11
    sigpend: <>
    sighold: <>
    pid: 3886870, ppid: 3868213, pgrp: 3886868, sid: 3868213
--
  CORE                 336  PRSTATUS
    info.si_signo: 11, info.si_code: 0, info.si_errno: 0, cursig: 11
    sigpend: <>
    sighold: <>
    pid: 3886869, ppid: 3868213, pgrp: 3886868, sid: 3868213
```

So we really do need to get the main TID from the `pid` field of `PRPSINFO` as described in #144:

```
$ eu-readelf -n core.3886869 | grep PRPSINFO -A 2
  CORE                 136  PRPSINFO
    state: 2, sname: D, zomb: 0, nice: 0, flag: 0x0000000040400404
    uid: 66107, gid: 100, pid: 3886869, ppid: 3868213, pgrp: 3886868
```

You should be able to do that in the same loop over notes in `drgn_program_cache_core_dump_notes()`. Note that you can ignore `drgn_program_cache_kdump_notes()`, as that's only used for kernel core dumps.

osandov (changes requested):
Thanks for taking this! The scaffolding all looks right here, but I commented on where we should be getting the main TID from.

libdrgn/program.c
mykola-lysenko: > Thanks for taking this! The scaffolding all looks right here, but I commented on where we should be getting the main TID from.

Thank for super quick review! Going to fix everything

libdrgn/program.c
```diff
@@ -860,6 +901,17 @@ drgn_program_cache_core_dump_notes(struct drgn_program *prog)
 		assert(it.entry);
 		prog->crashed_thread = it.entry;
 	}
+
+	if (found_main_thread) {
```
osandov, line 905: This should be `if (!prog->flags & DRGN_PROGRAM_IS_LINUX_KERNEL)` like `crashed_thread`.

libdrgn/program.c
```diff
@@ -860,6 +901,17 @@ drgn_program_cache_core_dump_notes(struct drgn_program *prog)
...
+		/*
+		 * Similar to the above, but now caching the main thread
+		 */
+		struct drgn_thread_set_iterator it =
+			drgn_thread_set_search(&prog->thread_set,
+					       &main_thread_tid);
+		assert(it.entry);
```
osandov, line 912: This assertion will fail if the core dump is malformed and doesn't have a `PRSTATUS` note matching the `PRPSINFO` note. We shouldn't crash on bad input. Remove this `assert` and the `if (!prog->main_thread)` check in `drgn_program_main_thread()` will take care of it.

(The assertion for `crashed_thread` is different: we know we inserted a thread into the set with that TID, so we must be able to find it.)

libdrgn/program.c
```diff
@@ -678,6 +678,30 @@ static struct drgn_error *get_prstatus_pid(struct drgn_program *prog, const char
...
+{
+	bool is_64_bit, bswap;
+	struct drgn_error *err = drgn_program_is_64_bit(prog, &is_64_bit);
+	if (err)
+		return err;
+	err = drgn_program_bswap(prog, &bswap);
+	if (err)
+		return err;
+
+	size_t offset = is_64_bit ? 24 : 20;
```
osandov, line 692: 24 looks correct for 64-bit, but on 32-bit, `unsigned long` is 4 bytes and `__kernel_uid_t` and `__kernel_gid_t` are `unsigned short` (2 bytes), so I believe the offsets are

```
struct elf_prpsinfo
{
	char	pr_state;	/* 0 */
	char	pr_sname;	/* 1 */
	char	pr_zomb;	/* 2 */
	char	pr_nice;	/* 3 */
	unsigned long pr_flag;	/* 4 */
	__kernel_uid_t	pr_uid;	/* 8 */
	__kernel_gid_t	pr_gid; /* 10 */
	pid_t	pr_pid /* 12 */, pr_ppid, pr_pgrp, pr_sid;
	/* Lots missing */
	/*
	 * The hard-coded 16 is derived from TASK_COMM_LEN, but it can't be
	 * changed as it is exposed to userspace. We'd better make it hard-coded
	 * here.
	 */
	char	pr_fname[16];	/* filename of executable */
	char	pr_psargs[ELF_PRARGSZ];	/* initial part of arg list */
};
```

Running on my i386 VM corroborates that:

```
root@vm-i386:~/drgn# python3 -Bm drgn
drgn 0.0.14+unknown (using Python 3.9.2, elfutils 0.183, without libkdumpfile)
For help, type help(drgn).
>>> import drgn
>>> offsetof(prog.type("struct elf_prpsinfo"), "pr_pid")
12
```

osandov (changes requested):
A few more minor things, thanks!

libdrgn/program.c
mykola-lysenko: Thanks for the explanation, makes sense! Fixed

libdrgn/program.c
mykola-lysenko: Fixed, need to setup 32bit VM to do proper testing myself

libdrgn/program.c
mykola-lysenko: Fixed

osandov (changes requested):
Thanks, this all looks functionally correct, I just had a few final nitpicks.

Issue #83, opened 2021-01-10
Add accessibility (public, private, protected) to drgn.TypeMember


osandov:
We need to:

1. Add the accessibility to `struct drgn_type_member` and `drgn.TypeMember` (probably as an enum).
2. Include it when pretty printing a type.
3. Parse it from DWARF.

From section 5.7.6 in the DWARF 5 spec:

> A data member entry may have a DW_AT_accessibility attribute. If no accessibility attribute is present, private access is assumed for an member of a class and public access is assumed for an member of a structure, union, or interface.

Section 2.8 defines the accessibility codes (`DW_ACCESS_public`, `DW_ACCESS_private`, and `DW_ACCESS_protected`).

prozak:
@osandov I will take this one if you do not mind

PR #158, opened 2022-02-15
Enable prog.type to work with classes

prozak:
Also added end to end CPP tests

This is a prerequisite to #83

Signed-off-by: prozak <nickolay.lysenko@gmail.com>

### DIFF
--- a/libdrgn/build-aux/gen_c_keywords_inc_strswitch.py
+++ b/libdrgn/build-aux/gen_c_keywords_inc_strswitch.py
@@ -7,6 +7,7 @@ C_KEYWORDS = (
     "_Bool",
     "_Complex",
     "char",
+    "class",
     "const",
     "double",
     "enum",
@@ -35,12 +36,15 @@ def main() -> None:
         print(f'\t[{token_kind}] = "{keyword}",')
     print("};")
     print()
-    print("static int identifier_token_kind(const char *s, size_t len)")
+    print("static int identifier_token_kind(const char *s, size_t len, bool cpp)")
     print("{")
     print("\t@memswitch (s, len)@")
     for token_kind, keyword in keywords:
         print(f'\t@case "{keyword}"@')
-        print(f"\t\treturn {token_kind};")
+        if keyword == "class":
+            print(f"\t\treturn cpp ? {token_kind} : C_TOKEN_IDENTIFIER;")
+        else:
+            print(f"\t\treturn {token_kind};")
     print("\t@default@")
     print(f"\t\treturn C_TOKEN_IDENTIFIER;")
     print("\t@endswitch@")
--- a/libdrgn/language_c.c
+++ b/libdrgn/language_c.c
@@ -1642,6 +1642,7 @@ enum {
 	MAX_QUALIFIER_TOKEN = C_TOKEN_ATOMIC,
 	C_TOKEN_STRUCT,
 	C_TOKEN_UNION,
+	C_TOKEN_CLASS,
 	C_TOKEN_ENUM,
 	MAX_KEYWORD_TOKEN = C_TOKEN_ENUM,
 	C_TOKEN_LPAREN,
@@ -1656,9 +1657,15 @@ enum {
 
 #include "c_keywords.inc"
 
-struct drgn_error *drgn_lexer_c(struct drgn_lexer *lexer,
-				struct drgn_token *token) {
+struct drgn_c_family_lexer {
+	struct drgn_lexer lexer;
+	bool cpp;
+};
+
+struct drgn_error *drgn_c_family_lexer_func(struct drgn_lexer *lexer,
+					    struct drgn_token *token) {
 	const char *p = lexer->p;
+	bool cpp = ((struct drgn_c_family_lexer *)lexer)->cpp;
 
 	while (isspace(*p))
 		p++;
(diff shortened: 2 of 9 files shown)


osandov (changes requested):
Thanks, this is something we've been meaning to do forever! As I noted in the comments, unfortunately doing it properly is a bit more work than this PR, but hopefully I laid it out clearly enough.

As for the test case, the failures are because core dumps don't contain debugging symbols, so on your machine drgn is able to find them in the original binary, but elsewhere they are not available. I don't think that it's worth the hassle of trying to bundle the binary in as well; I'd rather do this via unit tests. You could probably do that in a DWARF test in `tests/test_dwarf.py`: see `tests.test_dwarf.TestProgram.test_language` for how to set the program language. I'm happy to elaborate on that if needed.

prozak:
> Thanks, this is something we've been meaning to do forever! As I noted in the comments, unfortunately doing it properly is a bit more work than this PR, but hopefully I laid it out clearly enough.
> 
> As for the test case, the failures are because core dumps don't contain debugging symbols, so on your machine drgn is able to find them in the original binary, but elsewhere they are not available. I don't think that it's worth the hassle of trying to bundle the binary in as well; I'd rather do this via unit tests. You could probably do that in a DWARF test in `tests/test_dwarf.py`: see `tests.test_dwarf.TestProgram.test_language` for how to set the program language. I'm happy to elaborate on that if needed.

Thanks a lot for super quick reply Omar! All makes sense, I will try to get it over the next few days.

osandov:
I just pushed https://github.com/osandov/drgn/commit/50e4ac82452acd86c75e28a301a6eb36ce7e3d09, which also makes this easier to unit test without needing to involve DWARF. I believe something like this would work:

```diff
diff --git a/tests/test_program.py b/tests/test_program.py
index 53ea9d6..309b7c1 100644
--- a/tests/test_program.py
+++ b/tests/test_program.py
@@ -18,6 +18,7 @@ from drgn import (
     ProgramFlags,
     Qualifiers,
     TypeKind,
+    TypeMember,
     host_platform,
 )
 from tests import (
@@ -548,6 +549,29 @@ class TestTypes(MockProgramTestCase):
         self.assertIdentical(self.prog.type("union option"), self.option_type)
         self.assertIdentical(self.prog.type("enum color"), self.color_type)
 
+    def test_class_type(self):
+        struct_class = self.prog.struct_type(
+            "class",
+            8,
+            (
+                TypeMember(self.prog.pointer_type(self.prog.void_type()), "ptr"),
+            ),
+        )
+        class_point = self.prog.class_type(
+            "Point",
+            8,
+            (
+                TypeMember(self.prog.int_type("int", 4, True), "x", 0),
+                TypeMember(self.prog.int_type("int", 4, True), "y", 32),
+            ),
+        )
+        self.types.append(struct_class)
+        self.types.append(class_point)
+        self.prog.language = Language.C
+        self.assertIdentical(self.prog.type("struct class"), struct_class)
+        self.prog.language = Language.CPP
+        self.assertIdentical(self.prog.type("class Point"), class_point)
+
     def test_typedef(self):
         self.types.append(self.pid_type)
         self.assertIdentical(self.prog.type("pid_t"), self.pid_type)
```

(https://github.com/osandov/drgn/commit/5d65ebb04bf45a2c4bc63532841793a07677bf3b, which I pushed as a prerequisite to that change, moved the language definitions in a way that conflicts with the example patches I gave, but the ideas still hold.)

libdrgn/build-aux/gen_c_keywords_inc_strswitch.py
```diff
@@ -8,6 +8,7 @@
     "_Complex",
     "char",
     "const",
+    "class",
```
osandov, line 11: To keep this sorted alphabetically, this needs to be before "const".

libdrgn/language_c.c
```diff
@@ -1656,9 +1657,15 @@ enum {
 
 #include "c_keywords.inc"
 
+struct drgn_c_family_lexer {
+	struct drgn_lexer lexer;
+	bool cpp;
+};
+
 struct drgn_error *drgn_lexer_c(struct drgn_lexer *lexer,
```
osandov, line 1665: There are a couple of other callers of this function that are now accessing invalid memory since they're being given a normal `drgn_lexer`, not the full `drgn_c_family_lexer`. `c_parse_specifier_list` can just hardcode `cpp = false`, and `c_bit_offset` can be given the same treatment as `c_find_type` (add a `lang` parameter and rename it to `c_family_bit_offset`). I'd suggest also renaming this function to `drgn_c_family_lexer_func`. The name is more accurate for its new meaning and renaming it will catch anywhere else we missed.

osandov (changes requested):

libdrgn/language_c.c
prozak: thanks! There is one more place we need to fix in drgn_test_lexer_init. Let me check it

tests/libdrgn.py
```diff
@@ -203,10 +208,12 @@ def __repr__(self):
 
 
 class Lexer:
-    def __init__(self, func, str):
-        self._lexer = _drgn_lexer()
+    def __init__(self, func, str, cpp=False):
+        self._c_family_lexer = _drgn_c_family_lexer()
+        self._lexer = self._c_family_lexer.__getattribute__("lexer")
```
osandov, line 213: Is there some reason that getting the attribute directly doesn't work? I.e.

```suggestion
        self._lexer = self._c_family_lexer.lexer
```

tests/libdrgn.py
```diff
@@ -203,10 +208,12 @@ def __repr__(self):
...
         self._func = func
         self._str = str.encode()
+        self._c_family_lexer.__setattr__("cpp", cpp)
```
osandov, line 216: Similarly, I think this is equivalent:
```suggestion
        self._c_family_lexer.cpp = cpp
```

osandov (changes requested):
The lexer unit test was a tricky one, thanks for handling that, too! A couple of final comments/questions about that test, otherwise this looks good to me.

osandov (approved):
I just fixed some whitespace and will merge now, thanks for following through with this!

PR #159, opened 2022-02-24
packit: add s390x and build for EPEL 9 as well

davide125:
Start building for the newly available chroots as well.


osandov (approved):
I'm not sure what's going on with the CI failures:

```
/*stdin*\ : Read error (39) : premature end 
tar: Unexpected EOF in archive
tar: Unexpected EOF in archive
tar: Error is not recoverable: exiting now
```

I'll look into it.

PR #160, opened 2022-02-25
Make iwyu.py respect the `CFLAGS` environment variable

Svetlitski-FB:
Minor quality of life improvement.

Previously iwyu.py overwrote the value of the `CFLAGS` environment
variable regardless of whether it was already set. This was very
frustrating when trying to set custom `CFLAGS` to be used in the
generation of compile_commands.json, so the script now only sets
`CFLAGS` if it is not already present in the environment.

Signed-off-by: Kevin Svetlitski <svetlitski@fb.com>

### DIFF
--- a/scripts/iwyu.py
+++ b/scripts/iwyu.py
@@ -169,7 +169,8 @@ def ignore_line(path, state, line):
 
 
 def main():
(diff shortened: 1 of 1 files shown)


osandov:
Makes sense, thanks!

PR #163, opened 2022-03-11
helpers: porting slab allocating debugging tools to drgn helpers

alexlzhu:
Adding additional helpers to view slab info for debugging.
Allows for printing slabs, obtainig the struct kmem_cache object
corresponding to a slab and obtaining all slab objects of a given
type.

### DIFF
--- a/drgn/helpers/linux/slab.py
+++ b/drgn/helpers/linux/slab.py
@@ -1,0 +1,55 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+"""
+Slab Allocator
+--------------
+
+The ``drgn.helpers.linux.slab`` module provides helpers for working with the
+Linux slab allocator.
+"""
+
+from typing import Iterator, Optional, Union
+
+from drgn import Object, Program
+from drgn.helpers import escape_ascii_string
+from drgn.helpers.linux.list import list_for_each_entry
+
+__all__ = (
+    "find_slab_cache",
+    "for_each_slab_cache",
+    "print_slab_caches",
+)
+
+
+def for_each_slab_cache(prog: Program) -> Iterator[Object]:
+    """
+    Iterate over all slab caches.
+
+    :return: Iterator of ``struct kmem_cache *`` objects.
+    """
+    return list_for_each_entry(
+        "struct kmem_cache", prog["slab_caches"].address_of_(), "list"
+    )
+
+
+def find_slab_cache(prog: Program, name: Union[str, bytes]) -> Optional[Object]:
+    """
+    Return the slab cache with the given name.
+
+    :param name: Slab cache name.
+    :return: ``struct kmem_cache *``
+    """
+    if isinstance(name, str):
+        name = name.encode()
+    for s in for_each_slab_cache(prog):
+        if s.name.string_() == name:
+            return s
+    return None
+
+
+def print_slab_caches(prog: Program) -> None:
+    """Print the name and ``struct kmem_cache *`` value of all slab caches."""
+    for s in for_each_slab_cache(prog):
+        name = escape_ascii_string(s.name.string_(), escape_backslash=True)
+        print(f"{name} ({s.type_.type_name()})0x{s.value_():x}")
--- a/tests/helpers/linux/test_slab.py
+++ b/tests/helpers/linux/test_slab.py
@@ -1,0 +1,26 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+from drgn.helpers.linux.slab import find_slab_cache, for_each_slab_cache
+from tests.linux_kernel import LinuxKernelTestCase
+
+
+def get_proc_slabinfo_names():
+    with open("/proc/slabinfo", "rb") as f:
+        # Skip the version and header.
+        f.readline()
+        f.readline()
+        return [line.split()[0] for line in f]
+
+
+class TestSlab(LinuxKernelTestCase):
+    def test_for_each_slab_cache(self):
+        self.assertCountEqual(
+            get_proc_slabinfo_names(),
+            [s.name.string_() for s in for_each_slab_cache(self.prog)],
+        )
+
+    def test_find_slab_cache(self):
+        for name in get_proc_slabinfo_names():
+            slab = find_slab_cache(self.prog, name)
+            self.assertEqual(name, slab.name.string_())


drgn/helpers/linux/mm.py
```diff
@@ -291,3 +294,26 @@ def environ(task: Object) -> List[bytes]:
...
+def find_slab(prog: Program, slab_name: str) -> Optional[Object]:
+    """
+    Return the slab object corresponding to slab_name
+
+    :param slab_name: name of the slab
+    :return: ``struct kmem_cache``
+    """
+    slab_caches_addr = prog["slab_caches"].address_of_().read_()
+    for s in list_for_each_entry("struct kmem_cache", slab_caches_addr, "list"):
+        if s.name.string_().decode("utf-8") == slab_name:
```
osandov, line 308: Rather than decoding the name of each slab cache we find, it'd be faster to encode the name we're checking against, which we only need to do once. We can also make `slab_name` accept either `str` or `bytes` for convenience: change the parameter type annotation to `slab_name: Union[str, bytes]`, then at the beginning of the function you can do:

```python
if isinstance(slab_name, str):
    slab_name = slab_name.encode()
```

Then, this check becomes simply:

```suggestion
        if s.name.string_() == slab_name:
```

drgn/helpers/linux/mm.py
```diff
@@ -291,3 +294,26 @@ def environ(task: Object) -> List[bytes]:
...
+
+def find_slab(prog: Program, slab_name: str) -> Optional[Object]:
+    """
+    Return the slab object corresponding to slab_name
+
+    :param slab_name: name of the slab
+    :return: ``struct kmem_cache``
+    """
+    slab_caches_addr = prog["slab_caches"].address_of_().read_()
+    for s in list_for_each_entry("struct kmem_cache", slab_caches_addr, "list"):
```
osandov, line 307: `address_of_()` already returns a value object, so you don't need `read_()` in this case. Also, since this iteration is duplicated in a couple of places, it would be nice to split it into its own helper. Something like

```python
def for_each_slab_cache(prog: Program) -> Iterator[Object]:
    return list_for_each_entry(
        "struct kmem_cache", prog["slab_caches"].address_of_(), "list"
    )
```

(Along with a matching test case, probably just comparing the set of names of the caches it returns against the set of names from `/proc/slabinfo`.)

Then you can use that for both `find_slab()` and `print_slabs()`.

drgn/helpers/linux/mm.py
```diff
@@ -291,3 +294,26 @@ def environ(task: Object) -> List[bytes]:
...
+    return None
+
+
+def print_slabs(prog: Program) -> None:
+    """
+    Print the slab name and the value of each ``struct kmem_cache *`` for all slabs.
+    """
+    slab_caches_addr = prog["slab_caches"].address_of_().read_()
+    for s in list_for_each_entry("struct kmem_cache", slab_caches_addr, "list"):
+        print(f"{s.name.string_().decode()} ({s.type_.type_name()})0x{s.value_():x}")
```
osandov, line 319: You should use `escape_ascii_string()` from `drgn.helpers` instead of `.decode()` to protect against a garbage name here.

```suggestion
        name = escape_ascii_string(s.name.string_(), escape_backslash=True)
        print(f"{name} ({s.type_.type_name()})0x{s.value_():x}")
```

drgn/helpers/linux/mm.py
```diff
@@ -14,8 +14,9 @@
 from typing import Iterator, List, Optional, Union, overload
 
 from _drgn import _linux_helper_read_vm
-from drgn import IntegerLike, Object, Program, cast
+from drgn import FaultError, IntegerLike, Object, Program, Type, cast
```
osandov, line 17: I don't think you're using either one of these imports now that you limited it to the `slab_cache` helpers.

osandov (changes requested):
Thanks! A few minor comments. I'm also wondering whether we should name these `find_slab_cache` and `print_slab_caches`. It seems like the terminology is that a "slab" is a contiguous chunk of memory within a "slab cache", but it doesn't seem to be used consistently. What do you think?

tests/helpers/linux/test_mm.py
```diff
@@ -26,6 +28,15 @@
 from tests.linux_kernel import LinuxKernelTestCase, mlock
 
 
+def get_proc_slabinfo_names():
+    result = []
+    with open("/proc/slabinfo", "rb") as f:
+        proc_slabinfo = f.read().split(b"\n")[2:-1]
+        for info in proc_slabinfo:
+            result.append(info.split()[0])
+    return result
```
osandov, line 37: Slightly simpler:

```suggestion
    with open("/proc/slabinfo", "rb") as f:
        # Skip the version and header.
        f.readline()
        f.readline()
        return [line.split()[0] for line in f]
```

tests/helpers/linux/test_mm.py
```diff
@@ -134,3 +145,14 @@ def test_environ(self):
             proc_environ = f.read().split(b"\0")[:-1]
         task = find_task(self.prog, os.getpid())
         self.assertEqual(environ(task), proc_environ)
+
+    def test_slab_for_each(self):
+        proc_slabinfo_names = get_proc_slabinfo_names()
+        foreach_slab_names = [s.name.string_() for s in for_each_slab_cache(self.prog)]
+        self.assertEqual(proc_slabinfo_names, foreach_slab_names)
```
osandov, line 152: This relies on `for_each_slab_cache()` returning names in the same order as `/proc/slabinfo`. This happens to be true currently since they both iterate through the `slab_caches` list, but I don't think we want to guarantee that. The easy fix is:

```suggestion
        self.assertEqual(sorted(proc_slabinfo_names), sorted(foreach_slab_names))
```

tests/helpers/linux/test_mm.py
```diff
@@ -134,3 +145,14 @@ def test_environ(self):
...
+
+    def test_find_slab_cache(self):
+        proc_slabinfo_names = get_proc_slabinfo_names()
+        for slab_name in proc_slabinfo_names:
```
osandov, line 156: No need for this variable.

```suggestion
        for slab_name in get_proc_slabinfo_names():
```

osandov (changes requested):
A couple of comments on the test case. Also, it just occurred to me that the kernel has all of the slab-related stuff in its own header, `linux/slab.h`, separate from `linux/mm.h`. Maybe we should follow that and have the slab helpers in their own `drgn/helpers/linux/slab.py` module instead of lumping them in `mm`?

tests/helpers/linux/test_mm.py
osandov: Scratch that, apparently you can just do:

```suggestion
        self.assertCountEqual(proc_slabinfo_names, foreach_slab_names)
```

osandov (approved):
I edited the documentation a bit. Will merge as soon as tests pass, thanks!

PR #164, opened 2022-03-16
Debug a given executable without immediately starting a new process

Svetlitski-FB:
See the commit messages for a more detailed description of this new feature.

This is something that previously could be halfway achieved via the hack:
```sh
drgn -p 0 -s executable --no-default-symbols
```

But while that was sufficient for querying the types in a binary, it didn't provide any way to subsequently run the binary.

This addition changes this on the C API, Python bindings, and CLI fronts by providing easy mechanisms for specifying a binary (and arguments to be passed to it) to debug without immediately running it, as well as facilities for later launching a new process with the given binary and arguments. So now you can do the following, which is quite convenient:

```
drgn -a executable arg0 arg1 arg2
>>> prog.type('struct foo') # Can query debug info before running program
>>> prog.run() # Launch the program with the pre-specified arguments at your leisure
>>> prog['my_foo'] # Debug the program as normal 
```

### DIFF
--- a/_drgn.pyi
+++ b/_drgn.pyi
@@ -443,6 +443,17 @@ class Program:
             return an :class:`Object`.
         """
         ...
+    def set_argv(self, executable_path: Path, argv: Sequence[str]) -> None:
+        """
+        Set the program to a yet to be created process with the given executable
+        and arguments.
+
+        :param executable_path: Path to the executable file to be used for the
+            new process.
+        :param argv: argv Arguments supplied to the new process
+            (interpreted as by `execvp(3)`).
+        """
+        ...
     def set_core_dump(self, path: Path) -> None:
         """
         Set the program to a core dump.
@@ -513,6 +524,12 @@ class Program:
         the program.
 
         This is equivalent to ``load_debug_info(None, True)``.
+        """
+        ...
+    def run(self) -> None:
+        """
+        Launch a new process using the executable and arguments previously
+        supplied in a call to  :meth:`set_argv`, and begin debugging it.
         """
         ...
     cache: Dict[Any, Any]
--- a/drgn/internal/cli.py
+++ b/drgn/internal/cli.py
@@ -93,6 +93,12 @@ def main() -> None:
         type=int,
         help="debug the running process with the given PID",
     )
+    program_group.add_argument(
+        "-a",
+        "--argv",
+        action="store_true",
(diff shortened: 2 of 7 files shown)


osandov:
I like the idea of having a way for drgn to execute the program directly. However, I think this PR will have problems with Position Independent Executables when Address Space Layout Randomization is enabled.

To give some background:

* drgn tracks debug info as "modules" via libdwfl. A module is an executable or a library that is loaded at a specified address range. A module can correspond to one ELF file (if the actual binary contains debug info) or two ELF files (if debug info was split into a separate file, like Linux distros do).
* drgn also has a notion of "unloaded" modules: modules which we got debug info for but aren't actually loaded in the program.
* The load address is relevant for a few reasons: it allows us to go from an address to the module containing that address (e.g., for `prog.symbol()`); and in the case of PIE and shared libraries, it allows us to calculate the "bias", which is the difference between the addresses in the ELF file and the addresses in the program.
* `load_debug_info(default=True)` finds all of the modules in the program (including their addresses) and records them.
* drgn currently has a limitation that `load_debug_info([executable])` doesn't know how to find the address range (and thus the bias) for the executable. Instead, it always marks it as "unloaded". I'm fixing this as part of larger rework of debug info loading.

Let's pretend that that last bug is fixed. Even then, when you call `load_debug_info([executable])`, we don't yet know where the executable will be loaded. When we run the program, it will get placed somewhere, and accesses to `prog['foo']` will not necessarily get the right address. I see a couple of options for dealing with this:

1. Initially mark the executable module as "unloaded", then once the program starts, record its real address. libdwfl doesn't have a way to do this, so we would have to stop using libdwfl before we could do this (which is something I'm hoping to do long term anyways).
2. Disable ASLR when running the executable, and report the module with the original addresses from the binary. This has the downside that it would mask ASLR-related heisenbugs, but it's at least workable.

I think 2 is the only real option right now. Rather than doing it with an explicit `load_debug_info([executable])`, I think we should do it via `load_debug_info(default=True)`. I.e., the default debugging information for a program initialized with `set_argv()` is the executable.

One unrelated question: it seems like calling `prog.run()` a second time will result in an error, right? That's probably the safest for now, but in the future we should allow rerunning the process. I'm not sure how easy that will be to support right off the bat, though, so I wouldn't worry about it for now.

PR #166, opened 2022-03-17
helpers:add helpers for obtaining slab objects

alexlzhu:
These are ported from https://github.com/josefbacik/debug-scripts.

### DIFF
--- a/drgn/helpers/linux/slab.py
+++ b/drgn/helpers/linux/slab.py
@@ -9,16 +9,18 @@ The ``drgn.helpers.linux.slab`` module provides helpers for working with the
 Linux slab allocator.
 """
 
-from typing import Iterator, Optional, Union
+from typing import Iterator, Optional, Set, Union
 
-from drgn import Object, Program
+from drgn import FaultError, Object, Program, Type, cast
 from drgn.helpers import escape_ascii_string
 from drgn.helpers.linux.list import list_for_each_entry
+from drgn.helpers.linux.mm import for_each_page, page_to_virt
 
 __all__ = (
     "find_slab_cache",
     "for_each_slab_cache",
     "print_slab_caches",
+    "slab_cache_for_each_allocated_object",
 )
 
 
@@ -53,3 +55,107 @@ def print_slab_caches(prog: Program) -> None:
     for s in for_each_slab_cache(prog):
         name = escape_ascii_string(s.name.string_(), escape_backslash=True)
         print(f"{name} ({s.type_.type_name()})0x{s.value_():x}")
+
+
+def slab_cache_for_each_allocated_object(
+    slab_cache: Object, type: Union[str, Type]
+) -> Iterator[Object]:
+    """
+    Iterate over all allocated objects in a given slab cache.
+
+    Only the SLUB and SLAB allocators are supported; SLOB does not store enough
+    information to identify objects in a slab cache.
+
+    >>> dentry_cache = find_slab_cache(prog, "dentry")
+    >>> next(slab_cache_for_each_allocated_object(dentry_cache, "struct dentry"))
+    *(struct dentry *)0xffff905e41404000 = {
+        ...
+    }
+
+    :param slab_cache: ``struct kmem_cache *``
+    :param type: Type of object in the slab cache.
+    :return: Iterator of ``type *`` objects.
+    """
+    prog = slab_cache.prog_
+    slab_cache_size = slab_cache.size.value_()
+    pointer_type = prog.pointer_type(prog.type(type))
+
+    try:
+        freelist_type = prog.type("freelist_idx_t *")
+        slub = False
+    except LookupError:
+        freelist_type = prog.type("void **")
+        slub = True
+
+    if slub:
+        try:
+            red_left_pad = slab_cache.red_left_pad.value_()
+        except AttributeError:
+            red_left_pad = 0
+
+        try:
+            freelist_offset = slab_cache.offset.value_()
+        except AttributeError:
+            raise ValueError("SLOB is not supported") from None
+
+        def _slub_freelist(slab: Object, offset: int) -> Set[int]:
+            # In SLUB, the freelist is a linked list with the next pointer
+            # located at ptr + slab_cache->offset.
+            freelist_set = set()
+            freelist = slab.freelist.read_()
+
+            while freelist:
+                freelist_set.add(freelist.value_())
+                freelist = cast(freelist_type, freelist + offset)[0].read_()
+            return freelist_set
+
+        def _slab_page_objects(page: Object, slab: Object) -> Iterator[Object]:
+            freelist_set = _slub_freelist(slab, freelist_offset)
+            addr = page_to_virt(page).value_() + red_left_pad
+            end = addr + slab_cache_size * slab.objects
+            while addr < end:
+                if addr not in freelist_set:
+                    yield Object(prog, pointer_type, value=addr)
+                addr += slab_cache_size
+
+    else:
+        try:
+            obj_offset = slab_cache.obj_offset.value_()
+        except AttributeError:
+            obj_offset = 0
+
+        slab_cache_num = slab_cache.num.value_()
+
+        def _slab_freelist(slab: Object) -> Set[int]:
+            # In SLAB, the freelist is an array of free object indices.
+            freelist = cast(freelist_type, slab.freelist)
+            return {freelist[i].value_() for i in range(slab.active, slab_cache_num)}
+
+        def _slab_page_objects(page: Object, slab: Object) -> Iterator[Object]:
+            freelist_set = _slab_freelist(slab)
+            s_mem = slab.s_mem.value_()
+            for i in range(slab_cache_num):
+                if i in freelist_set:
+                    continue
+                yield Object(
+                    prog, pointer_type, value=s_mem + i * slab_cache_size + obj_offset
+                )
+
+    # Linux kernel commit d122019bf061cccc4583eb9ad40bf58c2fe517be ("mm: Split
+    # slab into its own type") (in v5.17) moved slab information from struct
+    # page to struct slab. The former can be casted to the latter.
+    try:
+        slab_type = prog.type("struct slab *")
+    except LookupError:
+        slab_type = prog.type("struct page *")
+
+    PG_slab_mask = 1 << prog.constant("PG_slab")
+    for page in for_each_page(prog):
+        try:
+            if not page.flags & PG_slab_mask:
+                continue
+        except FaultError:
+            continue
+        slab = cast(slab_type, page)
+        if slab.slab_cache == slab_cache:
+            yield from _slab_page_objects(page, slab)
--- a/tests/helpers/linux/test_slab.py
+++ b/tests/helpers/linux/test_slab.py
@@ -1,7 +1,12 @@
 # Copyright (c) Meta Platforms, Inc. and affiliates.
 # SPDX-License-Identifier: GPL-3.0-or-later
 
-from drgn.helpers.linux.slab import find_slab_cache, for_each_slab_cache
+from drgn.helpers.linux.kconfig import get_kconfig
+from drgn.helpers.linux.slab import (
+    find_slab_cache,
+    for_each_slab_cache,
+    slab_cache_for_each_allocated_object,
+)
 from tests.linux_kernel import LinuxKernelTestCase
 
 
@@ -28,6 +33,21 @@ def fallback_slab_cache_names(prog):
 
 
 class TestSlab(LinuxKernelTestCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        try:
+            kconfig = get_kconfig(cls.prog)
+        except LookupError:
+            cls.allocator = None
+        else:
+            for allocator in ("SLUB", "SLAB", "SLOB"):
+                if kconfig.get("CONFIG_" + allocator, "n") != "n":
+                    break
+            else:
+                raise Exception("couldn't find slab allocator config option")
+            cls.allocator = allocator
+
     def test_for_each_slab_cache(self):
         try:
             slab_cache_names = get_proc_slabinfo_names()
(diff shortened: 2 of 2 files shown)


drgn/helpers/linux/slab.py
```diff
@@ -53,3 +55,163 @@ def print_slab_caches(prog: Program) -> None:
...
+        Page may be cast to slab for v5.17 and on
+        :param slab_cache: ``struct kmem_cache``
+        :param page_or_slab: ``struct page`` or ``struct slab`` depending on the kernel version
+        :param obj_type: type of object contained in the slab_cache
+        """
+        freelist_set = obtain_freelist_slab(prog, page_or_slab, freelist_index_size)
+        ret = []
+        offset = 0
+        if prog.type("struct kmem_cache").has_member("obj_offset"):
+            offset = slab_cache.obj_offset.value_()
```
osandov, line 198: Get this once outside of the function like you did for `red_left_pad`.

drgn/helpers/linux/slab.py
```diff
@@ -53,3 +55,163 @@ def print_slab_caches(prog: Program) -> None:
...
+        addr = page_to_virt(cast(prog.type("struct page *"), page_or_slab)).value_()
+        if not isinstance(addr, int):
+            addr = addr.value_()
+        addr += red_left_pad
+
+        end = addr + slab_cache.size * page_or_slab.objects
+        ret = []
+
+        slab_cache_size = slab_cache.size.value_()
+        freelist_set = obtain_freelist_slub(page_or_slab, slab_cache.offset.value_())
```
osandov, line 177: Get `slab_cache.offset.value_()` once outside of the function (maybe name it `freelist_offset`? The kernel naming is very ambiguous.)

drgn/helpers/linux/slab.py
```diff
@@ -53,3 +55,163 @@ def print_slab_caches(prog: Program) -> None:
     for s in for_each_slab_cache(prog):
         name = escape_ascii_string(s.name.string_(), escape_backslash=True)
         print(f"{name} ({s.type_.type_name()})0x{s.value_():x}")
+
+
+def obtain_freelist_slub(page_or_slab: Object, offset: int) -> Set[Object]:
```
osandov, line 60: This actually returns a `Set[int]`, where each `int` is an address in the free list.

drgn/helpers/linux/slab.py
```diff
@@ -53,3 +55,163 @@ def print_slab_caches(prog: Program) -> None:
...
+        freelist = cast("void **", freelist + offset)[0].read_()
+    return freelist_set
+
+
+def obtain_freelist_slab(
+    prog: Program, page_or_slab: Object, freelist_index_size: int
+) -> Set[int]:
+    """slab freelist an array of free object indices"""
+    freelist_set = set()
+    freelist = page_or_slab.freelist.read_()
```
osandov, line 76: Rather than reading the raw bytes of the `freelist`, can you cast the `freelist` to `freelist_idx_t` and use it just like an array?

(You might want to make this a local function so that you only need to look up `freelist_idx_t` and don't need to pass it around, too).

drgn/helpers/linux/slab.py
```diff
@@ -53,3 +55,163 @@ def print_slab_caches(prog: Program) -> None:
...
+
+    if freelist:
+        freelist_set.add(
+            int.from_bytes(prog.read(freelist, freelist_index_size), "big")
+        )
+        index_bytes = b""
+        index = 1
```
osandov, line 83: Did you ever figure out why you need to skip 0?

drgn/helpers/linux/slab.py
```diff
@@ -53,3 +55,163 @@ def print_slab_caches(prog: Program) -> None:
...
+        freelist_set = obtain_freelist_slab(prog, page_or_slab, freelist_index_size)
+        ret = []
+        offset = 0
+        if prog.type("struct kmem_cache").has_member("obj_offset"):
+            offset = slab_cache.obj_offset.value_()
+
+        for i in range(0, slab_cache.num):
+            if i in freelist_set:
+                continue
+            addr = page_or_slab.s_mem.value_() + i * slab_cache.size + offset
```
osandov, line 203: Get `page_or_slab.s_mem.value_()` once outside of the loop, and get `slab_cache.size` once outside of the function (which can be shared with the SLUB code path).

drgn/helpers/linux/slab.py
```diff
@@ -53,3 +55,163 @@ def print_slab_caches(prog: Program) -> None:
...
+        index_bytes = b""
+        index = 1
+        while index_bytes != b"\0":
+            index_bytes = prog.read(freelist + index, freelist_index_size)
+            freelist_set.add(int.from_bytes(index_bytes, "big"))
+            index += 1
+    return freelist_set
+
+
+def for_each_slab_page(prog: Program) -> Iterator[Object]:
```
osandov, line 91: For this and the other module-level functions that aren't being exported, name them with a leading `_`:

```suggestion
def _for_each_slab_page(prog: Program) -> Iterator[Object]:
```

drgn/helpers/linux/slab.py
```diff
@@ -53,3 +55,163 @@ def print_slab_caches(prog: Program) -> None:
...
+    Get the list of all allocated objects of a given slab and type
+
+    >>> dentry_slab = find_slab_cache(prog, "dentry")
+    >>> dentry_type = prog.type("struct dentry")
+    >>> get_allocated_slab_objects(prog, dentry_slab, dentry_type)
+    [Object(prog, 'struct dentry', address=0xffff888c8417de00), Object(prog, 'struct dentry', address=0xffff888c8417dec0)]
+
+    :param slab_cache: ``struct kmem_cache``
+    :param obj_type: type of object contained in the slab
+    """
```
osandov, line 146: Helpers that take an `Object` shouldn't take a `Program` since you can get it from the `Object`.

```suggestion
def get_allocated_slab_objects(slab_cache: Object, obj_type: Type) -> List[Object]:
    """
    Get the list of all allocated objects of a given slab and type

    >>> dentry_slab = find_slab_cache(prog, "dentry")
    >>> dentry_type = prog.type("struct dentry")
    >>> get_allocated_slab_objects(prog, dentry_slab, dentry_type)
    [Object(prog, 'struct dentry', address=0xffff888c8417de00), Object(prog, 'struct dentry', address=0xffff888c8417dec0)]

    :param slab_cache: ``struct kmem_cache``
    :param obj_type: type of object contained in the slab
    """
    prog = slab_cache.prog_
```

drgn/helpers/linux/slab.py
```diff
@@ -53,3 +55,163 @@ def print_slab_caches(prog: Program) -> None:
...
+    # Linux kernel commit d122019bf061cccc4583eb9ad40bf58c2fe517be ("mm: Split slab into its own type")
+    # (in v5.17) moved slab fields out of struct page into struct slab and allows page to be cast to slab
```
osandov, line 148: This comment needs to go with the `struct slab *` lookup, not here.

drgn/helpers/linux/slab.py
```diff
@@ -53,3 +55,163 @@ def print_slab_caches(prog: Program) -> None:
...
+        freelist_index_size = sizeof(prog.type("freelist_idx_t"))
+    except LookupError:
+        slub = True
+
+    red_left_pad = 0
+    if prog.type("struct kmem_cache").has_member("red_left_pad"):
+        red_left_pad += slab_cache.red_left_pad.value_()
+
+    def _slub_page_objects(
+        prog: Program, slab_cache: Object, page_or_slab: Object, obj_type: Type
```
osandov, line 160: You don't need to pass in `prog`, `slab_cache`, or `obj_type` since those are all already in scope from the outer function (same thing for the SLAB version).

osandov (changes requested):
Some comments about the code structure and some cleanups.

One other thing I'm wondering is whether we should make this function a generator rather than having it return a list. But I'll think about that some more once you make these changes.

drgn/helpers/linux/slab.py
```diff
@@ -53,3 +55,163 @@ def print_slab_caches(prog: Program) -> None:
...
+    ) -> List[Object]:
+        """
+        Get the list of all objects from a slub page
+        :param slab_cache: ``struct kmem_cache``
+        :param page_or_slab: ``struct page`` or ``struct slab`` depending on the kernel version
+        :param obj_type: type of object contained in the slab_cache
+        """
+        addr = page_to_virt(cast(prog.type("struct page *"), page_or_slab)).value_()
+        if not isinstance(addr, int):
+            addr = addr.value_()
```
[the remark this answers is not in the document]
alexlzhu: In v4.4 this value is not necessarily an int. It will return an Object. Otherwise this would not be needed.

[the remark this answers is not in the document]
osandov: I'm confused. You already turned the `Object` into an `int` with the `value_()` call in `addr = page_to_virt(cast(prog.type("struct page *"), page_or_slab)).value_()`.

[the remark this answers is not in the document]
alexlzhu: Actually, you are right. I just tested it on 4.4 and this logic is not needed anymore.

osandov (changes requested):
This is much easier to follow now, thanks! There a few more cleanups and optimizations I see.

drgn/helpers/linux/slab.py
```diff
@@ -53,3 +55,141 @@ def print_slab_caches(prog: Program) -> None:
...
+def _obtain_freelist_slab(slab: Object) -> Set[int]:
+    """slab freelist is an array of free object indices"""
+    freelist_set = set()
+    freelist = slab.freelist.read_()
+
+    if freelist:
+        index = 0
+        freelist_array = cast("freelist_idx_t *", freelist)
+        elmt = freelist_array[index].value_()
+        while elmt != 0:
```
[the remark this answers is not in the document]
alexlzhu: 0 is the null terminator, I did not find a way to obtain the size directly but have observed a maximum size of 255 (char type). Would expect it to be similar for short type. 

I did verify that this is how the index is looked up though. See `index_to_obj` in mm/slab.c.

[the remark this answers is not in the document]
osandov: I just read through the kernel code and I'm not seeing any sort of null terminator. What I see is:

* `struct kmem_cache::num` is the number of objects per slab.
* `struct slab::active` is the number of allocated objects in the slab (therefore, `kmem_cache->num - slab->active` is the number of free objects in the slab). It is initially 0 (see `alloc_slabmgmt()`).
* In `cache_init_objs()`, the freelist is initialized to contain all of the indices from 0 to `kmem_cache->num - 1`.
* To get a free object from a slab, `slab_get_obj()` gets the index at `freelist[slab->active]` and increments `slab->active`.
* To return a free object to a slab, `slab_put_obj()` decrements `slab->active` then sets the index at `freelist[slab->active]` to the index of the freed object.

So, from what I can tell, the freelist is an array of `kmem_cache::num` elements of type `freelist_idx_t` (see the assignment to `freelist_size` in `__kmem_cache_create()). As objects are allocated, indices are taken off of the freelist from front to back, then they are returned from back to front. So, the free indices are actually `freelist[slab->active..kmem_cache->num - 1]`.

So I think this should actually look something like

```python3
        def _obtain_freelist_slab(slab: Object) -> Set[int]:
            freelist = cast(freelist_type, slab.freelist)
            return {freelist[i].value_() for i in range(slab.active, slab_cache_num)}
```

Does my explanation look right, or did I miss something?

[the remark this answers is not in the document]
alexlzhu: I believe that is right. I went through it in vmtest to be sure, the value of `slab->active` always matches with `get_free_obj(slab, slab->active)'  and is always less than `kmem_cache::num'. `slab->active` is incremented on `slab_get_obj()` and decremented on `slab_put_obj()`. That seems consistent with your explanation.

drgn/helpers/linux/slab.py
```diff
@@ -53,3 +55,125 @@ def print_slab_caches(prog: Program) -> None:
...
+    PGSlab = 1 << prog.constant("PG_slab")
+    for p in for_each_page(prog):
+        try:
+            if p.flags.value_() & PGSlab:
+                yield p
+        except FaultError:
+            pass
+
+
+def get_allocated_slab_objects(slab_cache: Object, obj_type: Type) -> List[Object]:
```
osandov, line 74: Another thing we discussed was making this return pointer objects for consistency with the other `for_each` helpers. I.e, if the caller passes in `struct dentry`, instead of `Object(prog, 'struct dentry', address=0xffff888c8417de00)`, return `Object(prog, 'struct dentry *', value=0xffff888c8417de00)`.

While we're at it, again for consistency with, .e.g., `list_for_each_entry`, let's allow the type to be either a `str` or `Type`. You could do that like:

```python
pointer_type = prog.pointer_type(prog.type(obj_type))
```

Then later when you construct the objects, do:

```python3
Object(prog, pointer_type, addr)
```

tests/helpers/linux/test_slab.py
```diff
@@ -51,3 +55,9 @@ def test_find_slab_cache(self):
         for name in slab_cache_names:
             slab = find_slab_cache(self.prog, name)
             self.assertEqual(name, slab.name.string_())
+
+    def test_get_allocated_slab_objects(self):
+        type_name = "struct dentry"
+        slab = find_slab_cache(self.prog, "dentry")
+        type_obj = self.prog.type(type_name)
+        self.assertTrue(len(get_allocated_slab_objects(slab, type_obj)) > 0)
```
osandov, line 63: Also as we discussed, we can make this test a bit stronger by checking that some valid dentry is in the list. For example, we can use the dentry for the root directory, `prog["init_fs"].root.dentry`. Something like

```suggestion
        self.assertIn(prog["init_fs"].root.dentry, slab_cache_for_each_allocated_object(slab, "struct dentry"))
```

osandov (changes requested):

osandov:
I just pushed the updates we talked about. I'll merge it as soon as tests pass.

osandov (approved):
I'm not sure what's going on with Packit, so let's just merge this.

PR #167, opened 2022-03-22
object.c: Add support for parsing 128 bit integers

jgkamat:
We still do not support reading values from these integers (ie: object support)
but we at least provide type information for them now. When not supported,
throw an error.

I tried to compile a list of places that needed to be checked with this type format:

```
* [[file:libdrgn/object.c::drgn_object_read_reference(const struct drgn_object *obj,]]
- Gated to only support bit sizes 64 or less
* [[file:libdrgn/object.c::drgn_object_read_bytes(const struct drgn_object *obj, void *buf)]]
- Seems to support large bit sizes
* [[file:libdrgn/object.c::drgn_object_read_value(const struct drgn_object *obj, union drgn_value *value,]]
- Seems to be covered by ~read_reference~
* [[file:libdrgn/object.c::drgn_object_convert_signed(const struct drgn_object *obj, uint64_t bit_size,]]
- Covered by ~read_value~
* [[file:libdrgn/object.c::drgn_object_convert_unsigned(const struct drgn_object *obj, uint64_t bit_size,]]
- Ditto
* [[file:libdrgn/object.c::binary_operands_signed(const struct drgn_object *lhs,]]
- Covered by ~drgn_object_convert_signed~
* [[file:libdrgn/object.c::binary_operands_unsigned(const struct drgn_object *lhs,]]
- Covered by ~drgn_object_convert_unsigned~
* Binary operators/operands
- Covered by ~drgn_object_convert_unsigned~ and ~drgn_object_convert_signed~
* [[file:libdrgn/object.c::drgn_array_object_is_zero(const struct drgn_object *obj,]]
* [[file:libdrgn/object.c::drgn_object_subscript(struct drgn_object *res, const struct drgn_object *obj,]]
```

I'm pretty sure we can handle almost all cases by just checking in read_reference. However, I'm still not sure about the slice/subscript/is_zero functions - but they seem to (not sure at the moment) handle arrays, and I think they can stay as is. I'm also not sure if the DRGN_OBJECT_VALUE case needs to be handled, in my testing I wasn't able to get that to run. 

This is definitely the kind of pull which is more investigation and less code, so please let me know if you think I'm missing any cases!

### DIFF
--- a/libdrgn/drgn.h.in
+++ b/libdrgn/drgn.h.in
@@ -81,6 +81,8 @@ enum drgn_error_code {
 	DRGN_ERROR_OUT_OF_BOUNDS,
 	/** Operation was attempted with absent object. */
 	DRGN_ERROR_OBJECT_ABSENT,
+	/** Functionality is not implemented. */
+	DRGN_ERROR_NOT_IMPLEMENTED,
 	/** Number of defined error codes. */
 	DRGN_NUM_ERROR_CODES,
 } __attribute__((__packed__));
--- a/libdrgn/dwarf_info.c
+++ b/libdrgn/dwarf_info.c
@@ -5276,16 +5276,14 @@ drgn_object_from_dwarf_constant(struct drgn_debug_info *dbinfo, Dwarf_Die *die,
 			return drgn_error_create(DRGN_ERROR_OTHER,
 						 "invalid DW_AT_const_value");
 		}
-		drgn_object_set_signed_internal(ret, &type, svalue);
-		return NULL;
+		return drgn_object_set_signed_internal(ret, &type, svalue);
 	} else if (type.encoding == DRGN_OBJECT_ENCODING_UNSIGNED) {
 		Dwarf_Word uvalue;
 		if (dwarf_formudata(attr, &uvalue)) {
 			return drgn_error_create(DRGN_ERROR_OTHER,
 						 "invalid DW_AT_const_value");
 		}
-		drgn_object_set_unsigned_internal(ret, &type, uvalue);
-		return NULL;
+		return drgn_object_set_unsigned_internal(ret, &type, uvalue);
 	} else {
 		return drgn_error_create(DRGN_ERROR_OTHER,
 					 "unknown DW_AT_const_value form");
--- a/libdrgn/object.c
+++ b/libdrgn/object.c
@@ -93,7 +93,13 @@ drgn_object_type_impl(struct drgn_type *type, struct drgn_type *underlying_type,
 			}
 			ret->bit_size = bit_field_size;
 		}
-		if (ret->bit_size < 1 || ret->bit_size > 64) {
+		// The maximum bit size we can represent is UINT64_MAX. However,
+		// pick an arbitrary smaller limit to keep things from getting
+		// out of hand: 2^24 is the maximum bit width that Clang 13
+		// allows for _ExtInt() (limited by the LLVM IR representation;
+		// technically it's 2^24-1, but it gets rounded up to a
+		// DW_AT_byte_size of 2^24/8).
+		if (ret->bit_size < 1 || ret->bit_size > 16777216) {
 			return drgn_error_format(DRGN_ERROR_INVALID_ARGUMENT,
 						 "unsupported integer bit size (%" PRIu64 ")",
 						 ret->bit_size);
@@ -142,12 +148,21 @@ drgn_object_type_operand(const struct drgn_operand_type *op_type,
 				     op_type->bit_field_size, ret);
 }
 
-void drgn_object_set_signed_internal(struct drgn_object *res,
-				     const struct drgn_object_type *type,
-				     int64_t svalue)
+static struct drgn_error drgn_integer_too_big = {
+	.code = DRGN_ERROR_NOT_IMPLEMENTED,
+	.message = "integer values larger than 64 bits are not yet supported",
+};
+
+struct drgn_error *
+drgn_object_set_signed_internal(struct drgn_object *res,
+				const struct drgn_object_type *type,
+				int64_t svalue)
 {
+	if (type->bit_size > 64)
+		return &drgn_integer_too_big;
 	drgn_object_reinit(res, type, DRGN_OBJECT_VALUE);
 	res->value.svalue = truncate_signed(svalue, type->bit_size);
+	return NULL;
 }
 
 LIBDRGN_PUBLIC struct drgn_error *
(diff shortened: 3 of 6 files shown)


osandov:
I think the general rule is that we can have reference and absent objects with >64-bit integer types, but we currently can't have a value object with >64-bit integer type, and we can't read the value of a >64-bit integer reference. I think your change is all that's required to enforce the second part, but for the first part we need some additional checks.

Everywhere that sets an object calls either `drgn_object_reinit` or `drgn_object_reinit_copy`. Grepping for that comes up with the following callsites that create value objects, striking through ones that aren't relevant here:

- `drgn_object_set_signed_internal`
- `drgn_object_set_unsigned_internal`
- `drgn_object_set_from_buffer_internal`
- ~~`drgn_object_set_float_internal`~~ floating-point
- ~~`drgn_object_copy`~~ only creates value objects as copies of existing value objects
- ~~`drgn_object_read`~~ caught by your patch
- ~~`drgn_object_from_dwarf_location`~~ only for buffer objects
- ~~`buffer_object_from_value`~~ only for buffer objects

Checking that the size is <= 64 bits in those first three functions (in addition to your change) should therefore be sufficient. The only complication is that `drgn_object_set_signed_internal` and `drgn_object_set_unsigned_internal` don't currently return errors, so it touches a bit more code to change that.

I was also curious if there was a better limit for the maximum bit size for references than 128. Clang has an extension that allows you to specify N-bit integers: `_ExtInt(N)` (this was standardized as `_BitInt(N)` in C23). On Clang 13, the maximum bit width is apparently 2^24-1 (which gets rounded up to 2^24/8 bytes in the DWARF):

```
$ echo '_ExtInt(16777215) x;' | clang -x c -g -c - -o test.o
$ eu-readelf --debug-dump=info test.o

DWARF section [ 5] '.debug_info' at offset 0x6c:
 [Offset]
 Compilation unit at offset 0:
 Version: 4, Abbreviation section offset: 0, Address size: 8, Offset size: 4
 [     b]  compile_unit         abbrev: 1
           producer             (strp) "clang version 13.0.1"
           language             (data2) C99 (12)
           name                 (strp) "-"
           stmt_list            (sec_offset) 0
           comp_dir             (strp) "/home/osandov"
 [    1e]    variable             abbrev: 2
             name                 (strp) "x"
             type                 (ref4) [    33]
             external             (flag_present) yes
             decl_file            (data1) <stdin> (1)
             decl_line            (data1) 1
             location             (exprloc)
              [ 0] addr .bss+0 <x>
 [    33]    base_type            abbrev: 3
             name                 (strp) "_ExtInt"
             encoding             (data1) signed (5)
             byte_size            (data4) 2097152
$ echo '_ExtInt(16777216) x;' | clang -x c -g -c - -o test.o
<stdin>:1:1: error: signed _ExtInt of bit sizes greater than 16777215 not supported
_ExtInt(16777216) x;
^
1 error generated.
```

This was recently reduced to 2^23: https://github.com/llvm/llvm-project/commit/2fd180bbb9a7bb8604a5aca31f1ca9dc5358a433.

Since this limit is somewhat arbitrary, maybe we should just go with 2^24: it's small enough to keep things sane but large enough that I can't imagine it being hit in practice.

I pushed these changes combined with your patch along with some additional tests to https://github.com/osandov/drgn/commits/int128. Would you mind taking a look at that?

osandov:
I think the general rule is that we can have reference and absent objects with >64-bit integer types, but we currently can't have a value object with >64-bit integer type, and we can't read the value of a >64-bit integer reference. I think your change is all that's required to enforce the second part, but for the first part we need some additional checks.

Everywhere that sets an object calls either `drgn_object_reinit` or `drgn_object_reinit_copy`. Grepping for that comes up with the following callsites that create value objects, striking through ones that aren't relevant here:

- `drgn_object_set_signed_internal`
- `drgn_object_set_unsigned_internal`
- `drgn_object_set_from_buffer_internal`
- ~~`drgn_object_set_float_internal`~~ floating-point
- ~~`drgn_object_copy`~~ only creates value objects as copies of existing value objects
- ~~`drgn_object_read`~~ caught by your patch
- ~~`drgn_object_from_dwarf_location`~~ only for buffer objects
- ~~`buffer_object_from_value`~~ only for buffer objects

Checking that the size is <= 64 bits in those first three functions (in addition to your change) should therefore be sufficient. The only complication is that `drgn_object_set_signed_internal` and `drgn_object_set_unsigned_internal` don't currently return errors, so it touches a bit more code to change that.

I was also curious if there was a better limit for the maximum bit size for references than 128. Clang has an extension that allows you to specify N-bit integers: `_ExtInt(N)` (this was standardized as `_BitInt(N)` in C23). On Clang 13, the maximum bit width is apparently 2^24-1 (which gets rounded up to 2^24/8 bytes in the DWARF):

```
$ echo '_ExtInt(16777215) x;' | clang -x c -g -c - -o test.o
$ eu-readelf --debug-dump=info test.o

DWARF section [ 5] '.debug_info' at offset 0x6c:
 [Offset]
 Compilation unit at offset 0:
 Version: 4, Abbreviation section offset: 0, Address size: 8, Offset size: 4
 [     b]  compile_unit         abbrev: 1
           producer             (strp) "clang version 13.0.1"
           language             (data2) C99 (12)
           name                 (strp) "-"
           stmt_list            (sec_offset) 0
           comp_dir             (strp) "/home/osandov"
 [    1e]    variable             abbrev: 2
             name                 (strp) "x"
             type                 (ref4) [    33]
             external             (flag_present) yes
             decl_file            (data1) <stdin> (1)
             decl_line            (data1) 1
             location             (exprloc)
              [ 0] addr .bss+0 <x>
 [    33]    base_type            abbrev: 3
             name                 (strp) "_ExtInt"
             encoding             (data1) signed (5)
             byte_size            (data4) 2097152
$ echo '_ExtInt(16777216) x;' | clang -x c -g -c - -o test.o
<stdin>:1:1: error: signed _ExtInt of bit sizes greater than 16777215 not supported
_ExtInt(16777216) x;
^
1 error generated.
```

This was recently reduced to 2^23: https://github.com/llvm/llvm-project/commit/2fd180bbb9a7bb8604a5aca31f1ca9dc5358a433.

Since this limit is somewhat arbitrary, maybe we should just go with 2^24: it's small enough to keep things sane but large enough that I can't imagine it being hit in practice.

I pushed these changes combined with your patch along with some additional tests to https://github.com/osandov/drgn/commits/int128. Would you mind taking a look at that?

jgkamat:
Yup, things look great to me! I think this should cover all the cases I know of, and the test cases look good too. Thanks again for working on this! :)

osandov:
No problem, thanks for your help. I'll merge it as soon as tests pass.

PR #168, opened 2022-04-20
Use dedicated copr for packit builds

davide125:
Builds will show up under https://copr.fedorainfracloud.org/coprs/g/meta/drgn/

Also start building for ELN (which tracks the next RHEL/CentOS major release)

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -17,6 +17,8 @@ jobs:
 - job: copr_build
   trigger: commit
   metadata:
+    owner: "@meta"
+    project: drgn
     targets:
       - fedora-all-aarch64
       - fedora-all-armhfp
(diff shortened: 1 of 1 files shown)


osandov:
fedora-eln-ppc64le looks pretty broken:

```
/var/tmp/rpm-tmp.csTMUo: line 3:  7488 Illegal instruction     (core dumped) touch /var/lib/rpm/.migratedb
```

Also, what's the context for the ownership change? Is it an accounting/visibility thing? (Nothing wrong with that, I'm just curious.)

davide125:
The fedora-eln-ppc64le is an infra issue, filed https://bugzilla.redhat.com/show_bug.cgi?id=2077299 for that.

davide125:
It's mostly so we can have all the build results in the same place, otherwise packit uses an ephemeral repo that's deleted after a while. If you're rather not have this under `@meta` that's also fine, feel free to make a dedicated project on copr.fedorainfracloud.org if you prefer.

osandov:
I'm totally fine having it under `@meta`. Can we drop `fedora-eln-ppc64le` for now and add it back when the SIGILL stuff is fixed?

davide125:
Yeah let's do that, I've just updated this.

PR #180, opened 2022-06-09
Provide source code info for Symbol

PaulZ-98:
For function symbols, provide the source code file, optional line number, and optional column.  This is useful for noting the source code file/line/col when disassembling a function.  In the code, look up the module by the function Symbol address, and use the module to help get the line number.

```
root@ub20-sdb:~/drgn_dev/drgn# drgn
drgn 0.0.19+2.g57320ac.dirty (using Python 3.8.10, elfutils 0.176, with libkdumpfile)
For help, type help(drgn).
>>> import drgn
>>> 
>>> prog.symbol('mutex_lock').source(offset=0, line_number=True, line_column=False)
'/build/linux-hwe-5.8-Ox7OXf/linux-hwe-5.8-5.8.0/kernel/locking/mutex.c: 280'
>>> prog.symbol('mutex_lock').source(offset=42, line_number=True, line_column=False)
'/build/linux-hwe-5.8-Ox7OXf/linux-hwe-5.8-5.8.0/kernel/locking/mutex.c: 284'
>>> 
```

First drgn PR, so was unsure how to develop a test for this... let me know.  Also how to create drgn_Symbol_source_DOC instead of putting in a string directly?  Thanks!


brenns10:
Hi @PaulZ-98, I just wanted to provide some feedback on this. I'm not a maintainer so Omar's opinion will matter more here, but he may not be able to get back to you for a week or two.

First, this is a great idea, thanks for contributing it!

Second, regarding your question:

>  Also how to create drgn_Symbol_source_DOC instead of putting in a string directly?

These are generated from the `_drgn.pyi` stubs file directly. (See the script `docs/exts/drgndoc/docstrings.py`). You've already added the docstring to the stub file, so you should be able to just use `drgn_Symbol_source_DOC` directly, and it will be generated. If not, try doing a clean build.

Third, I have a few notes regarding the approach.

* The stack frame object provides this information already. I know that it can't satisfy all the use cases that this interface could, since stack frames are limited to whatever is currently on some thread's stack. But I think the [API is a bit better](https://drgn.readthedocs.io/en/latest/api_reference.html#drgn.StackFrame.source) - it simply returns a tuple of (filename, line, column), and leaves the string formatting up to the user. It would be more consistent if your API ultimately returned the same tuple, and it would also ensure that users wouldn't need to parse the string returned by `Symbol.source()` if they did care about the column or line and wanted them as integers.
* I'm not sure that this interface belongs with the `Symbol` object. Symbols may be a lot of things - code and data, which should have source information, but also untyped linker symbols or section names, or a variety of other _stuff_ that doesn't really relate to a line in a source file. Symbols also have a lot of unpleasantness: namely, they aren't unique! The same symbol name can appear multiple times, and the same address can be contained by multiple symbols (thus the `Prog.symbols()` API). This does really happen due to static inline functions, or due to name collisions for static functions in different compilation units. Imagine you have an address and you'd like to get some source info about it. First, you'd need to create a symbol and deal with all of the above, including the possibility that there's no symbol containing that address -- something which seems distinctly possible to me. It seems that it would be better if we just had some sort of method `prog.source_info(address)`, which would take a virtual address and return the tuple mentioned above. It would still be easy to use this API if you already have a Symbol, but it would prevent you from needing to interact with Symbols if you don't need to.

I'll also put a few code review notes in the PR itself, which should help regardless of whether Omar agrees with me on the approach.

Again, thanks for this and I hope to see it merged :)

libdrgn/symbol.c
```diff
@@ -40,6 +45,39 @@ void drgn_symbol_from_elf(const char *name, uint64_t address,
 		ret->kind = DRGN_SYMBOL_KIND_UNKNOWN;
 }
 
+LIBDRGN_PUBLIC void drgn_symbol_source(struct drgn_symbol *sym,
```
brenns10, line 48: This API has no way of showing whether it succeeded or failed. Return `struct drgn_error *` and return NULL on success.

```suggestion
LIBDRGN_PUBLIC struct drgn_error *
drgn_symbol_source(struct drgn_symbol *sym,
```

libdrgn/symbol.c
```diff
@@ -40,6 +45,39 @@ void drgn_symbol_from_elf(const char *name, uint64_t address,
...
+	const char *src;
+	int lineno, linecol;
+	if ((src = dwfl_lineinfo(line, &sym_address, &lineno, &linecol,
+				 NULL, NULL)) != NULL) {
+		strncpy(src_file_ret, src, size);
+		*line_num_ret = lineno;
+		*line_col_ret = linecol;
+	}
+
+	return;
```
brenns10, line 78:
```suggestion
	return NULL;
```

libdrgn/symbol.c
```diff
@@ -40,6 +45,39 @@ void drgn_symbol_from_elf(const char *name, uint64_t address,
...
+LIBDRGN_PUBLIC void drgn_symbol_source(struct drgn_symbol *sym,
+				       struct drgn_program *prog,
+				       unsigned long offset,
+				       size_t size,
+				       char *src_file_ret,
+				       int *line_num_ret,
+				       int *line_col_ret)
+{
+	if (sym->kind != DRGN_SYMBOL_KIND_FUNC)
+		return;
```
brenns10, line 57: Needs an error return, otherwise caller won't know if `src_file_ret` is a valid string.

libdrgn/symbol.c
```diff
@@ -40,6 +45,39 @@ void drgn_symbol_from_elf(const char *name, uint64_t address,
...
+
+	unsigned long sym_address = sym->address + offset;
+	struct drgn_debug_info_module *dim;
+	dim = drgn_debug_info_module_byaddress(prog->dbinfo, sym_address);
+	if (dim == NULL)
+		return;
```
brenns10, line 63: Same, needs error return.

libdrgn/symbol.c
```diff
@@ -40,6 +45,39 @@ void drgn_symbol_from_elf(const char *name, uint64_t address,
...
+
+	Dwfl_Line *line = dwfl_module_getsrc(dim->dwfl_module, sym_address);
+	if (line == NULL)
+		return;
```
brenns10, line 67: Same.

libdrgn/python/symbol.c
```diff
@@ -36,6 +36,43 @@ static PyObject *Symbol_richcompare(Symbol *self, PyObject *other, int op)
 	Py_RETURN_BOOL(ret);
 }
 
+static PyObject *Symbol_source(PyObject *self, PyObject *args, PyObject* kwargs)
+{
+	char source[256];
+	char formatted[296];
```
brenns10, line 42: Where do the magic numbers come from? Comments would be nice here, though if you end up ditching the string formatting, then it will be irrelevant.

brenns10 (changes requested):

PaulZ-98:
Thanks for reviewing @brenns10 I will go through your comments and start making changes.

PaulZ-98:
I think I've covered most of your comments @brenns10.  I can always move the code out of Symbol if that's what we decide to do.

pfactum:
I'd love to have something like `prog.source_info(address)` implemented.

My use-case is [`crush`](https://codeberg.org/pf-kernel/crush), where I pair `drgn` with `capstone` to have a nice disassembler (_see [disassemble.py](https://codeberg.org/pf-kernel/crush/src/branch/master/crush/command/disassemble.py)_), and the only thing I miss now is translating arbitrary virtual address into a source code info.

osandov (changes requested):
Sorry about the delay, I was on vacation for a month while you submitted this and have been perpetually behind on code review since then :) I agree with the other comments here that have the interface for this should be `prog.source_info(address)`. I additionally left some comments about the implementation. Thanks for the contribution!

osandov:
FYI I finally got around to adding this as [`source_location()`](https://drgn.readthedocs.io/en/latest/api_reference.html#drgn.source_location)/[`prog.source_location()`](https://drgn.readthedocs.io/en/latest/api_reference.html#drgn.Program.source_location).

sdimitro:
@PaulZ-98 - the most up-to-date sdb repo is https://github.com/sdimitro/sdb now. Long story short: I went to try different things after Delphix for a few years and 2 weeks ago I joined CoreWeave to help with the kernel team so I moved sdb under my profile because Delphix (now Perforce) have their own way of doing things and I don't want to mess that up :)

PR #183, opened 2022-06-22
cli: Set path when importing a script

lsgunth:
When running a python script with the drgn cli, an import of
a module in the current directory does not find that module.
This is surprising for regular python users.

According to the python documentation, sys.path is modified when
a script is passed on the command line:

  "If the script name refers directly to a Python file, the
   directory containing that file is added to the start of
   sys.path, and the file is executed as the __main__ module."

Add the same operation in drgn so that importing modules from
the same directory as the script work as expected.

Link: https://docs.python.org/3/using/cmdline.html#using-on-interface-options
Signed-off-by: Logan Gunthorpe <logang@deltatee.com>


drgn/internal/cli.py
```diff
@@ -176,6 +176,8 @@ def main() -> None:
     init_globals: Dict[str, Any] = {"prog": prog}
     if args.script:
         sys.argv = args.script
+        if os.path.isfile(args.script[0]):
+            sys.path.insert(0, os.path.dirname(args.script[0]) or ".")
```
osandov, line 180: It looks like the `python` command uses the absolute path of the parent directory:

```console
[osandov@relinquished dir]$ cat ../test.py
import sys
print(sys.path)
[osandov@relinquished dir]$ python3 ../test.py
['/home/osandov', '/usr/lib/python310.zip', '/usr/lib/python3.10', '/usr/lib/python3.10/lib-dynload', '/usr/lib/python3.10/site-packages']
```

If the script changes directory, this would result in different behavior. So I think we want:

```suggestion
            sys.path.insert(0, os.path.dirname(os.path.abspath(args.script[0])))
```

drgn/internal/cli.py
```diff
@@ -176,6 +176,8 @@ def main() -> None:
     init_globals: Dict[str, Any] = {"prog": prog}
     if args.script:
         sys.argv = args.script
+        if os.path.isfile(args.script[0]):
```
osandov, line 179: This is an obscure corner case, but if we want exactly the same behavior as the `python` command, then we don't want to do this for zip files. For directories and zip files, the path itself is inserted, not the parent directory (and `runpy` already does that: https://github.com/python/cpython/blob/4e796f56465f7264ddba63c8396d2649e098b617/Lib/runpy.py#L280).

So cargo culting the `runpy` code, maybe we want this:

```suggestion
        import pkgutil
        if pkgutil.get_importer(args.script[0]) is None:
```

I'm not sure whether it's worth doing that.

lsgunth:
Thanks, yes, those are both great suggestions. I've reviewed, tested and force pushed the branch with them made.

osandov:
Merged, thank you again!

Issue #188, opened 2022-07-06
Packit failing on Fedora ELN


osandov:
All Fedora ELN builds are failing with something like:

```
ERROR: Command failed: 
 # /usr/bin/dnf --installroot /var/lib/mock/fedora-eln-x86_64-bootstrap-1657124559.712774/root/ --releasever eln --setopt=deltarpm=False --allowerasing --disableplugin=local --disableplugin=spacewalk --disableplugin=versionlock install dnf dnf-plugins-core --setopt=tsflags=nocontexts --setopt=tsflags=nocontexts --setopt=tsflags=nocontexts
No matches found for the following disable plugin patterns: local, spacewalk, versionlock
Updating Subscription Management repositories.
Unable to read consumer identity

This system is not registered with an entitlement server. You can use subscription-manager to register.

Copr repository                                  68 kB/s | 3.3 kB     00:00    
Fedora - ELN - Developmental packages for the n 1.5 kB/s | 196  B     00:00    
Errors during downloading metadata for repository 'eln':
  - Status code: 404 for https://odcs.fedoraproject.org/composes/production/latest-Fedora-ELN/compose/Everything/x86_64/os/repodata/repomd.xml (IP: 2620:52:3:1:dead:beef:cafe:fed6)
Error: Failed to download metadata for repo 'eln': Cannot download repomd.xml: Cannot download repodata/repomd.xml: All mirrors were tried
```

See https://dashboard.packit.dev/results/copr-builds/393591 for an example.

@davide125 any idea what's going on here? Should we just disable these?

davide125:
Probably a bad compose, I'll follow up with upstream and see what's going on there.

davide125:
This is due to the repo layout changing, it will be fixed once https://github.com/rpm-software-management/mock/commit/aa512e124775a43d5ac6e0cfbf8c75a4e2f84eb0 gets into production

osandov:
Looks like we're making it further now, but we're missing `libkdumpfile-devel` on x86-64, aarch64, ppc64le, and s390x:

```
Copr repository                                  62 kB/s | 3.3 kB     00:00    
Fedora - ELN BaseOS - Developmental packages fo  41 kB/s | 2.3 kB     00:00    
Fedora - ELN AppStream - Developmental packages  42 kB/s | 2.3 kB     00:00    
Fedora - ELN CodeReady Linux Builders - Build p  40 kB/s | 2.3 kB     00:00    
[...]

PR #189, opened 2022-07-06
Partial support for floats which are not 32/64 bits

Svetlitski-FB:
Near-identical to #167, but adds partial support for floats (as opposed to integers) which ~are >64-bit, namely 80-bit x87 floats and 128-bit floats~ are not 32/64 bits.

### DIFF
--- a/libdrgn/object.c
+++ b/libdrgn/object.c
@@ -110,7 +110,7 @@ drgn_object_type_impl(struct drgn_type *type, struct drgn_type *underlying_type,
 						 "bit field must be integer");
 		}
 		if (ret->encoding == DRGN_OBJECT_ENCODING_FLOAT &&
-		    ret->bit_size != 32 && ret->bit_size != 64) {
+		    (ret->bit_size < 1 || ret->bit_size > 256)) {
 			return drgn_error_format(DRGN_ERROR_INVALID_ARGUMENT,
 						 "unsupported floating-point bit size (%" PRIu64 ")",
 						 ret->bit_size);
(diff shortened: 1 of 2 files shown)


Svetlitski-FB:
Looks like build failures are again just a result of CI, per #188.

libdrgn/object.c
```diff
@@ -110,7 +110,8 @@ drgn_object_type_impl(struct drgn_type *type, struct drgn_type *underlying_type,
 						 "bit field must be integer");
 		}
 		if (ret->encoding == DRGN_OBJECT_ENCODING_FLOAT &&
-		    ret->bit_size != 32 && ret->bit_size != 64) {
+		    ret->bit_size != 32 && ret->bit_size != 64 &&
+		    ret->bit_size != 80 && ret->bit_size != 128) {
```
osandov, line 114: It looks like [GCC](https://gcc.gnu.org/onlinedocs/gcc/Floating-Types.html) and [Clang](https://clang.llvm.org/docs/LanguageExtensions.html#half-precision-floating-point) both also have a 16-bit floating-point type. Maybe rather than allowing a few specific sizes here, we can be permissive and allow anything within a "reasonable" range here like we do for integers (perhaps 1-256?). Note that this means that the checks for value creation would have to check `bit_size != 32 && bit_size != 64` instead of `bit_size > 64`.

osandov (commented):
This looks good, just one suggestion about possibly making this more permissive.

libdrgn/object.c
Svetlitski-FB: Sounds good, I modified it to allow 1-256, as you mentioned. Conditions, error messages, and test-cases have been updated accordingly.

osandov:
Thank you!

PR #194, opened 2022-08-01
Add helpers for lockless single lists (llist).

imran-kn:
Kernel makes use of several lockless singly lists (free_ipc_list,
delayed_mntput_list etc.) so having some helpers to traverse
these lists can be useful.

Signed-off-by: Imran Khan <imran.f.khan@oracle.com>

### DIFF
--- a/drgn/helpers/linux/list.py
+++ b/drgn/helpers/linux/list.py
@@ -33,6 +33,13 @@ __all__ = (
     "validate_list",
     "validate_list_for_each",
     "validate_list_for_each_entry",
+    "llist_empty",
+    "llist_first_entry",
+    "llist_first_entry_or_null",
+    "llist_for_each",
+    "llist_for_each_entry",
+    "llist_is_singular",
+    "llist_next_entry",
 )
 
 
@@ -276,4 +283,98 @@ def hlist_for_each_entry(
     """
     type = head.prog_.type(type)
     for pos in hlist_for_each(head):
+        yield container_of(pos, type, member)
+
+
+def llist_empty(head: Object) -> bool:
+    """
+    Return whether a llist is empty.
+
+    :param head: ``struct llist_head *``
+    """
+    return head.first.value_() == 0x0
+
+
+def llist_is_singular(head: Object) -> bool:
+    """
+    Return whether an llist has only one element.
+
+    :param head: ``struct llist_head *``
+    """
+    return head.first.value_() != 0x0 and head.first.next.value_() == 0x0
+
+
+def llist_first_entry(head: Object, type: Union[str, Type], member: str) -> Object:
+    """
+    Return the first entry in an llist.
+
+    The list is assumed to be non-empty.
+
+    See also :func:`llist_first_entry_or_null()`.
+
+    :param head: ``struct llist_head *``
+    :param type: Entry type.
+    :param member: Name of llist_node member in entry type.
+    :return: ``type *``
+    """
+    return container_of(head.first, type, member)
+
+
+def llist_first_entry_or_null(
+    head: Object, type: Union[str, Type], member: str
+) -> Object:
+    """
+    Return the first entry in an llist or ``NULL`` if the llist is empty.
+
+    See also :func:`llist_first_entry()`.
+
+    :param head: ``struct llist_head *``
+    :param type: Entry type.
+    :param member: Name of list_node member in entry type.
+    :return: ``type *``
+    """
+    if head.first.value_() == 0x0:
+        return NULL(head.prog_, head.prog_.pointer_type(head.prog_.type(type)))
+    else:
+        return container_of(head.first, type, member)
+
+
+def llist_next_entry(pos: Object, member: str) -> Object:
+    """
+    Return the next entry in an llist.
+
+    :param pos: ``type*``
+    :param member: Name of list_node member in entry type.
+    :return: ``type *``
+    """
+    return container_of(getattr(pos, member).next, pos.type_.type, member)
+
+
+def llist_for_each(head: Object) -> Iterator[Object]:
+    """
+    Iterate over all of the nodes in an llist.
+
+    :param head: ``struct llist_head *``
+    :return: Iterator of ``struct llist_node *`` objects.
+    """
+    head = head.read_()
+    pos = head.first.read_()
+    while pos.value_() != 0x0:
+        yield pos
+        pos = pos.next.read_()
+
+
+def llist_for_each_entry(
+    type: Union[str, Type], head: Object, member: str
+) -> Iterator[Object]:
+    """
+    Iterate over all of the entries in an llist.
+
+    :param type: Entry type.
+    :param head: ``struct llist_head *``
+    :param member: Name of list_node member in entry type.
+    :return: Iterator of ``type *`` objects.
+    """
+    type = head.prog_.type(type)
+    for pos in llist_for_each(head):
         yield container_of(pos, type, member)
--- a/drgn/helpers/linux/llist.py
+++ b/drgn/helpers/linux/llist.py
@@ -1,0 +1,119 @@
+# Copyright (c) 2022, Oracle and/or its affiliates.
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+"""
+Lock-less NULL terminated single linked lists
+------------
+
+The ``drgn.helpers.linux.llist`` module provides helpers for working with the
+lockless single linked list implementations (``struct llist_head`` and ``struct
+llist_node``) in :linux:`include/linux/llist.h`.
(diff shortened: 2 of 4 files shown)


drgn/helpers/linux/list.py
```diff
@@ -277,3 +284,97 @@ def hlist_for_each_entry(
...
+def llist_for_each(head: Object) -> Iterator[Object]:
+    """
+    Iterate over all of the nodes in an llist.
+
+    :param head: ``struct llist_head *``
+    :return: Iterator of ``struct llist_node *`` objects.
+    """
+    head = head.read_()
+    pos = head.first.read_()
+    while pos.value_() != 0x0:
```
osandov, line 362: Slightly cleaner:

```suggestion
    while pos:
```

drgn/helpers/linux/list.py
```diff
@@ -277,3 +284,97 @@ def hlist_for_each_entry(
...
+    :return: ``type *``
+    """
+    return container_of(getattr(pos, member).next, pos.type_.type, member)
+
+
+def llist_for_each(head: Object) -> Iterator[Object]:
+    """
+    Iterate over all of the nodes in an llist.
+
+    :param head: ``struct llist_head *``
```
osandov, line 357: It looks like this macro in the kernel source actually takes a `struct llist_node *`:

```c

/**
 * llist_for_each - iterate over some deleted entries of a lock-less list
 * @pos:	the &struct llist_node to use as a loop cursor
 * @node:	the first entry of deleted list entries
 *
 * In general, some entries of the lock-less list can be traversed
 * safely only after being deleted from list, so start with an entry
 * instead of list head.
 *
 * If being used on entries deleted from lock-less list directly, the
 * traverse order is from the newest to the oldest added entry.  If
 * you want to traverse from the oldest to the newest, you must
 * reverse the order by yourself before traversing.
 */
#define llist_for_each(pos, node)			\
	for ((pos) = (node); pos; (pos) = (pos)->next)
```

Obviously the rationale for this doesn't apply to drgn, but in general, drgn helpers mimic the kernel's calling conventions. So, maybe we should take a `struct llist_node *pos` instead of `struct llist_head *head` for consistency with the kernel. It would need to be called as `llist_for_each(head.first)` if you do have a `struct llist_head *`, which doesn't seem like a big deal.

drgn/helpers/linux/list.py
```diff
@@ -277,3 +284,97 @@ def hlist_for_each_entry(
...
+
+
+def llist_for_each_entry(
+    type: Union[str, Type], head: Object, member: str
+) -> Iterator[Object]:
+    """
+    Iterate over all of the entries in an llist.
+
+    :param type: Entry type.
+    :param head: ``struct llist_head *``
```
osandov, line 374: Same thing here re: taking a `struct llist_node *pos` instead.

drgn/helpers/linux/list.py
```diff
@@ -277,3 +284,97 @@ def hlist_for_each_entry(
...
+
+def llist_next_entry(pos: Object, member: str) -> Object:
+    """
+    Return the next entry in an llist.
+
+    :param pos: ``type*``
+    :param member: Name of list_node member in entry type.
+    :return: ``type *``
+    """
+    return container_of(getattr(pos, member).next, pos.type_.type, member)
```
osandov, line 350: None of these functions exist in the kernel. That doesn't necessarily mean that we shouldn't add them as helpers if they're useful, but I'm just curious whether you specifically wanted these or you just added them to duplicate the `list_*` helpers.

osandov (commented):
Awesome, thank you! I've run into `llist` before and I noticed that we didn't have any helpers for it, so this is very appreciated.

I have a couple of higher-level comments in addition to the ones I left inline:

* It looks like in the kernel source, these are in their own `llist.h` headers. We should mimic that and put these in their own `drgn.helpers.linux.llist` module. All you need to do is split them into their own `drgn/helpers/linux/llist.py` file and add an SPDX header and a module docstring.
* Could you please also add unit tests for the new helpers? You can add some `llist` instances to the [test kernel module](https://github.com/osandov/drgn/blob/main/tests/linux_kernel/kmod/drgn_test.c) for the test case to use. See a9815bb2877311699dba18557d18b9ee870def63 for an example (although note that the tests have since moved from `tests/helpers/linux` to `tests/linux_kernel/helpers`). You can run the tests locally in a VM  on the latest kernel version with `python3 setup.py test -k '*'`, or just push here for the CI to run it.

drgn/helpers/linux/list.py
```diff
@@ -277,3 +284,97 @@ def hlist_for_each_entry(
...
         yield container_of(pos, type, member)
+
+
+def llist_empty(head: Object) -> bool:
+    """
+    Return whether a llist is empty.
+
+    :param head: ``struct llist_head *``
+    """
+    return head.first.value_() == 0x0
```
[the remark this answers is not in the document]
imran-kn: Done

drgn/helpers/linux/list.py
```diff
@@ -277,3 +284,97 @@ def hlist_for_each_entry(
     type = head.prog_.type(type)
     for pos in hlist_for_each(head):
         yield container_of(pos, type, member)
+
+
+def llist_empty(head: Object) -> bool:
+    """
+    Return whether a llist is empty.
```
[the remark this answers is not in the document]
imran-kn: Corrected

drgn/helpers/linux/list.py
imran-kn: Done

drgn/helpers/linux/list.py
imran-kn: Sure, this makes sense. I have made these helpers consistent with respect to their kernel counter parts now. This would avoid confusion in using these helpers

drgn/helpers/linux/list.py
imran-kn: Done

drgn/helpers/linux/list.py
imran-kn: I am writing some scripts to dump per-CPU call_single_queue ( an llist) and per-CPU call function data. While writing that script the need for these helpers came up. From the above list only llist_next_entry is one which I have not used so far and so far it exists only as a result of duplication from list_* helpers. If needed I can remove this. Other helpers in this list seem helpful to me but I will await your feedback in this regard.

imran-kn:
I have addressed inline review comments. 
Also I have moved llist helpers to a separate file and have added test cases to test these helpers

imran-kn:
> Awesome, thank you! I've run into `llist` before and I noticed that we didn't have any helpers for it, so this is very appreciated.
> 
> I have a couple of higher-level comments in addition to the ones I left inline:
> 
>     * It looks like in the kernel source, these are in their own `llist.h` headers. We should mimic that and put these in their own `drgn.helpers.linux.llist` module. All you need to do is split them into their own `drgn/helpers/linux/llist.py` file and add an SPDX header and a module docstring.
> 
>     * Could you please also add unit tests for the new helpers? You can add some `llist` instances to the [test kernel module](https://github.com/osandov/drgn/blob/main/tests/linux_kernel/kmod/drgn_test.c) for the test case to use. See [a9815bb](https://github.com/osandov/drgn/commit/a9815bb2877311699dba18557d18b9ee870def63) for an example (although note that the tests have since moved from `tests/helpers/linux` to `tests/linux_kernel/helpers`). You can run the tests locally in a VM  on the latest kernel version with `python3 setup.py test -k '*'`, or just push here for the CI to run it.

Thank you for reviewing this. 

I have addressed inline comments.

Also I have moved llist helpers to a separate file and have added test cases for these helpers and these tests are passing as shown in below snippet:
`test_llist_empty (tests.linux_kernel.helpers.test_llist.TestLlist) ... ok
test_llist_first_entry (tests.linux_kernel.helpers.test_llist.TestLlist) ... ok
test_llist_first_entry_or_null (tests.linux_kernel.helpers.test_llist.TestLlist) ... ok
test_llist_for_each (tests.linux_kernel.helpers.test_llist.TestLlist) ... ok
test_llist_for_each_entry (tests.linux_kernel.helpers.test_llist.TestLlist) ... ok
test_llist_is_singular (tests.linux_kernel.helpers.test_llist.TestLlist) ... ok
test_llist_next_entry (tests.linux_kernel.helpers.test_llist.TestLlist) ... ok
`

imran-kn:
Hmm. Looks like I have messed this PR because of a wrong merge. I will clean a new PR with all of the changes that have been suggested so far

osandov:
Closing this copy since we continued in #195. In the future, feel free to force-push to your existing PR branch if you need to fix the history.

PR #197, opened 2022-08-10
Add name to drgn.Thread

### LINKED ISSUE #143: Add name to drgn.Thread
Follow-up to #92. Let's add a `drgn.Thread.name` attribute which is the name of the thread.

* For the Linux kernel, this can be retrieved from `struct task_struct::comm`.
* For live userspace processes, this can be retrieved from `/proc/$tid/comm` or `/proc/$pid/task/$tid/comm`.
* For userspace core dumps, the name of the main thread is saved as the `fname` field of the `PRPSINFO` note:

```console
$ eu-readelf -n crashme_multithreaded_core | grep -A 3 PRPSINFO
  CORE                 136  PRPSINFO
    state: 2, sname: D, zomb: 0, nice: 0, flag: 0x0000000040400404
    uid: 1000, gid: 985, pid: 39143, ppid: 39053, pgrp: 39143, sid: 39053
    fname: crashme_multith, psargs: ./crashme_multithreaded 
```

Unfortunately, the names of other threads don't appear to be saved in the core dump.

I think this is more or less the Python API we want:

```python
class Thread:
    name: Optional[str]
    """
    Thread name, or ``None`` if unknown.

    See `PR_SET_NAME
    <https://man7.org/linux/man-pages/man2/PR_SET_NAME.2const.html>`_
    and `/proc/[pid]/comm
    <https://man7.org/linux/man-pages/man5/proc_pid_comm.5.html>`_.

    .. note::
        Linux userspace core dumps only save the name of the main thread, so
        :attr:`name` will be ``None`` for other threads.
    """
```

This would be implemented in the Python bindings on top of a C API. There are a few options for the C API:

1. Get the thread name eagerly from the above sources, store it in `struct drgn_thread`, and return it from a simple getter `const char *drgn_thread_name(struct drgn_thread *thread)`. The returned string is owned by the `struct drgn_thread`.
2. Get the thread name when it is requested with `struct drgn_error *drgn_thread_name(struct drgn_thread *thread, char **ret)`. The caller is responsible for freeing the returned string.
3. Get the thread name when it is requested, but also cache it in `struct drgn_thread` for subsequent calls.

I'm leaning towards option 2, although it's up for debate.

rtadepalli:
This PR aims to resolve #143 

First off, my apologies for the delay. Had to step away for a bit and didn't have access to my personal laptop.

This is also my first time working in a project that interfaces Python with C, exciting stuff! I wrote some stuff to try out fetching the name of the thread for userspace programs. I think it should work, or outright crash if there are some errors while opening the file or using string functions but instead I get this:
image

The other functions and attributes of threads check out for a userspace program (stack unwinding not being available, `thread.object` not available outside the kernel) but  I don't get why `name` isn't recognized as a valid attribute after i added it to the interface. This may actually be a naive question but I am quite confused by this.

Learning a lot through this project though! 

@osandov

### DIFF
--- a/libdrgn/program.c
+++ b/libdrgn/program.c
@@ -1188,6 +1188,24 @@ drgn_program_find_thread_userspace_live(struct drgn_program *prog, uint32_t tid,
 		(*ret)->prog = prog;
 		(*ret)->tid = tid;
 		(*ret)->prstatus = (struct nstring){};
+
+		const int path_len = strlen(path);
+		const int comm_file_path_len = path_len + 6;
+		char *comm_file_path = malloc(comm_file_path_len);
+
+		if (!comm_file_path)
+			return &drgn_enomem;
+		strncpy(comm_file_path, path, path_len);
+		strncat(comm_file_path, "/comm", 5);
+		strncat(comm_file_path, "\0", 1);
+		
+		FILE *comm_file = fopen(comm_file_path, "r");
+		if (!comm_file)
+			return drgn_error_create_os("access", errno, path);
+		fscanf(comm_file, "%s", (*ret)->name);
+		fclose(comm_file);
+		free(comm_file_path);
+
 		return NULL;
 	} else if (errno == ENOENT) {
 		*ret = NULL;
--- a/_drgn.pyi
+++ b/_drgn.pyi
@@ -847,6 +847,19 @@ class FindObjectFlags(enum.Flag):
 class Thread:
     """A thread in a program."""
 
+    name: Optional[str]
+    """
+    Thread name, or ``None`` if unknown.
+
+    See `PR_SET_NAME
+    <https://man7.org/linux/man-pages/man2/prctl.2.html#:~:text=PR_SET_NAME>`_
+    and `/proc/[pid]/comm
+    <https://man7.org/linux/man-pages/man5/proc.5.html#:~:text=%2Fproc%2F%5Bpid%5D%2Fcomm>`_.
+
+    .. note::
+        Linux userspace core dumps only save the name of the main thread, so
+        :attr:`name` will be ``None`` for other threads.
+    """
     tid: int
     """
     Thread ID (as defined by `gettid(2)
--- a/libdrgn/program.h
+++ b/libdrgn/program.h
@@ -46,6 +46,7 @@ struct drgn_symbol;
 
 struct drgn_thread {
 	struct drgn_program *prog;
+	char name[64];
 	uint32_t tid;
 	struct nstring prstatus;
 	struct drgn_object object;


libdrgn/program.c
```diff
@@ -1188,6 +1188,24 @@ drgn_program_find_thread_userspace_live(struct drgn_program *prog, uint32_t tid,
...
+
+		const int path_len = strlen(path);
+		const int comm_file_path_len = path_len + 6;
+		char *comm_file_path = malloc(comm_file_path_len);
+
+		if (!comm_file_path)
+			return &drgn_enomem;
+		strncpy(comm_file_path, path, path_len);
+		strncat(comm_file_path, "/comm", 5);
+		strncat(comm_file_path, "\0", 1);
```
osandov, line 1200: We can avoid the extra allocation and all of this string processing if we reserve enough space in `path` to append `"/comm"`:

```diff
diff --git a/libdrgn/program.c b/libdrgn/program.c
index 1a5103e..83b5a21 100644
--- a/libdrgn/program.c
+++ b/libdrgn/program.c
@@ -1177,7 +1177,7 @@ drgn_program_find_thread_userspace_live(struct drgn_program *prog, uint32_t tid,
 		- sizeof("%ld%" PRIu32)
 		+ max_decimal_length(long)
 		+ max_decimal_length(uint32_t)
-		+ 1];
+		+ sizeof("/comm")];
 	snprintf(path, sizeof(path), FORMAT, (long)prog->pid, tid);
 #undef FORMAT
 	int r = access(path, F_OK);
@@ -1188,6 +1188,11 @@ drgn_program_find_thread_userspace_live(struct drgn_program *prog, uint32_t tid,
 		(*ret)->prog = prog;
 		(*ret)->tid = tid;
 		(*ret)->prstatus = (struct nstring){};
+
+		strcat(path, "/comm");
+		FILE *comm_file = fopen(path, "r");
+		...
+
 		return NULL;
 	} else if (errno == ENOENT) {
 		*ret = NULL;
```

(Note that `sizeof("/comm")` includes the null terminator, and `strcat()` copies the null terminator).

libdrgn/program.c
```diff
@@ -1188,6 +1188,24 @@ drgn_program_find_thread_userspace_live(struct drgn_program *prog, uint32_t tid,
...
+		
+		FILE *comm_file = fopen(comm_file_path, "r");
+		if (!comm_file)
+			return drgn_error_create_os("access", errno, path);
```
osandov, line 1204: We need to free the thread we just allocated if this fails. Also, the error message needs to be updated to `"fopen"`.

```suggestion
		if (!comm_file) {
			free(*ret);
			return drgn_error_create_os("fopen", errno, path);
		}
```

libdrgn/program.c
```diff
@@ -1188,6 +1188,24 @@ drgn_program_find_thread_userspace_live(struct drgn_program *prog, uint32_t tid,
...
+		fscanf(comm_file, "%s", (*ret)->name);
```
osandov, line 1205: A few things here:

1. If `comm` is ever increased to be more than 64 bytes (unlikely but theoretically possible), this will overflow the buffer.
2. I believe this will have a trailing newline that needs to be removed.
3. We should check if there was an error reading from the file.

I think these would be easier to fix by using `fread()` instead of `fscanf()`.

osandov (commented):
Thanks for looking into this! I left a few comments inline.

The Python bindings need to be explicitly told about the new attribute. You'll need to add a new [getter](https://docs.python.org/3/c-api/typeobj.html#c.PyTypeObject.tp_getset) to `Thread_getset`: https://github.com/osandov/drgn/blob/42966530906b4eff00181cb880687688509ede47/libdrgn/python/thread.c#L74-L78

You'll want to look at [`PyBytes_FromString()`](https://docs.python.org/3/c-api/bytes.html#c.PyBytes_FromString) to return `bytes` or [`PyUnicode_DecodeFSDefault()`](https://docs.python.org/3/c-api/unicode.html#c.PyUnicode_DecodeFSDefault) to return `str`. ([`PyUnicode_FromString()`](https://docs.python.org/3/c-api/unicode.html#c.PyUnicode_FromString) is the typical function to create a `str` from a C string, but since the thread name is not guaranteed to be valid UTF-8, I think `PyUnicode_DecodeFSDefault()` is more appropriate here.)

osandov:
@rtadepalli are you still planning on working on this?

osandov:
I'm going to close this and put the issue back up for grabs, but let me know if you'd like to pick it back up.

Issue #199, opened 2022-08-12
[Request] Implement pretty printing support in Jupyter notebook


shunghsiyu:
[Jupyter notebook and drgn](https://twitter.com/omsandov/status/1557822779755442176/photo/1) seems to be a really great combination, it'd would be nice to see pretty printing of drgn object in Jupyter notebook implemented for a even better synergy (similar to what's being done in [interactive mode](https://drgn.readthedocs.io/en/latest/user_guide.html#interactive-mode)).

In theory implementing [`_repr_pretty()`](https://ipython.readthedocs.io/en/stable/config/integrating.html) is all that's needed, something like what [stackoverflow suggests](https://stackoverflow.com/a/41454816), but done with C inside libdrgn/python/object.c.

```python
    def _repr_pretty_(self, p, cycle):
       p.text(str(self) if not cycle else '...')
```

osandov:
I'm closing this one now since your pretty printing support was previously merged. #318 will track further integration.

PR #200, opened 2022-08-15
Pretty printing support in Jupyter notebook

shunghsiyu:
The PR implements #199 to support pretty printing in Jupyter notebook by adding the `_repr_pretty()` method to the base classes, using `str()` for the heavy lifting.

This save the user from having to call `print()` inside Jupyter notebook, similar to what's being done in [interactive mode](https://drgn.readthedocs.io/en/latest/user_guide.html#interactive-mode).


tests/test_object.py
```diff
@@ -1718,3 +1719,13 @@ def test_iter(self):
             iter,
             Object(self.prog, "int []", address=0),
         )
+
+    def test__repr_pretty(self):
+        obj = Object(self.prog, "int", value=0)
+        pretty_printer_mock = unittest.mock.Mock()
```
shunghsiyu, line 1725: I hesitate a bit about using the Mock class from unittest as I don't see Mock being uses elsewhere, not sure if there's a better alternative.

shunghsiyu:
Re-push to correct the method name to `_repr_pretty_()` (previously it's missing the underscore at the end).

Still need to implement that same thing for `StackFrame`, `StackTrace`, and `Type`.

libdrgn/python/object.c
```diff
@@ -908,6 +908,30 @@ static PyObject *DrgnObject_str(DrgnObject *self)
...
+	int cycle;
+
+	if (!PyArg_ParseTupleAndKeywords(args, kwds, "Op:_repr_pretty_", keywords,
+					 &p, &cycle))
+		return NULL;
+
+	if (cycle)
+		return PyObject_CallMethod(p, "text", "s", "...");
+
+	str_obj = DrgnObject_str(self);
```
osandov, line 925: If you use `PyObject_Str()` instead of `DrgnObject_str` (and make `self` a `PyObject *`), then you can reuse the same implementation for all of the types that need it.

tests/test_object.py
osandov: I think Mock makes sense for this test.

osandov (commented):
Thanks for doing this! It looks straightforward, but I haven't gotten to test it yet.

libdrgn/python/object.c
osandov: To elaborate on this, you could name this function something more generic (`repr_pretty_from_str()` or something like that?) put it somewhere common (maybe `libdrgn/module/python.c`), and use it in the method table for the other classes.

shunghsiyu:
Thanks for the suggestion and especially the elaboration of it, really useful since this is the first time I'm trying to work with Python C-extension.

Here's the changes I've made:
- Generalize the original method into `libdrgn/python/util.c:repr_pretty_from_str()`
    - Move docstring into `libdrgn/python/drgnpy.h`
- Add `_repr_pretty_()` to `Type`, `StackTrace`, and `StackFrame` as well
- Generalize the original test into `assertReprPrettyEqualsStr` test helper, and add test for the above classes

btw the [vmtest](/osandov/drgn/tree/main/vmtest) kernel testing facility drgn has is awesome!

tests/linux_kernel/test_stack_trace.py
```diff
@@ -97,3 +98,13 @@ def test_prog(self):
             self.prog.stack_trace(Object(self.prog, "struct pt_regs", value={})).prog,
             self.prog,
         )
+
+    def test_stack__repr_pretty_(self):
```
shunghsiyu, line 102: We don't really need to run the `_repr_pretty_()` test with root privilege and the kernel module loaded.

Do we want to mock `StackTrace` and `StackFrame` instead? Says in `tests/test_stack_trace.py`.

shunghsiyu:
Sorry for the force pushes, I regret not testing it somewhere else first. (Every time I thought I finally got reStructuredText right, I was proven wrong...)

Anyway added another commit to documentation about how drgn can be used with [Jupyter](https://jupyter.org), as well as what feature it provides. It's a bit of Jupyter evangelism, admittedly, which I'm having problem toning down :sweat_smile:. Feel free to edit it as you see fit.

tests/linux_kernel/test_stack_trace.py
osandov: We don't have a great way to test stack traces otherwise right now, so I think it's fine to do it here for now.

osandov:
Thanks for this! I pushed the first two commits for `_repr_pretty_()` (with minor edits) for now. I want to give the documentation a little more thought before merging it.

shunghsiyu:
Thanks!

Regarding the documentation changes, I'm also working on something that might better showcase Jupyter & drgn. I'll post it here when it's ready, and then perhaps we can have a more concrete discussion.

PR #202, opened 2022-08-17
Add identify_address helper function

nhatsmrt:
This helper function identifies the type of the address (slab allocated or symbol) and returns a string representation of the address accordingly. This will be useful for another helper function which prints the stack trace with more information about each item on the stack.


drgn/helpers/memory.py
```diff
@@ -0,0 +1,104 @@
...
+from drgn import IntegerLike, Object, Program, SymbolKind
+from drgn.helpers import escape_ascii_string
+from drgn.helpers.linux.slab import find_containing_slab_cache
+
+__all__ = ("identify_address",)
+
+
+SYMBOL_KIND_STR = {
+    SymbolKind.UNKNOWN: "symbol",
+    SymbolKind.OBJECT: "symbol object",
```
osandov, line 24: `object symbol` reads more naturally to me than `symbol object`. Can you swap the wording for all of the kinds?

drgn/helpers/memory.py
```diff
@@ -0,0 +1,104 @@
...
+    SymbolKind.FUNC: "symbol function",
+    SymbolKind.SECTION: "symbol section",
+    SymbolKind.FILE: "symbol file",
+    SymbolKind.COMMON: "symbol common",
+    SymbolKind.TLS: "symbol tls",
+    SymbolKind.IFUNC: "symbol ifunc",
```
osandov, line 30: Actually, I think only object and function symbols deserve special treatment; the rest of the kinds are really obscure. You can delete the rest, then use `_SYMBOL_KIND_STR.get(symbol.kind, "symbol")` instead of `_SYMBOL_KIND_STR[symbol.kind]` so that anything else gets the default `"symbol"`.

```suggestion
    SymbolKind.OBJECT: "object symbol",
    SymbolKind.FUNC: "function symbol",
```

drgn/helpers/memory.py
```diff
@@ -0,0 +1,104 @@
...
+
+@overload
+def identify_address(addr: Object) -> Optional[str]:
+    """
+    Identify an address (given as an :class:`.Object`) as a symbol or slab object.
+
+    This will currently identify the following types of addresses as follows:
+
+    * Symbols (e.g., addresses in functions or global variables): `symbol symbol_kind symbol_name+offset` or `symbol symbol_kind symbol` if offset is 0.
+    * Slab objects: `slab object slab_cache_name`.
```
osandov, line 42: This is inconsistent with the code, which has an additional `:`. I like the colon, so let's keep that and update the documentation. In fact, I think it'd be nice to add a colon for symbols, as well (i.e., `function symbol: foo`).

drgn/helpers/memory.py
```diff
@@ -0,0 +1,104 @@
...
+    SymbolKind.FILE: "symbol file",
+    SymbolKind.COMMON: "symbol common",
+    SymbolKind.TLS: "symbol tls",
+    SymbolKind.IFUNC: "symbol ifunc",
+}
+
+
+@overload
+def identify_address(addr: Object) -> Optional[str]:
+    """
```
osandov, line 36: After your last PR made me realize how much it sucked to duplicate the documentation, I made it so that's not necessary. See https://github.com/osandov/drgn/commit/5b6a8c27a9586052b3f81d768880e4c39ac3414e; you can replace the docstring in the first overload with an empty one (`""""""`) and then put all of the documentation in the second one.

osandov (changes requested):
Thank you, this looks great! Just a few nitpicks. Additionally, we should make the CLI automatically import everything from `drgn.helpers.memory` (for any program type) like it does for `drgn.helpers.linux` for the kernel. See https://github.com/osandov/drgn/blob/30c9ad452db65d745cfecea12ce8e46b1f829a80/drgn/internal/cli.py#L235-L239 and let me know if you need more pointers.

osandov:
Actually, scratch my last comment about the automatic import, that's going to get out of hand with more modules so I need to think of a better way. But the rest of my comments still stand.

drgn/helpers/memory.py
```diff
@@ -0,0 +1,104 @@
...
+    Identify an address (given as an :class:`.Object`) as a symbol or slab object.
+
+    This will currently identify the following types of addresses as follows:
+
+    * Symbols (e.g., addresses in functions or global variables): `symbol symbol_kind symbol_name+offset` or `symbol symbol_kind symbol` if offset is 0.
```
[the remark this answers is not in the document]
nhatsmrt: Offset converted to hexadecimal, and added in the case it's 0 too!

drgn/helpers/memory.py
```diff
@@ -0,0 +1,104 @@
...
+
+import drgn
+from drgn import IntegerLike, Object, Program, SymbolKind
+from drgn.helpers import escape_ascii_string
+from drgn.helpers.linux.slab import find_containing_slab_cache
+
+__all__ = ("identify_address",)
+
+
+SYMBOL_KIND_STR = {
```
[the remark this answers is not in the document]
nhatsmrt: Fixed.

drgn/helpers/memory.py
nhatsmrt: Fixed.

drgn/helpers/memory.py
nhatsmrt: Fixed.

drgn/helpers/memory.py
nhatsmrt: Fixed.

drgn/helpers/memory.py
nhatsmrt: Oh wow that's much better, fixed!

drgn/helpers/common/memory.py
```diff
@@ -0,0 +1,88 @@
...
+@overload
+def identify_address(addr: Object) -> Optional[str]:
+    """"""
+    ...
+
+
+@overload
+def identify_address(prog: Program, addr: IntegerLike) -> Optional[str]:
+    """
+    Identify an address (given as a :class:`.Program` and an integer) as a symbol or slab object.
```
osandov, line 37: How about we make this more generic:

```suggestion
    Try to identify what an address refers to.
```

(Yeah I know the original was my wording :stuck_out_tongue:)

drgn/helpers/common/memory.py
```diff
@@ -0,0 +1,88 @@
...
+
+    This will currently identify the following types of addresses as follows:
+
+    * Symbols (e.g., addresses in functions or global variables): `symbol_kind: symbol_name+offset`.
+    * Slab objects: `slab object: slab_cache_name`.
+
+    This may learn to recognize other types of addresses in the future.
```
osandov, line 44: Looking at this again, it could use more clarification:

```suggestion
    For all programs, this will identify addresses as follows:

    * Object symbols (e.g., addresses in global variables):
      ``object symbol: {symbol_name}+{hex_offset}``.
    * Function symbols (i.e., addresses in functions):
      ``function symbol: {symbol_name}+{hex_offset}``.
    * Other symbols: ``symbol: {symbol_name}+{hex_offset}``.

    Additional, for the Linux kernel, this will identify:

    * Slab objects: ``slab object: {slab_cache_name}``.

    This may recognize other types of addresses in the future.
```
```

File: drgn/helpers/common/memory.py [hunk 9]
```diff
@@ -0,0 +1,88 @@
...
+
+    The address can be given as an :class:`~drgn.Object` or as a
+    :class:`~drgn.Program` and an integer.
+
+    :param addr: ``void *``
+    :return: Identity as string or ``None` if the address is unrecognized.
```
osandov, line 50: Missing a comma and a backtick here (again, my fault).

```suggestion
    :return: Identity as string, or ``None`` if the address is unrecognized.
```

drgn/helpers/common/memory.py
```diff
@@ -0,0 +1,88 @@
...
+    else:
+        assert isinstance(prog_or_addr, Program)
+        prog = prog_or_addr
+    addr = operator.index(addr)
+
+    if prog.flags & drgn.ProgramFlags.IS_LINUX_KERNEL:
+        # Linux kernel-specific identification:
+        slab_cache = find_containing_slab_cache(prog, addr)
+
+        if slab_cache.value_():
```
osandov, line 71: You can check the truthiness of a `drgn.Object` directly:

```suggestion
        if slab_cache:
```

osandov (changes requested):
This is almost ready! Left one comment in the code, and a few comments to improve the documentation that I originally gave you. Additionally, could you also rename `test_identify_address.py` to `test_common.py`? Then we can use that for all tests of Linux kernel-specific behavior in common helpers.

drgn/helpers/common/memory.py
nhatsmrt: Fixed

drgn/helpers/common/memory.py
nhatsmrt: Fixed 😛

drgn/helpers/common/memory.py
nhatsmrt: Expanded!

drgn/helpers/common/memory.py
nhatsmrt: Yeah I don't claim originality for any of the documentation 😛 Fixed!

nhatsmrt:
Rename: `tests/linux_kernel/helpers/test_identify_address.py -> tests/linux_kernel/helpers/test_common.py`

drgn/helpers/common/memory.py
```diff
@@ -0,0 +1,95 @@
...
+
+    For all programs, this will identify addresses as follows:
+
+    * Object symbols (e.g., addresses in global variables):
+      ``object symbol: {symbol_name}+{hex_offset}``.
+    * Function symbols (i.e., addresses in functions):
+      ``function symbol: {symbol_name}+{hex_offset}``.
+    * Other symbols: ``symbol: {symbol_name}+{hex_offset}``.
+
+    Additional, for the Linux kernel, this will identify:
```
osandov, line 47: Oops, I made a typo here (this should be "Additionally"). I can fix it on my end.

osandov (approved):
Alright force-pushed a fix to the typo. I'll merge this now, thanks!

PR #203, opened 2022-08-18
libdrgn: Zero fill excluded pages in kernel core dumps rather than FaultError

gkmccready:
makecoredump will exclude zero pages. We found a core file where a
structure straddled a page boundary and the end of the structure
was all zeros so the page was excluded and we were generating a
FaultError trying to access the structure.

This change reverts a portion of that behaviour such that when we are
debugging a kernel core we go back to the zero fill behaviour. To do this
we go back to creating segments based on memsz instead of filesz and
handling the filesz->memsz gap in drgn_read_memory_file.

Fixes: 02912ca7d073 ("libdrgn: fix handling of p_filesz < p_memsz in core dumps")
Signed-off-by: Glen McCready <gkm@mysteryinc.ca>

### DIFF
--- a/libdrgn/memory_reader.c
+++ b/libdrgn/memory_reader.c
@@ -273,19 +273,20 @@ struct drgn_error *drgn_read_memory_file(void *buf, uint64_t address,
 					 void *arg, bool physical)
 {
 	struct drgn_memory_file_segment *file_segment = arg;
-
-	if (offset > file_segment->file_size ||
-	    count > file_segment->file_size - offset) {
-		if (offset <= file_segment->file_size)
-			address += file_segment->file_size - offset;
-		return drgn_error_create_fault("memory not saved in core dump",
-					       address);
-	}
-
 	uint64_t file_offset = file_segment->file_offset + offset;
+	uint64_t file_end = file_segment->file_offset + file_segment->file_size;
 	char *p = buf;
+
+	if (!file_segment->zerofill && file_offset + count > file_end)
+		return drgn_error_create_fault("memory not saved in core dump",
+					       address + file_segment->file_size);
 	while (count) {
-		ssize_t ret = pread(file_segment->fd, p, count, file_offset);
+		size_t readlen = min(file_end - file_offset, count);
+		ssize_t ret;
+
+		if (readlen == 0)
+			break;
+		ret = pread(file_segment->fd, p, readlen, file_offset);
 		if (ret == -1) {
 			if (errno == EINTR) {
 				continue;
@@ -304,5 +305,7 @@ struct drgn_error *drgn_read_memory_file(void *buf, uint64_t address,
 		count -= ret;
 		file_offset += ret;
 	}
+	/* p_filesz < p_memsz case; end of file reached, zero fill the buffer */
+	memset(p, '\0', count);
 	return NULL;
 }
--- a/libdrgn/program.c
+++ b/libdrgn/program.c
@@ -433,14 +433,13 @@ drgn_program_set_core_dump(struct drgn_program *prog, const char *path)
 		prog->file_segments[j].file_size = phdr->p_filesz;
 		prog->file_segments[j].fd = prog->core_fd;
 		prog->file_segments[j].eio_is_fault = false;
+		/*
+		 * Only zerofill between p_filesz and p_memsz for saved
+		 * kernel cores. makedumpfile excludes zero pages; userland
+		 * and /proc/kcore are not affected.
+		 */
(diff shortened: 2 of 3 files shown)


libdrgn/program.c
```diff
@@ -433,14 +433,20 @@ drgn_program_set_core_dump(struct drgn_program *prog, const char *path)
...
 		prog->file_segments[j].fd = prog->core_fd;
 		prog->file_segments[j].eio_is_fault = false;
+		/*
+		 * vmcoreinfo_note is a good proxy for the current core being a
+		 * kernel core and we use it to decide whether to zerofill pages
+		 * between p_filesz and p_memsz. We do this because at this
+		 * point only makecoredump excludes zero pages; userland is
+		 * not affected.
+		 */
+		prog->file_segments[j].zerofill = vmcoreinfo_note != NULL;
```
osandov, line 443: It looks like `/proc/kcore` always sets `filesz` to the same value as `memsz`: https://github.com/torvalds/linux/blob/072e51356cd5a4a1c12c1020bc054c99b98333df/fs/proc/kcore.c#L396

I don't see a reason why `filesz` would ever be less than `memsz` for `/proc/kcore`. So, I'd rather be conservative and exclude it from the zero filling behavior.

```suggestion
		prog->file_segments[j].zerofill = vmcoreinfo_note && !is_proc_kcore;
```

libdrgn/program.c
```diff
@@ -433,14 +433,20 @@ drgn_program_set_core_dump(struct drgn_program *prog, const char *path)
...
 		err = drgn_program_add_memory_segment(prog, phdr->p_vaddr,
 						      /*
 						       * Don't override the page
 						       * table reader for
 						       * unsaved regions.
 						       */
```
osandov, line 449: This comment is no longer relevant.

```suggestion
```

osandov (commented):
Thank you for the fix! This looks correct, but I left a couple of suggestions for minor improvements.

libdrgn/program.c
gkmccready: I think I addressed everything; let me know how you feel about the new PR.

libdrgn/memory_reader.c
```diff
@@ -273,19 +273,20 @@ struct drgn_error *drgn_read_memory_file(void *buf, uint64_t address,
...
+	if (!file_segment->zerofill && file_offset + count > file_end)
 		return drgn_error_create_fault("memory not saved in core dump",
 					       address);
-	}
-
-	uint64_t file_offset = file_segment->file_offset + offset;
-	char *p = buf;
 	while (count) {
-		ssize_t ret = pread(file_segment->fd, p, count, file_offset);
+		int readlen = min(file_end - file_offset, count);
```
osandov, line 284: In theory, this can be over `INT_MAX`, so this should be `size_t`:

```suggestion
		size_t readlen = min(file_end - file_offset, count);
```

osandov (changes requested):
Looking at the CI results, it looks like this fails the test case for the "memory not saved in core dump address":

```
  FAIL: test_unsaved (tests.test_program.TestCoreDump)
    File "/home/runner/work/drgn/drgn/tests/test_program.py", line 875, in test_unsaved
      self.assertEqual(cm.exception.address, 0xFFFF000C)
  AssertionError: 4294901760 != 4294901772
```

It's returning the address of the beginning of the segment rather than the subset that is not saved. Other than that, I left one small comment.

gkmccready:
Thanks for the CI re-run, @osandov -- let me know if there's anything left to address. Thank you!

libdrgn/program.c
```diff
@@ -433,14 +433,13 @@ drgn_program_set_core_dump(struct drgn_program *prog, const char *path)
 		prog->file_segments[j].file_size = phdr->p_filesz;
 		prog->file_segments[j].fd = prog->core_fd;
 		prog->file_segments[j].eio_is_fault = false;
+		/*
+		 * Only zerofill between p_filesz and p_memsz for saved
+		 * kernel cores. makecoredump excludes zero pages; userland
```
osandov, line 438: I missed this earlier, but the tool is called `makedumpfile`. The same thing needs to be fixed in the commit message.

```suggestion
		 * kernel cores. makedumpfile excludes zero pages; userland
```

osandov (changes requested):
Noticed a last minute typo in the commit message and comment. There's also an empty merge commit in your branch. To keep the git history flat, you should be able to get rid of it by rebasing on the `osandov:main` branch.

osandov (approved):
Perfect, thank you for your patience!

PR #204, opened 2022-08-18
Fix spelling error

michel-slm:
abbrevation => abbreviation; caught by Debian's lintian

Signed-off-by: Michel Alexandre Salim <michel@michel-slm.name>


osandov:
Ha, thanks! Was that the only typo it found? If so, I'm shocked! I'm not sure what's going on with Packit, but this PR is obviously okay.

PR #209, opened 2022-09-02
Add helpers for Linux Traffic Control (TC) filter

ypl-coffee:
Hi all,

These are helpers for Linux TC filter (`struct tcf_proto`) and filter chain (`struct tcf_chain`) that I found useful when debugging TC issues.  They are only supported since Linux v4.13 because several related data structures didn't exist before that.

The test is skipped if kernel is older than v4.13 commit 5bc1701881e3 ("net: sched: introduce multichain support for filters"), and depends on config `CONFIG_NET_SCH_HTB` and `CONFIG_NET_CLS_U32`.  It uses the `TCA_CHAIN` Netlink message attribute, which pyroute2 doesn't fully support yet, which is why the test may look a bit over-complicated.  I will add `TCA_CHAIN` support to pyroute2 and clean it up.

Thanks,
Peilin Ye


ypl-coffee:
Fixed a `nlmsg` not found issue found by the rpm-build job.

drgn/helpers/linux/tc.py
```diff
@@ -10,11 +10,102 @@
...
+    # Before Linux kernel commit 5bc1701881e3 ("net: sched: introduce
+    # multichain support for filters") (in v4.13), each block contained only
+    # one chain.
+    try:
+        chain_list = block.chain_list.address_of_()
+    except AttributeError:
+        # Before Linux kernel commit 2190d1d0944f ("net: sched: introduce
+        # helpers to work with filter chains") (in v4.13), struct tcf_chain
+        # didn't exist.
+        return block.chain
```
osandov, line 45: It looks like both of these commits went in during v4.13-rc1. I wouldn't bother checking for the intermediate state between those two commits; let's just assume that the kernel is an official tagged release, which I think will simplify this.

drgn/helpers/linux/tc.py
```diff
@@ -10,11 +10,102 @@
...
+    "get_tcf_proto_by_prio",
+    "qdisc_lookup",
+)
+
+
+def for_each_tcf_chain(block: Object) -> Iterator[Object]:
+    """
+    Iterate over all TC filter chains on a block.
+
+    This is only supported since Linux v4.13.
```
osandov, line 31: I'm not familiar with TC at all, but looking at the commits you reference here, it looks like TC filter chains didn't exist at all before 4.13, is that correct? Maybe we can say something like "This is only supported since Linux v4.13, before which TC filter chains didn't exist.".

drgn/helpers/linux/tc.py
```diff
@@ -10,11 +10,102 @@
...
+    try:
+        chain_list = block.chain_list.address_of_()
+    except AttributeError:
+        # Before Linux kernel commit 2190d1d0944f ("net: sched: introduce
+        # helpers to work with filter chains") (in v4.13), struct tcf_chain
+        # didn't exist.
+        return block.chain
+
+    for chain in list_for_each_entry("struct tcf_chain", chain_list, "list"):
+        yield chain
```
osandov, line 48: I think you can simplify this to:

```suggestion
    return list_for_each_entry("struct tcf_chain", chain_list, "list")
```

osandov (commented):
I only gave this a quick look and had some questions about the kernel versions involved. I'll give it a more thorough review once that's cleared up. Thank you!

PR #212, opened 2022-09-19
S390 support

svens-s390:
This add minimum support for IBM Z / The s390 architecture.

### DIFF
--- a/_drgn.pyi
+++ b/_drgn.pyi
@@ -959,6 +959,9 @@ class Architecture(enum.Enum):
     RISCV32 = ...
     """The 32-bit RISC-V architecture."""
 
+    S390 = ...
+    """The s390 architecture, a.k.a. IBM Z."""
+
     UNKNOWN = ...
     """
     An architecture which is not known to drgn. Certain features are not
--- a/libdrgn/Makefile.am
+++ b/libdrgn/Makefile.am
@@ -37,6 +37,7 @@ libdrgnimpl_la_SOURCES = $(ARCH_DEFS_PYS:_defs.py=.c) \
 			 arch_arm.c \
 			 arch_i386.c \
 			 arch_riscv.c \
+			 arch_s390.c \
 			 array.h \
 			 binary_buffer.c \
 			 binary_buffer.h \
--- a/libdrgn/Makefile.am
+++ b/libdrgn/Makefile.am
@@ -17,7 +17,8 @@ include_HEADERS = drgn.h
 
 ARCH_DEFS_PYS = arch_aarch64_defs.py \
 		arch_ppc64_defs.py \
-		arch_x86_64_defs.py
+		arch_x86_64_defs.py \
+		arch_s390_defs.py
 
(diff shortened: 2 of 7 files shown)


_drgn.pyi
```diff
@@ -959,6 +959,9 @@ class Architecture(enum.Enum):
     RISCV32 = ...
     """The 32-bit RISC-V architecture."""
 
+    S390 = ...
+    """The s390 architecture, a.k.a. IBM Z."""
```
osandov, line 963: I've seen the names s390 and s390x. Would you happen to know what the difference is?

libdrgn/Makefile.am
```diff
@@ -17,7 +17,8 @@ include_HEADERS = drgn.h
 
 ARCH_DEFS_PYS = arch_aarch64_defs.py \
 		arch_ppc64_defs.py \
-		arch_x86_64_defs.py
+		arch_x86_64_defs.py \
+		arch_s390_defs.py
```
osandov, line 21: I think you may have forgotten to `git add` `arch_s390_defs.py`, I don't see it here.

Also, would you mind putting s390 before x86_64 to keep this alphabetically sorted?

libdrgn/arch_s390.c
```diff
@@ -0,0 +1,214 @@
...
+		return err;
+
+	nextbc = *(uint64_t *)(frame + bit_offset / 8);
+	if (!nextbc)
+		return &drgn_stop;
+	drgn_register_state_set_from_u64(prog, *ret, bc, nextbc);
+	return NULL;
+}
+
+static struct drgn_error *
```
osandov, line 159: Could you please add an explanation and reference to the corresponding kernel code? See the AArch64 version for an example: https://github.com/osandov/drgn/blob/330c71b5b59906c1336df9f2de19257c20f749ab/libdrgn/arch_aarch64.c#L194-L199

osandov (commented):
Wow, this is awesome! I love seeing support for additional architectures. I still have to look at the relocation and stack unwinding code in detail, but I left a few high-level comments to start. Thank you!

libdrgn/Makefile.am
svens-s390: Indeed. Sorry.

_drgn.pyi
svens-s390: s390 is the older architecture name when it was 31 bit. Somewhere around 2000 it was extended to 64 bit, and than it was called 's390x'. The kernel however kept the 's390' name as architecture, while userspace programs usually use s390x. As drgn is a kernel debugger, i would name it the same way as the architecture in the kernel.

libdrgn/arch_s390.c
svens-s390: On s390, we allocate a struct pt_regs on the stack when we enter the kernel. I haven't added the code to handle that yet, but will do when i submit the next version. The current code only handles the stack frame entries written when a function is called.

PR #214, opened 2022-09-28
S390 support/v2

svens-s390:
Hi,

this is v2 of the s390 support pull request. Changes:

- add code to handle struct pt_regs in irq/pgm check/system call entry stack frames
- add a comment where to find the kernel switch_to() source
- reorder arch files in Makefile.am


svens-s390:
I see that CI is failing. From reading the vmtest source, it loos like the tests are trying to fetch prebuilt kernels from github assets. Is that correct? What is the process to get kernels in there?

osandov:
The failure is actually in the "Build extension" step:

```
  make[1]: Entering directory '/home/runner/work/drgn/drgn/build/temp.linux-x86_64-cpython-310'
    CC       libdrgnimpl_la-arch_aarch64.lo
    CC       libdrgnimpl_la-arch_ppc64.lo
    CC       libdrgnimpl_la-arch_s390.lo
  ../../libdrgn/arch_s390.c: In function ‘unwind_check_pt_regs’:
  ../../libdrgn/arch_s390.c:182:27: error: unused variable ‘member’ [-Werror=unused-variable]
    182 |  struct drgn_type_member *member;
        |                           ^~~~~~
  ../../libdrgn/arch_s390.c:181:29: error: unused variable ‘frame_type’ [-Werror=unused-variable]
    181 |  struct drgn_qualified_type frame_type;
        |                             ^~~~~~~~~~
  ../../libdrgn/arch_s390.c:180:11: error: unused variable ‘nextbc’ [-Werror=unused-variable]
    180 |  uint64_t nextbc, frame_size, ptregs_size, *pswa, *pswm;
        |           ^~~~~~
  ../../libdrgn/arch_s390.c:179:62: error: unused variable ‘bc_offset’ [-Werror=unused-variable]
    179 |  uint64_t psw_offset, mask_offset, addr_offset, gprs_offset, bc_offset;
        |                                                              ^~~~~~~~~
  ../../libdrgn/arch_s390.c: In function ‘linux_kernel_get_initial_registers_s390’:
  ../../libdrgn/arch_s390.c:278:29: error: unused variable ‘frame_type’ [-Werror=unused-variable]
    278 |  struct drgn_qualified_type frame_type;
        |                             ^~~~~~~~~~
  cc1: all warnings being treated as errors
  make[1]: *** [Makefile:954: libdrgnimpl_la-arch_s390.lo] Error 1
  make[1]: *** Waiting for unfinished jobs....
  make[1]: Leaving directory '/home/runner/work/drgn/drgn/build/temp.linux-x86_64-cpython-310'
```

The CI builds with `-Werror`.

I'm giving this a proper review now, thanks for your patience!

svens-s390:
Thanks, i'll fix those warnings. I missed them in the output. Did you had a chance to review the patch?

libdrgn/arch_s390.c
```diff
@@ -0,0 +1,327 @@
...
+		break;
+	default:
+		err = DRGN_UNKNOWN_RELOCATION_TYPE(r_type);
+		break;
+	}
+	return err;
+}
+
+static struct drgn_error *
+unwind_get_pt_regs_sizes(struct drgn_program *prog,
```
osandov, line 131: Ideally, we should be able to avoid looking up types in `fallback_unwind` for a couple of reasons:

1. The fallback path is intended for cases where we couldn't find debugging information, so we may not be able to look them up at all.
2. The fallback path is also used for userspace targets, which may not have the same type definitions.

`struct pt_regs` is ABI (or at least the `user_pt_regs` part that you're using here), so rather than looking up the sizes and offsets, it's simpler to hard-code them. This is what all of the other architectures that drgn supports do; see `get_initial_registers_from_struct_aarch64()` for an example.

libdrgn/arch_s390.c
```diff
@@ -0,0 +1,327 @@
...
+        uint64_t *gprs = (uint64_t *)(ptregs + gprs_offset / 8);
+	drgn_register_state_set_range_from_buffer(regs2, r0, r15, gprs);
+	drgn_register_state_set_from_buffer(regs2, pswa, pswa);
+	drgn_register_state_set_from_buffer(regs2, pswm, pswm);
+	drgn_register_state_set_pc_from_register(prog, regs2, pswa);
+	drgn_register_state_set_from_buffer(regs2, bc, &gprs[15]);
+
+	*ret = regs2;
+
+	if (*pswm & (1UL << 48)) {
```
osandov, line 219: I actually prefer that we include the userspace frame; it can be useful to know what the original syscall arguments were. The fallback unwind callback should check whether it gets a fault trying to read memory and stop unwinding.

libdrgn/arch_s390.c
```diff
@@ -0,0 +1,327 @@
...
+
+	err = drgn_program_find_type(prog, "struct stack_frame", NULL, &frame_type);
+	if (err)
+		return err;
+
+	err = drgn_type_find_member(frame_type.type, "gprs", &member, &bit_offset);
+	if (err)
+		return err;
+
+	drgn_register_state_set_range_from_buffer(*ret, r6, r15, frame + bit_offset / 8);
```
osandov, line 256: From my limited understanding based on reading the ABI documentation and looking at some disassembly, the only registers that are guaranteed to be saved in the frame are `r14` and `r15`; the rest may or may not be saved (other than for the initial frame saved in `__switch_to()`, which always saves `r6`-`r15`). Is that correct?

If so, this could lead us to report bogus register values to the user. Let's be conservative and only recover `r14` and `r15`.

libdrgn/arch_s390.c
```diff
@@ -0,0 +1,327 @@
...
+	if (err)
+		goto out;
+
+	regs = drgn_register_state_create(bc, false);
+	if (!regs)
+		return &drgn_enomem;
+
+	drgn_register_state_set_from_u64(prog, regs, bc, ksp);
+
+	regs2 = drgn_register_state_create(bc, false);
```
osandov, line 303: You can probably avoid creating a fake frame by having this function and `fallback_unwind_s390()` both call another function that takes the backchain as a `uint64_t`. However, I suspect that given my point above about only recovering `r14` and `r15` for the fallback case, I suspect that these two versions will be different enough that you won't need to share anything anyways.

libdrgn/arch_s390_defs.py
```diff
@@ -0,0 +1,16 @@
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+REGISTERS = [
+    *[DrgnRegister(f"r{i}") for i in range(0,16)],
```
osandov, line 4: This is equivalent to the preferred form without the start index:

```suggestion
    *[DrgnRegister(f"r{i}") for i in range(16)],
```

libdrgn/arch_s390_defs.py
```diff
@@ -0,0 +1,16 @@
...
+    DrgnRegister("pswm"),
+    DrgnRegister("pswa"),
+    DrgnRegister("bc"),
+]
+
+REGISTER_LAYOUT = [
+    # Caller-saved registers.
+    *[DrgnRegisterLayout(f"r{i}", size=8, dwarf_number=i) for i in range(0,16)],
```
osandov, line 12: It might be worth it to put the callee-saved registers before the caller-saved registers since the latter are only recovered for the interrupt/syscall case and are thus unused for the vast majority of cases:

```suggestion
    # Callee-saved registers and return address (r14).
    *[DrgnRegisterLayout(f"r{i}", size=8, dwarf_number=i) for i in range(6, 16)],
    # Caller-saved registers.
    *[DrgnRegisterLayout(f"r{i}", size=8, dwarf_number=i) for i in range(6)],
```

`libdrgn/build-aux/gen_arch_inc_strswitch.py` has some documentation about this.

libdrgn/arch_s390_defs.py
```diff
@@ -0,0 +1,16 @@
...
+    DrgnRegisterLayout("pswm", size=8, dwarf_number=64),
+    DrgnRegisterLayout("pswa", size=8, dwarf_number=65),
+    DrgnRegisterLayout("bc", size=8, dwarf_number=127),
```
osandov, line 15: I don't see any mentions of DWARF register number 127, so I suspect you did this because it's a "fake" register. You can use `dwarf_number=None` to indicate that it's not used by DWARF.

However, I'm not sure whether we need this at all. Can't we get the backchain to the next frame by reading from `r15 + 152` at the current frame?

libdrgn/arch_s390_defs.py
```diff
@@ -0,0 +1,16 @@
...
+]
```
osandov, line 16: Since you submitted this, I added the stack pointer register to the register definition format:

```suggestion
]

STACK_POINTER_REGISTER = "r15"
```

_drgn.pyi
```diff
@@ -954,6 +954,9 @@ class Architecture(enum.Enum):
     RISCV32 = ...
     """The 32-bit RISC-V architecture."""
 
+    S390 = ...
+    """The s390 architecture, a.k.a. IBM Z."""
```
osandov, line 958: Thanks for the explanation of s390 vs s390x. drgn does actually have some support for userspace debugging, so I think we need definitions for both 31-bit s390 and s390x.

libdrgn/arch_s390.c
```diff
@@ -0,0 +1,327 @@
...
+	DRGN_CFI_SAME_VALUE_INIT(DRGN_REGISTER_NUMBER(r15)),
+	);
+
+static struct drgn_error drgn_invalid_rel = {
+	.code = DRGN_ERROR_OTHER,
+	.message = "invalid relocation entry",
+};
+
+static struct drgn_error *
+apply_rela_bits(char *dest, uint64_t loc, uint64_t val,
```
osandov, line 35: This seems to make assumptions that the byte order and word size match that of s390x, which may not be true. Please use the existing relocation helper functions that handle byte order and unaligned accesses. We don't need to handle all of these relocation forms, since only a few tend to be used for debug info.

osandov (changes requested):
So sorry for the delay. Between some vacation, holidays, and me needing to learn enough about s390x, this took me much longer to get to than I would've liked. I've left some comments inline.

Additionally, to make up for my delay (and to give myself a better understanding of s390x), I pushed a branch addressing most of these comments: https://github.com/osandov/drgn/commits/s390. Please take a look and let me know what you think, and feel free to base future revisions on that. Keep in mind that you can force push to your branch to update this PR rather than creating a new PR. The main things still missing are:

1. What do we do about `-mpacked-stack` vs `-mno-packed-stack`? Maybe we can assume that userspace uses the default no-packed format and the kernel uses packed? It appears that it isn't necessarily true, but is it usually true in practice?
2. We still need `pt_regs_get_initial_registers()`. That should be straightforward to implement by refactoring the code that we already have for getting the registers from `struct pt_regs` in the interrupt case.
3. We still need `prstatus_get_initial_registers()`. It looks like s390x uses `s390_regs` for that, which is different from `pt_regs` (unlike other architectures, where `PRSTATUS` typically uses the same format as `pt_regs`).

This has also made it clear to me that the architecture support callbacks need much better documentation, so I will add that in parallel.

svens-s390:
> So sorry for the delay. Between some vacation, holidays, and me needing to learn enough about s390x, this took me much longer to get to than I would've liked. I've left some comments inline.

That's fine, we all have a life besides Software :-)
 
> Additionally, to make up for my delay (and to give myself a better understanding of s390x), I pushed a branch addressing most of these comments: https://github.com/osandov/drgn/commits/s390. Please take a look and let me know what you think, and feel free to base future revisions on that. Keep in mind that you can force push to your branch to update this PR rather than creating a new PR. The main things still missing are:
> 
>     1. What do we do about `-mpacked-stack` vs `-mno-packed-stack`? Maybe we can assume that userspace uses the default no-packed format and the kernel uses packed? It appears that it isn't necessarily true, but is it usually true in practice?

That wouldn't work, as the long term goal is to switch userspace also to -mpacked-stack. Stackframes would still look different because the kernel is compiled with -mpacked-stack -msoft-float -mbackchain. However, dwarf should handle this(?)
I must admit that i didn't thought about userspace stack unwinding...

>     2. We still need `pt_regs_get_initial_registers()`. That should be straightforward to implement by refactoring the code that we already have for getting the registers from `struct pt_regs` in the interrupt case.

Ok.

>     3. We still need `prstatus_get_initial_registers()`. It looks like s390x uses `s390_regs` for that, which is different from `pt_regs` (unlike other architectures, where `PRSTATUS` typically uses the same format as `pt_regs`).

Ok.

> 
> This has also made it clear to me that the architecture support callbacks need much better documentation, so I will add that in parallel.

Thanks!

libdrgn/arch_s390.c
svens-s390: Ok, i tried to make the code smart, but with your explanation lets use hardcoded offsets.

_drgn.pyi
svens-s390: We have no plans to support 31-bit mode - s390/IBM Z has support for 64 bit since 2000. We dropped 31-bit support from the Linux kernel for quite some time now, the only thing we still have is CONFIG_COMPAT to let customers run old 31-bit binaries in userspace. But i'm not aware of anyone doing development on 31-bit. So 31-bit support would add a lot more complexity to the code with no addtional gain.

osandov: I see. I was aware that the kernel doesn't support 31 bit s390, but I didn't realize it was that uncommon for userspace. I agree that it's not worth supporting it. However, it's not too much effort to at least detect 31-bit s390 and have a stub architecture definition for it -- that's what I did in my branch.

osandov:
> That wouldn't work, as the long term goal is to switch userspace also to -mpacked-stack.

That does make things trickier.

> Stackframes would still look different because the kernel is compiled with -mpacked-stack -msoft-float -mbackchain. However, dwarf should handle this(?)

If the common case for userspace is no backchains, then `fallback_unwind` isn't very useful for it anyways, and we can make it always fail (i.e., return `&drgn_stop`) for userspace, relying solely on DWARF/`.eh_frame`.

P.S. I added a bunch of documentation for `drgn_architecture_info` in 174b797ae320fa927cd10d8562b7969c5c61b920, so that should help with the remaining callbacks.

svens-s390:
>     2. We still need `pt_regs_get_initial_registers()`. That should be straightforward to implement by refactoring the code that we already have for getting the registers from `struct pt_regs` in the interrupt case.

What is a good testcase for pt_regs_get_initial_registers()? From reading the source, i thought it's used when unwinding from an IRQ handler. I added some delay to IRQ handling to catch it, but that made drgn responsd with "cannot unwind stack of running task"

osandov (changes requested):
Thanks for the followup!

> What is a good testcase for pt_regs_get_initial_registers()? From reading the source, i thought it's used when unwinding from an IRQ handler. I added some delay to IRQ handling to catch it, but that made drgn responsd with "cannot unwind stack of running task"

I didn't have a good test case until you asked, but I just added https://github.com/osandov/drgn/commit/7b64aee4ddfb208f13634ee3d58c76dc5f971b53. It requires building an external kernel module. The CI does this automatically, but unfortunately I don't have that set up to test architectures other than x86-64 yet. Here's how I did it manually in a Debian VM:

```console
root@debian-s390x:~/drgn# apt install linux-headers-$(uname -r)
...
root@debian-s390x:~/drgn# make -C /lib/modules/$(uname -r)/build M="$PWD/tests/linux_kernel/kmod"
make: Entering directory '/usr/src/linux-headers-5.10.0-18-s390x'
  CC [M]  /root/drgn/tests/linux_kernel/kmod/drgn_test.o
  MODPOST /root/drgn/tests/linux_kernel/kmod/Module.symvers
  CC [M]  /root/drgn/tests/linux_kernel/kmod/drgn_test.mod.o
  LD [M]  /root/drgn/tests/linux_kernel/kmod/drgn_test.ko
make: Leaving directory '/usr/src/linux-headers-5.10.0-18-s390x'
root@debian-s390x:~/drgn# export DRGN_TEST_KMOD="$PWD/tests/linux_kernel/kmod/drgn_test.ko"
root@debian-s390x:~/drgn# insmod "$DRGN_TEST_KMOD"
root@debian-s390x:~/drgn# python3 -Bm unittest tests.linux_kernel.test_stack_trace -v
test_by_pid_dwarf (tests.linux_kernel.test_stack_trace.TestStackTrace) ... ok
test_by_pid_orc (tests.linux_kernel.test_stack_trace.TestStackTrace) ... skipped 's390x does not use ORC'
test_by_pt_regs (tests.linux_kernel.test_stack_trace.TestStackTrace) ... FAIL
test_by_task_struct (tests.linux_kernel.test_stack_trace.TestStackTrace) ... ok
test_local_variable (tests.linux_kernel.test_stack_trace.TestStackTrace) ... ok
test_locals (tests.linux_kernel.test_stack_trace.TestStackTrace) ... ok
test_prog (tests.linux_kernel.test_stack_trace.TestStackTrace) ... ok
test_registers (tests.linux_kernel.test_stack_trace.TestStackTrace) ... ok
test_sp (tests.linux_kernel.test_stack_trace.TestStackTrace) ... ok
test_stack__repr_pretty_ (tests.linux_kernel.test_stack_trace.TestStackTrace) ... ok

FAIL: test_by_pt_regs (tests.linux_kernel.test_stack_trace.TestStackTrace)
  File "/root/drgn/tests/linux_kernel/test_stack_trace.py", line 63, in test_by_pt_regs
    self._test_drgn_test_kthread_trace(self.prog.stack_trace(pt_regs))
  File "/root/drgn/tests/linux_kernel/test_stack_trace.py", line 25, in _test_drgn_test_kthread_trace
    self.fail("Couldn't find drgn_test_kthread_fn3 frame")
AssertionError: Couldn't find drgn_test_kthread_fn3 frame

Ran 10 tests in 32.072s

FAILED (failures=1, skipped=1)
```

As you can see, it looks like stack tracing by `task_struct` or PID works (yay!), but the `pt_regs` case doesn't yet.

libdrgn/arch_s390x.c
```diff
@@ -190,6 +190,66 @@ linux_kernel_get_initial_registers_s390x(const struct drgn_object *task_obj,
...
+	drgn_register_state_set_pc_from_register(prog, regs, pswa);
+	*ret = regs;
+	return NULL;
+}
+
+static struct drgn_error *
+pt_regs_get_initial_registers_s390x(const struct drgn_object *obj,
+				    struct drgn_register_state **ret)
+{
+	return get_initial_registers_from_struct_s390x(drgn_object_program(obj),
```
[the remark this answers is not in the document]
svens-s390: Indeed, i mixed up the structures. I changed that in my code, but it looks like i get an all zero buffer in pt_regs_get_initial_registers_s390x(). Haven't looked into the code why.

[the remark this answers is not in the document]
svens-s390: just noted that the kernel module uses crash_setup_regs(), but s390 doesn't implement that.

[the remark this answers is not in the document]
svens-s390: `Finished processing dependencies for drgn==0.0.21+69.gc1aa7d8fcd40
test_by_pid_dwarf (tests.linux_kernel.test_stack_trace.TestStackTrace) ... ok
test_by_pid_orc (tests.linux_kernel.test_stack_trace.TestStackTrace) ... skipped 's390x does not use ORC'
test_by_pt_regs (tests.linux_kernel.test_stack_trace.TestStackTrace) ... ok
test_by_task_struct (tests.linux_kernel.test_stack_trace.TestStackTrace) ... ok
test_local_variable (tests.linux_kernel.test_stack_trace.TestStackTrace) ... ok
test_locals (tests.linux_kernel.test_stack_trace.TestStackTrace) ... ok
test_prog (tests.linux_kernel.test_stack_trace.TestStackTrace) ... ok
test_registers (tests.linux_kernel.test_stack_trace.TestStackTrace) ... ok
test_sp (tests.linux_kernel.test_stack_trace.TestStackTrace) ... ok
test_stack__repr_pretty_ (tests.linux_kernel.test_stack_trace.TestStackTrace) ... ok

Ran 10 tests in 0.682s

OK (skipped=1)`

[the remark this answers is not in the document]
osandov: Awesome! Would it be possible to put your implementation of `crash_setup_regs()` in the test module itself, so that it will work without your patch? Something like:

```diff
diff --git a/tests/linux_kernel/kmod/drgn_test.c b/tests/linux_kernel/kmod/drgn_test.c
index 681de009..2db79b00 100644
--- a/tests/linux_kernel/kmod/drgn_test.c
+++ b/tests/linux_kernel/kmod/drgn_test.c
@@ -404,6 +404,20 @@ static struct task_struct *drgn_test_kthread;
 static DECLARE_COMPLETION(drgn_test_kthread_ready);
 struct pt_regs drgn_test_kthread_pt_regs;
 
+#if __s390x__
+// s390x doesn't implement crash_setup_regs().
+static inline void drgn_test_get_pt_regs(struct pt_regs *regs)
+{
+	// YOUR IMPLEMENTATION HERE
+}
+#else
+// crash_setup_regs() is an internal function for kexec, but it does exactly
+// what we need: saves the current registers in a struct pt_regs. If it goes
+// away or doesn't work on other architectures, we might need to do it ourselves
+// like we do for s390x.
+#define drgn_test_get_pt_regs(regs) crash_setup_regs(regs, NULL)
+#endif
+
  __attribute__((__optimize__("O0")))
 static void drgn_test_kthread_fn3(void)
 {
@@ -423,11 +437,7 @@ static void drgn_test_kthread_fn3(void)
 		}
 		if (kthread_should_park()) {
 			__set_current_state(TASK_RUNNING);
-
-			// This is an internal function for kexec, but I
-			// couldn't find a better way to get current pt_regs.
-			crash_setup_regs(&drgn_test_kthread_pt_regs, NULL);
-
+			drgn_test_get_pt_regs(&drgn_test_kthread_pt_regs);
 			kthread_parkme();
 			continue;
 		}
```

(I might end up doing that for every architecture anyways instead of relying on an implementation detail.)

Other than that, I think the only unresolved issue is what to do about the fallback unwinder in userspace. What do you think about my earlier suggestion:

> If the common case for userspace is no backchains, then fallback_unwind isn't very useful for it anyways, and we can make it always fail (i.e., return `&drgn_stop`) for userspace, relying solely on DWARF/`.eh_frame`.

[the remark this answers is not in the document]
svens-s390: I copied your suggestion drgn_test_get_pt_regs(), test are fine with that on s390. For the 'userspace with no backchain' case i moved

```
f (!(prog->flags & DRGN_PROGRAM_IS_LINUX_KERNEL))
                return &drgn_stop;
```

to the beginning of fallback_unwind_s390x(), so it will never try to unwind userspace with the help of backchains.

osandov (approved):
Great, tests pass on my end as well. I made some minor style changes, but other than that, this is ready to go.

Seeing a stack trace on a new architecture is magical every time, thank you again! Do you have any plans for adding virtual address translation support?

svens-s390:
Awesome, thanks for merging this!

Regarding translation support: I'm currently working on that. It's already translating va's.

One thing i would need is some callback/extension to get the pagetable address from mm.context.asce instead of mm.pgd.

Reason for this is that the register where the page table address is written to (Control register 1/7/13), also contains the type of page table that is used and some other stuff. Therefore s390 writes the address of the page table to mm.pgd, but
records the register value in mm.context.asce.

PR #215, opened 2022-10-07
helpers: slab: Add get_slab_aliases()

brenns10:
When SLUB is in use, and the CONFIG_SYSFS is enabled (a very common situation), we are able to identify which slab caches have been merged. Provide a helper to expose this information so that users can lookup the correct cache name, or identify all other caches which have been merged with a given cache.

### DIFF
--- a/drgn/helpers/linux/slab.py
+++ b/drgn/helpers/linux/slab.py
@@ -16,9 +16,20 @@ Linux slab allocator.
 """
 
 import operator
-from typing import Iterator, Optional, Set, Union, overload
+from os import fsdecode
+from typing import Dict, Iterator, Optional, Set, Union, overload
 
-from drgn import NULL, FaultError, IntegerLike, Object, Program, Type, cast, sizeof
+from drgn import (
+    NULL,
+    FaultError,
+    IntegerLike,
+    Object,
+    Program,
+    Type,
+    cast,
+    container_of,
+    sizeof,
+)
 from drgn.helpers.common.format import escape_ascii_string
 from drgn.helpers.linux.cpumask import for_each_online_cpu
 from drgn.helpers.linux.list import list_for_each_entry
@@ -30,11 +41,13 @@ from drgn.helpers.linux.mm import (
     virt_to_page,
 )
 from drgn.helpers.linux.percpu import per_cpu_ptr
+from drgn.helpers.linux.rbtree import rbtree_inorder_for_each_entry
 
 __all__ = (
     "find_containing_slab_cache",
     "find_slab_cache",
     "for_each_slab_cache",
+    "get_slab_cache_aliases",
     "print_slab_caches",
     "slab_cache_for_each_allocated_object",
     "slab_cache_is_merged",
@@ -90,6 +103,68 @@ def slab_cache_is_merged(slab_cache: Object) -> bool:
     :param slab_cache: ``struct kmem_cache *``
     """
     return slab_cache.refcount > 1
+
+
+def get_slab_cache_aliases(prog: Program) -> Dict[str, str]:
+    """
+    Return a dict mapping slab cache name to the cache it was merged with.
+
+    The SLAB and SLUB subsystems can merge caches with similar settings and
+    object sizes, as described in the documentation of
+    :func:`slab_cache_is_merged()`. In some cases, the information about which
+    caches were merged is lost, but in other cases, we can reconstruct the info.
+    This function reconstructs the mapping, but requires that the kernel is
+    configured with ``CONFIG_SLUB`` and ``CONFIG_SYSFS``.
+
+    The returned dict maps from original cache name, to merged cache name. You
+    can use this mapping to discover the correct cache to lookup via
+    :func:`find_slab_cache()`. The dict contains an entry only for caches which
+    were merged into a cache of a different name.
+
+    >>> cache_to_merged = get_slab_cache_aliases(prog)
+    >>> cache_to_merged["dnotify_struct"]
+    'avc_xperms_data'
+    >>> "avc_xperms_data" in cache_to_merged
+    False
+    >>> find_slab_cache(prog, "dnotify_struct") is None
+    True
+    >>> find_slab_cache(prog, "avc_xperms_data") is None
+    False
+
+    :warning: This function will only work on kernels which are built with
+      ``CONFIG_SLUB`` and ``CONFIG_SYSFS`` enabled.
+
+    :param prog: Program to search
+    :returns: Mapping of slab cache name to final merged name
+    :raises LookupError: If the helper fails because the debugged kernel
+      doesn't have the required configuration
+    """
+    try:
+        slab_kset = prog["slab_kset"]
+    except KeyError:
+        raise LookupError(
+            "Couldn't find SLUB sysfs information: get_slab_cache_aliases() "
+            "requires CONFIG_SLUB and CONFIG_SYSFS enabled in the debugged "
+            "kernel."
+        ) from None
+    link_flag = prog.constant("KERNFS_LINK")
+    name_map = {}
+    for child in rbtree_inorder_for_each_entry(
+        "struct kernfs_node",
+        slab_kset.kobj.sd.dir.children.address_of_(),
+        "rb",
+    ):
+        if child.flags & link_flag:
+            cache = container_of(
+                cast("struct kobject *", child.symlink.target_kn.priv),
+                "struct kmem_cache",
+                "kobj",
+            )
+            original_name = fsdecode(child.name.string_())
+            target_name = fsdecode(cache.name.string_())
+            if original_name != target_name:
+                name_map[original_name] = target_name
+    return name_map
 
 
 def for_each_slab_cache(prog: Program) -> Iterator[Object]:
--- a/tests/linux_kernel/helpers/test_slab.py
+++ b/tests/linux_kernel/helpers/test_slab.py
@@ -10,6 +10,7 @@ from drgn.helpers.linux.slab import (
     find_containing_slab_cache,
     find_slab_cache,
     for_each_slab_cache,
+    get_slab_cache_aliases,
     slab_cache_for_each_allocated_object,
     slab_cache_is_merged,
 )
@@ -18,6 +19,8 @@ from tests.linux_kernel import (
     skip_unless_have_full_mm_support,
     skip_unless_have_test_kmod,
 )
+
+SLAB_SYSFS_PATH = Path("/sys/kernel/slab")
 
 
 def get_proc_slabinfo_names():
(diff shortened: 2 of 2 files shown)


brenns10:
(already ran the vmtest locally so hopefully we're all good with CI!)

drgn/helpers/linux/slab.py
```diff
@@ -321,3 +333,66 @@ def find_containing_slab_cache(  # type: ignore  # Need positional-only argument
...
+    discover whether a given slab cache has been merged, and if so, which slab
+    cache it was merged with.
+
+    This function returns a mapping from original cache name, to merged cache
+    name. You can use this mapping to discover the correct cache to lookup via
+    :func:`find_slab_cache()`. If a slab cache name does not appear as a key in
+    this mapping, then it has not been merged with any other cache, and can
+    directly be looked up. A cache name may be mapped to itself; in this case,
+    the cache was merged with others, and the final merge cache used that
+    particular name.
```
osandov, line 355: I'm pondering whether we should include these names that map to themselves. It seems like their presence might cause confusion. You can already tell that a slab cache was "merged into" by its presence in `.values()`.

The only exception is caches that _only_ show up as a mapping to themselves. Do you know how that happens?

drgn/helpers/linux/slab.py
```diff
@@ -321,3 +333,66 @@ def find_containing_slab_cache(  # type: ignore  # Need positional-only argument
...
+    :returns: Mapping of slab cache name to final merged name
+    """
+    children = rbtree_inorder_for_each_entry(
+        "struct kernfs_node",
+        prog["slab_kset"].kobj.sd.dir.children.address_of_(),
+        "rb",
+    )
+    link_flag = prog.constant("KERNFS_LINK")
+    cache_alias_map = {}
+    alias_true_name_map = {}
```
osandov, line 380: Do we need these two layers of maps? If we only look at symlinks and skip over everything else, we can get the "true" name directly from `target_kn`. Something like:

```python
    link_flag = prog.constant("KERNFS_LINK")
    name_map = {}
    for child in rbtree_inorder_for_each_entry(
        "struct kernfs_node",
        prog["slab_kset"].kobj.sd.dir.children.address_of_(),
        "rb",
    ):
        if child.flags & link_flag:
            cache = container_of(
                cast("struct kobject *", child.symlink.target_kn.priv),
                "struct kmem_cache",
                "kobj",
            )
            name_map[child.name.string_().decode()] = cache.name.string_().decode()
    return name_map
```

drgn/helpers/linux/slab.py
```diff
@@ -321,3 +333,66 @@ def find_containing_slab_cache(  # type: ignore  # Need positional-only argument
...
+    for child in children:
+        if child.flags & link_flag:
+            kernfs_target = child.symlink.target_kn
+            cache_alias_map[
+                child.name.string_().decode()
```
osandov, line 385: Let's also use [`os.fsdecode()`](https://docs.python.org/3/library/os.html#os.fsdecode) instead of `bytes.decode()` for everything we decode here: these are technically filenames, and `os.fsdecode()` usually uses the `surrogateescape` error handler which is robust against invalid input.

osandov (changes requested):
Thanks, this is super useful! I commented on a few potential improvements and a question on whether we should include names that map to themselves.

drgn/helpers/linux/slab.py
brenns10: Yes, even in the vmtest kernels, the returned mapping contains several caches that only show up as a mapping to themselves. It's easier to see in the reverse mapping (final cache -> list of names merged with it). Here's a sample:
```
'kmalloc-8': ['kmalloc-8'],
'kmalloc-8192': ['kmalloc-8192'],
'kmalloc-96': ['blkdev_ioc', 'kmalloc-96'],
'mnt_cache': ['mnt_cache', 'ip6_dst_cache'],
'net_namespace': ['net_namespace'],
'numa_policy': ['numa_policy'],
'pid_namespace': ['pid_namespace'],
'posix_timers_cache': ['posix_timers_cache'],
'shared_policy_node': ['shared_policy_node', 'fasync_cache', 'nsproxy', 'ip_fib_trie'],
'signal_cache': ['RAW', 'PING', 'signal_cache', 'mm_struct'],
```

It's easy enough to see this on a running kernel with sysfs too:
```
ls -lh /sys/kernel/slab | grep -- '->' | awk '{print $NF}' | sort | uniq -c | sort -nr
```
This will show tons of slab IDs which are symlinked to by just one name. I believe all of those slabs are "mergeable", but no matching caches were actually created, so they're left with just the one name. The caches which exist just as a name, with no link, are those which are "unmergeable" due to some setting -- see [sysfs_slab_add()](https://elixir.bootlin.com/linux/v6.0.1/source/mm/slub.c#L5949).

So I think what I'll do with this is call that out in the docstring. I guess the other option would be to filter this mapping so that it only contains caches which were _actually merged_, but I don't think it's worth having the extra logic just to remove some info. Let me know if you disagree.

drgn/helpers/linux/slab.py
brenns10: Ha, thanks for rewriting my first attempt code into something far simpler!

brenns10:
Alright, this should be ready for a second look. I took your suggestions (thanks for cleaning up my crappy code, I wrote it while debugging without much thought about making the code _good_). And I did get rid of the names mapping to themselves - documenting them would have been more trouble than it's worth.

tests/linux_kernel/helpers/test_slab.py
```diff
@@ -75,6 +76,31 @@ def test_slab_cache_is_merged_true(self):
...
 
+    def test_slab_cache_aliases(self):
+        alias_to_name = get_slab_cache_aliases(self.prog)
+        for aliases in self._slab_cache_aliases().values():
+            # Some entries in sysfs are symlinks, and some are not.
+            # Ignore alias groups which are of length 1 -- either they
+            # refer to symlinks but there's only one cache mapping to them, in
+            # which case it doesn't matter if they appear in the mapping, or
+            # they are not symlinks, and won't show up in the mapping anyway.
+            if len(aliases) <= 1:
```
osandov, line 87: Now that you removed entries with that only alias to themselves, maybe we should test that explicitly? E.g.,

```python3
if len(aliases) == 1:
    self.assertNotIn(aliases[0], alias_to_name)
else:
    ... your original test ...
```

tests/linux_kernel/helpers/test_slab.py
```diff
@@ -75,6 +76,31 @@ def test_slab_cache_is_merged_true(self):
             self.fail("couldn't find slab cache")
         self.assertTrue(slab_cache_is_merged(slab_cache))
 
+    def test_slab_cache_aliases(self):
+        alias_to_name = get_slab_cache_aliases(self.prog)
```
osandov, line 80: This will blow up on kernels with SLAB or SLOB. We have kernels with SLAB and SLOB built, we just don't run them in the CI because it'd take forever. You can test both manually with:

```
python3 setup.py test -k '*alternative,*tiny'
```

If `/sys/kernel/slab` doesn't exist, we should check that `get_slab_cache_aliases()` raises your exception (e.g., with `self.assertRaisesRegex()`).

drgn/helpers/linux/slab.py
```diff
@@ -92,6 +105,66 @@ def slab_cache_is_merged(slab_cache: Object) -> bool:
...
+    :warning: This function will only work on kernels which are built with
+      ``CONFIG_SLUB`` and ``CONFIG_SYSFS`` enabled.
+
+    :param prog: Program to search
+    :returns: Mapping of slab cache name to final merged name
+    """
+    try:
+        slab_kset = prog["slab_kset"]
+    except KeyError:
+        raise ValueError(
```
osandov, line 143: The `get_kconfig()` helper uses `LookupError` for a similar case: https://github.com/osandov/drgn/blob/72f1db796844c36ef7cf542166ad44336881e19e/drgn/helpers/linux/kconfig.py#L49-L51

Do you think that would make more sense here, too?

Either way, this should be `raise foo from None` so that we don't get extra noise from the original `KeyError`; I just fixed this for `get_kconfig()`, too: https://github.com/osandov/drgn/commit/72f1db796844c36ef7cf542166ad44336881e19e.

osandov (changes requested):
This looks great overall. I had a few final nits (one of which I missed in the previous iteration).

tests/linux_kernel/helpers/test_slab.py
```diff
@@ -75,6 +76,31 @@ def test_slab_cache_is_merged_true(self):
             self.fail("couldn't find slab cache")
         self.assertTrue(slab_cache_is_merged(slab_cache))
 
+    def test_slab_cache_aliases(self):
```
osandov, line 79: Oops, one more, let's give this the same name as the helper it's testing:

```suggestion
    def test_get_slab_cache_aliases(self):
```

drgn/helpers/linux/slab.py
brenns10: LookupError is fine by me - I'll note it in the documentation too, since the whole point here seems to be that we want users to be able to know _which_ exception to catch if their kernel doesn't support the helper.

tests/linux_kernel/helpers/test_slab.py
brenns10: I also forgot to assert that, for len(aliases) >= 1, the target alias should not be present as a key in the dict. Added that in my latest revision too.

brenns10:
Should be ready for another look! I ended up using `assertRaises()` rather than `assertRaisesRegex()`... I'm not a fan of asserting about the content of a string literal, but I can definitely switch it if you want.

tests/linux_kernel/helpers/test_slab.py
```diff
@@ -75,6 +77,37 @@ def test_slab_cache_is_merged_true(self):
             self.fail("couldn't find slab cache")
         self.assertTrue(slab_cache_is_merged(slab_cache))
 
+    def test_get_slab_cache_aliases(self):
+        if not SLAB_SYSFS_PATH.exists():
+            # A SLOB or SLAB kernel, or one without SYSFS. Test that the
+            # helper fails as expected.
+            self.assertRaises(LookupError, get_slab_cache_aliases, self.prog)
```
osandov, line 84: I think it's valuable to sanity check that we caught the right error and not some random unrelated error, so how about we at least do something like:

```suggestion
            self.assertRaisesRegex(LookupError, "CONFIG_SYSFS", get_slab_cache_aliases, self.prog)
```

osandov (commented):
I think we can do `assertRaisesRegex` without making the test too fragile. Other than that, this is good to go!

brenns10:
You got it!

PR #218, opened 2022-10-27
helper: partially follow link in path_lookup

Wengang-oracle:
Current path_lookup() doesn't follow symbol links. The symbol link following is FS specific, it's hard and not neccesary to support all file systems. This patch supports:
1) EXT4 simple symbol link (target path stored in inode.i_link). 
2) XFS inlined symbol link.

Signed-off-by: Wengang Wang <wen.gang.wang@oracle.com>
Reviewed-by: Stephen Brennan <stephen.s.brennan@oracle.com>

### DIFF
--- a/drgn/helpers/linux/fs.py
+++ b/drgn/helpers/linux/fs.py
@@ -70,6 +70,37 @@ def _follow_dotdot(
         mnt = mnt_parent
     return _follow_mount(mnt, dentry)
 
+S_IFMT      = 0o170000
+S_IFLNK     = 0o120000
+def is_inode_symlink(
+    inode: Object) -> bool:
+    return (inode.i_mode & S_IFMT) == S_IFLNK
+
+XFS_IFINLINE            = 0x01
+XFS_DINODE_FMT_LOCAL    = 0x01
+def xfs_get_link(
+    inode: Object) -> bytes:
+    xfs_inode = container_of(inode, "struct xfs_inode", "i_vnode")
+    ifork = xfs_inode.i_df
+    # xfs_ifork_t has different members in different versions.
+    if hasattr(ifork, "if_format") and ifork.if_format == XFS_DINODE_FMT_LOCAL:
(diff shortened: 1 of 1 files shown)


drgn/helpers/linux/fs.py
```diff
@@ -120,6 +151,19 @@ def path_lookup(
...
+                s = get_link(dentry.d_inode)
+                if not s:
+                    raise Exception(f"could not get link from inode")
+
+                # full symbol link target path
+                if s[0:1] == b'/':
+                    link_target = s
+                else:
+                    solved_path = b"/" + b"/".join(components[0:i+1])
+                    link_target = solved_path + "/" + s
```
osandov, line 164: This will redo the entire lookup up to the symlink, right? That sounds pretty inefficient (and racy on live systems), especially if there are multiple symlinks in the path traversal.

There are probably multiple ways to avoid this, but the one that comes to mind is to keep a stack of names being resolved and avoid recursion. Something like:

```python3
class PathLookupName:
    def __init__(self, name: str) -> None:
        self.components = name.split(b"/")
        self.index = 0

name_stack = [PathLookupName(os.fsencode(path))]
while name_stack:
    name = name_stack[-1]
    if name.index >= len(name.components):
        name_stack.pop()
        continue

    component = name_stack.components[name.index]
    name.index += 1

    ... handle component ...

    if is_symlink(...):
        link_target = ...
        name_stack.push(PathLookupName(link_target))
        if link_target.startswith(b"/"):
            # Restart at the root.
            mnt = root_mnt
            dentry = root_dentry
```

I _think_ that general idea will work. We also need to be careful about cycles (the kernel handles this by returning `ELOOP` if it encounters too many symlinks.)

This is going to need lots of test cases :smile:

osandov (changes requested):
This is a very cool but very tricky feature. I left a few comments.

Repeating what I mentioned inline, I'd like to have an `inode_readlink()` helper first so we can test that independently, then add this.

PR #219, opened 2022-10-28
Update installation instructions

michel-slm:
- add repology link
- add Debian and Ubuntu instructions

Signed-off-by: Michel Alexandre Salim <michel@michel-slm.name>

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -64,6 +64,10 @@ Package Manager
 
 drgn can be installed using the package manager on some Linux distributions.
 
+.. image:: https://repology.org/badge/vertical-allrepos/drgn.svg
+    :target: https://repology.org/project/drgn/versions
+    :alt: Packaging Status
+
 * Fedora >= 32
 
   .. code-block:: console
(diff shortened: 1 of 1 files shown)


michel-slm:
note: there are two `drgn` in Repology now, which is why the Fedora and openSUSE ones are not showing. These are being merged in https://github.com/repology/repology-rules/pull/648

README.rst
```diff
@@ -83,12 +87,30 @@ drgn can be installed using the package manager on some Linux distributions.
   Install the `drgn <https://aur.archlinux.org/packages/drgn/>`_ package from
   the `AUR <https://wiki.archlinux.org/title/Arch_User_Repository>`_.
 
+* Debian
+
+  Install the `python3-drgn <https://packages.debian.org/search?keywords=python3-drgn>`_
+  package. Note: only on `testing` and `unstable` until `bookworm` is released as `stable`.
```
osandov, line 93: What do you think of condensing this down to:

```suggestion
* Debian >= 12 (Bookworm)

```

Which is similar to what we have for Fedora.

PR #220, opened 2022-11-01
qsort_r -> qsort with TLVs

boryas:
qsort_r is a non-standard glibc extension and turns out to be the only thing that prevents drgn from working on a musl system. "Fix" the use of qsort_r by switching it to qsort with a thread local variable for the parameter.

Tested in a clean chroot install of musl voidlinux.

Signed-off-by: Boris Burkov <boris@bur.io>

### DIFF
--- a/libdrgn/dwarf_info.c
+++ b/libdrgn/dwarf_info.c
@@ -7354,11 +7354,12 @@ static void drgn_debug_info_cache_sh_addr(struct drgn_module *module,
 	}
 }
 
-static int drgn_dwarf_fde_compar(const void *_a, const void *_b, void *arg)
+static _Thread_local struct drgn_dwarf_cie *drgn_dwarf_fde_compar_cies;
+static int drgn_dwarf_fde_compar(const void *_a, const void *_b)
 {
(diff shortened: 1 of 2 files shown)


osandov:
I considered whether instead we should just add a stub implementation of `qsort_r` if we don't detect `qsort_r`. Something like:

```c
static _Thread_local int (*qsort_r_compar)(const void *, const void *, void*);
static _Thread_local void *qsort_r_arg;

static int qsort_r_compar_wrapper(const void *a, const void *b)
{
	return qsort_r_compar(a, b, qsort_r_arg);
}

void qsort_r(void *base, size_t nmemb, size_t size,
	     int (*compar)(const void *, const void *, void*), void *arg)
{
	qsort_r_compar = compar;
	qsort_r_arg = arg;
	qsort(base, nmemb, size, qsort_r_compar_wrapper);
}
```

But that's not actually re-entrant, only thread-safe, which would be misleading. So, let's stick with this and reevaluate if we end up needing to do this in many more places.

I pushed some naming changes. Once tests pass, I'll merge this. Thanks!

PR #222, opened 2022-11-21
libdrgn: dwarf_info: Support DW_TAG_GNU_template_parameter_pack

ajor:
This DWARF tag is used by C++ classes which take a variable number of template parameters, such as std::variant and std::tuple.

Signed-off-by: Alastair Robertson <ajor@meta.com>

### DIFF
--- a/libdrgn/dwarf_info.c
+++ b/libdrgn/dwarf_info.c
@@ -5940,11 +5940,22 @@ drgn_dwarf_template_value_parameter_thunk_fn(struct drgn_object *res,
 }
 
 static struct drgn_error *
-parse_template_parameter(struct drgn_debug_info *dbinfo,
-			 struct drgn_elf_file *file, Dwarf_Die *die,
-			 drgn_object_thunk_fn *thunk_fn,
-			 struct drgn_template_parameters_builder *builder)
+maybe_parse_template_parameter(struct drgn_debug_info *dbinfo,
+			       struct drgn_elf_file *file, Dwarf_Die *die,
+			       struct drgn_template_parameters_builder *builder)
 {
+	drgn_object_thunk_fn *thunk_fn;
+	switch (dwarf_tag(die)) {
+	case DW_TAG_template_type_parameter:
+		thunk_fn = drgn_dwarf_template_type_parameter_thunk_fn;
+		break;
+	case DW_TAG_template_value_parameter:
+		thunk_fn = drgn_dwarf_template_value_parameter_thunk_fn;
+		break;
+	default:
+		return NULL;
+	}
+
 	char tag_buf[DW_TAG_STR_BUF_LEN];
 
 	Dwarf_Attribute attr_mem, *attr;
@@ -5984,6 +5995,27 @@ parse_template_parameter(struct drgn_debug_info *dbinfo,
 	if (err)
 		drgn_lazy_object_deinit(&argument);
 	return err;
+}
+
+static struct drgn_error *
+drgn_parse_template_parameter_pack(struct drgn_debug_info *dbinfo,
+				   struct drgn_elf_file *file, Dwarf_Die *die,
+				   struct drgn_template_parameters_builder *builder)
+{
+	struct drgn_error *err;
+	Dwarf_Die child;
+	int r = dwarf_child(die, &child);
+	while (r == 0) {
+		err = maybe_parse_template_parameter(dbinfo, file, &child, builder);
+		if (err)
+			return err;
+		r = dwarf_siblingof(&child, &child);
+	}
+	if (r == -1) {
+		return drgn_error_create(DRGN_ERROR_OTHER,
+					 "libdw could not parse DIE children");
+	}
+	return NULL;
 }
 
 static struct drgn_error *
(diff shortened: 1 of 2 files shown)


libdrgn/dwarf_info.c
```diff
@@ -6097,6 +6097,17 @@ drgn_compound_type_from_dwarf(struct drgn_debug_info *dbinfo,
 			if (err)
 				goto err;
 			break;
+		case DW_TAG_GNU_template_parameter_pack:
+			r = dwarf_child(&child, &child_param);
+			while (r == 0) {
+				err = parse_template_parameter(dbinfo, module, &child_param,
```
osandov, line 6103: This assumes that the child is `DW_TAG_template_type_parameter`, but it may also be a `DW_TAG_template_value_parameter` (or neither if GCC/Clang decides to put some other information in there in the future).

So, this needs to check the child's tag and handle it just like the other places where we parse `DW_TAG_template_type_parameter`/`DW_TAG_template_value_parameter`. Since this would be the third copy of that logic (in addition to the one just above here and the one in `drgn_function_type_from_dwarf()`), it might be worth refactoring it about to get rid of the duplication, but I'll leave that up to you.

osandov (changes requested):
Thanks! I left one comment inline.

In addition to that, would it be valuable to add a boolean `is_pack` to `struct drgn_type_template_parameter` indicating whether the parameter was in a template parameter pack? That would require plumbing it through the template parameter builder, the structure itself, and the Python bindings, but it wouldn't be too hard.

libdrgn/dwarf_info.c
ajor: Good point, fixed!

ajor:
> In addition to that, would it be valuable to add a boolean `is_pack` to `struct drgn_type_template_parameter` indicating whether the parameter was in a template parameter pack?

It doesn't make a difference for my use-case, but I'd be happy to add that too if you think it'd be useful in general.

tests/test_dwarf.py
```diff
@@ -1680,6 +1680,83 @@ def test_template_value_parameter_missing_value(self):
...
+                            DwarfDie(
+                                DW_TAG.GNU_template_parameter_pack,
+                                (DwarfAttrib(DW_AT.name, DW_FORM.string, "Params"),),
+                                (char_die,),
+                            ),
+                        ),
+                    ),
+                    *labeled_unsigned_int_die,
+                )
+            ).type("TEST").type.template_parameters[0].argument
```
osandov, line 1758: I don't think we should require that all of the children of `DW_TAG_GNU_template_parameter_pack` are either `DW_TAG_template_type_parameter` or `DW_TAG_template_value_parameter`. It's better to be forwards-compatible and allow producers to stick anything else they want in there. (This means that the parsing code needs to handle this differently, too.)

libdrgn/dwarf_info.c
```diff
@@ -6588,15 +6604,8 @@ drgn_function_type_from_dwarf(struct drgn_debug_info *dbinfo,
 			is_variadic = true;
 			break;
 		case DW_TAG_template_type_parameter:
-			err = parse_template_parameter(dbinfo, module, &child,
-						       drgn_dwarf_template_type_parameter_thunk_fn,
-						       &builder.template_builder);
-			if (err)
-				goto err;
-			break;
 		case DW_TAG_template_value_parameter:
```
osandov, line 6607: It just occurred to me, can we have variadic template functions, too? If so, we need to handle that in this function, too.

osandov (changes requested):
> It doesn't make a difference for my use-case, but I'd be happy to add that too if you think it'd be useful in general.

Let's not worry about it. It wouldn't be too hard to add later if someone wants it.

A couple more inline comments.

ajor:
Here's the test program I used for function parameter packs:

```
#include <iostream>

template <typename ...Ts>
__attribute__((noinline)) void foo(Ts... t) {
  std::cout << "hello\n";
}

template <int ...Ns>
__attribute__((noinline)) void baz() {
  std::cout << "hello\n";
}

void bar() {
  foo(1, "yes");
  baz<1,2,3>();
}
```

libdrgn/dwarf_info.c
```diff
@@ -5942,9 +5942,20 @@ drgn_dwarf_template_value_parameter_thunk_fn(struct drgn_object *res,
 static struct drgn_error *
 parse_template_parameter(struct drgn_debug_info *dbinfo,
```
osandov, line 5943: Since this only parses a template parameter if the `dwarf_tag()` matches, can we give it a different name? Something like `maybe_parse_template_parameter()` unless you have a better idea.

libdrgn/dwarf_info.c
```diff
@@ -5942,9 +5942,20 @@ drgn_dwarf_template_value_parameter_thunk_fn(struct drgn_object *res,
...
+	switch (dwarf_tag(die)) {
+		case DW_TAG_template_type_parameter:
+			thunk_fn = drgn_dwarf_template_type_parameter_thunk_fn;
+			break;
+		case DW_TAG_template_value_parameter:
+			thunk_fn = drgn_dwarf_template_value_parameter_thunk_fn;
+			break;
+		default:
+			return NULL;
+	}
```
osandov, line 5957: Style nitpick:

```suggestion
	switch (dwarf_tag(die)) {
	case DW_TAG_template_type_parameter:
		thunk_fn = drgn_dwarf_template_type_parameter_thunk_fn;
		break;
	case DW_TAG_template_value_parameter:
		thunk_fn = drgn_dwarf_template_value_parameter_thunk_fn;
		break;
	default:
		return NULL;
	}
```

libdrgn/dwarf_info.c
```diff
@@ -6563,19 +6577,22 @@ drgn_function_type_from_dwarf(struct drgn_debug_info *dbinfo,
...
+		case DW_TAG_GNU_template_parameter_pack:
+			r = dwarf_child(&child, &child_param);
+			while (r == 0) {
+				err = parse_template_parameter(dbinfo, file, &child_param,
+							       &builder.template_builder);
+				if (err)
+					goto err;
+				r = dwarf_siblingof(&child_param, &child_param);
+			}
+			break;
```
osandov, line 6595: This (and the other copy) are missing error handling for `dwarf_child()`/`dwarf_siblingof()`. It's probably easier to deal with by refactoring this into its own function at this point. Something like:

```c
static struct drgn_error *
drgn_parse_template_parameter_pack(struct drgn_debug_info *dbinfo,
				   struct drgn_elf_file *file, Dwarf_Die *die,
				   struct drgn_template_parameters_builder *builder)
{
	struct drgn_error *err;
	Dwarf_Child child;
	int r = dwarf_child(die, &child);
	while (r == 0) {
		err = parse_template_parameter(dbinfo, file, &child, builder);
		if (err)
			return err;
		r = dwarf_siblingof(&child, &child);
	}
	if (r == -1) {
		return drgn_error_create(DRGN_ERROR_OTHER,
					 "libdw could not parse DIE children");
	}
	return NULL;
}
```

osandov (changes requested):
A few final comments, but this looks almost ready to go.

PR #223, opened 2022-11-23
Correctly interpret template parameters in the C++ lexer

Svetlitski-FB:
Previously, attempting to lookup a type with template parameters via `prog.type()` would result in an "invalid character" error message. This has been remedied by properly handling template parameters in the C++ lexer, enabling type lookups as desired.

### DIFF
--- a/libdrgn/language_c.c
+++ b/libdrgn/language_c.c
@@ -1650,6 +1650,7 @@ enum {
 	C_TOKEN_DOT,
 	C_TOKEN_NUMBER,
 	C_TOKEN_IDENTIFIER,
+	C_TOKEN_TEMPLATE_ARGUMENTS,
 };
 
 #include "c_keywords.inc"
@@ -1696,6 +1697,40 @@ struct drgn_error *drgn_c_family_lexer_func(struct drgn_lexer *lexer,
 		token->kind = C_TOKEN_DOT;
 		p++;
 		break;
+	case '<':
+		// This is a hack for
+		// cpp_append_template_arguments_to_identifier(). We don't want
+		// to deal with actually parsing template arguments, and we
+		// don't care about "<" otherwise, so this scans a token from
+		// the "<" to its matching ">".
+		if (cpp) {
+			token->kind = C_TOKEN_TEMPLATE_ARGUMENTS;
+			p++;
+			size_t less_thans = 1;
+			bool in_single_quotes = false;
+			do {
+				switch (*p++) {
+				case '<':
+					if (!in_single_quotes)
+						less_thans++;
+					break;
+				case '>':
+					if (!in_single_quotes)
+						less_thans--;
+					break;
+				case '\'':
+					// Handling the edge-case of an escaped single-quote
+					if (!(in_single_quotes && *(p - 2) == '\\'))
+						in_single_quotes = !in_single_quotes;
+					break;
+				case '\0':
+					return drgn_error_create(DRGN_ERROR_SYNTAX,
+								 "invalid template arguments");
+				}
+			} while (less_thans > 0);
+			break;
+		}
+		fallthrough;
 	default:
 		if (isalpha(*p) || *p == '_') {
 			do {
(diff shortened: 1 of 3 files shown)


osandov (changes requested):
I don't love the idea of extending the definition of an identifier to include template arguments. For one, it breaks from the standard C/C++ grammar that drgn's parser tries to follow closely. More practically, it would allow for `<...>` garbage to appear where it shouldn't, like in member names.

An alternative approach is to add tokens for "<" and ">" (plus "," and ":" which can appear inside of template arguments) and do some crude "parsing" of template arguments only where they are expected. I took a stab at that here: https://github.com/osandov/drgn/commits/type-template-arguments. Can you test whether that works for your use case?

Svetlitski-FB:
Just checked out that branch, and yes that works perfectly! I can open a separate PR with just the test cases from my changes if you'd like.

Svetlitski-FB:
That version _almost_ works. One issue is that it doesn't properly handle template parameters which are numbers. For example:
```
template <size_t Size>
struct MyArray {
    int x[Size];
};
```

With my version, looking up `prog.type('struct MyArray<32UL>')` works, while the proposed version is unhappy with a `Syntax Error: invalid number`, but I could just extend the number parsing logic to accept suffixes if that's fine with you.

osandov:
Good catch. Yeah, it wouldn't hurt to support suffixes on numbers in general. Here's the grammar from section 6.4.4.1 "Integer constants" in the C standard:

```
integer-suffix:
		unsigned-suffix long-suffix(opt)
		unsigned-suffix long-long-suffix
		long-suffix unsigned-suffix(opt)
		long-long-suffix unsigned-suffix(opt)
unsigned-suffix: one of
		u U
long-suffix: one of
		l L
long-long-suffix: one of
		ll LL
```

I don't know whether C++ is different. I'd be happy to merge support for that in a separate PR.

osandov:
Just saw your edit. Do those ampersands show up in the debugging information? If not, then it's not super beneficial to handle them yet since we need an exact match to the debug info anyways. If they do, then we can either:

1. Add an `&` token (and keep playing whack-a-mole if anything else comes up).
2. Do something similar to your PR where the lexer returns a single token containing the entire template argument list, but separate from the identifier token.

1 is the "purest" solution, 2 is an okay compromise between that and your PR. What do you think?

Svetlitski-FB:
Unfortunately it gets worse:
- C++ allows character literals as template parameters, so we'd have to handle single-quotes
- C++20 allows floating pointer literals as template parameters, and some compilers that I've tried at least (like `g++12`) format the literal in the DWARF using exponent form (e.g. `MyFloat<1.22345228e+1f>` even if the source code doesn't use this form, so we'd have to support this number format.
- C++ allows pointer literals as template parameters. Not only do we have to handle the ampersand as before, but some compilers (i.e. in this case `g++`, but _not_ `clang++`) for some reason decide to surround such parameters with parentheses in the DWARF, so we'd have to handle that.

I realize this probably seems pedantic, but if I've learned anything working on C++ tooling, it's that given a sufficiently large code base, every feature gets used _somewhere_.

osandov:
C++ is amazing. Okay, I agree. Here's my attempt at that, please test it out: https://github.com/osandov/drgn/commits/type-template-arguments2. If it works, it'd be great if you could adapt your test cases for it.

One hilarious corner case it doesn't handle is a template argument that is a `'<'` or `'>'` character literal. I could use a hand fixing that.

Svetlitski-FB:
That appears to work correctly for all of the examples I've tried. I'll address the `'<'` and `'>'` literal edge-case and port the tests.

Svetlitski-FB:
_Technically_ if we want to be super pedantic the approach I took to handling the `'<'` and `'>'` edge-cases doesn't work if you have an integer literal template parameter using single-quotes as a digit separator, but on both `clang++` and `g++` the single-quote digit separators are stripped out and aren't present in the DWARF, so I'm fine with this.

osandov (changes requested):
Thanks for the update and the test cases. I think there's one more case we can handle: when the character template argument itself is a single quote:

```
$ cat test.cpp
template <char C>
class Foo {};
Foo<'\''> foo;
$ g++ -Wall -c -g test.cpp
$ readelf --debug-dump=info test.o | grep Foo
    <1f>   DW_AT_name        : (indirect string, offset: 0x0): Foo<'\''>
```

Svetlitski-FB:
> Thanks for the update and the test cases. I think there's one more case we can handle: when the character template argument itself is a single quote:
> 
> ```
> $ cat test.cpp
> template <char C>
> class Foo {};
> Foo<'\''> foo;
> $ g++ -Wall -c -g test.cpp
> $ readelf --debug-dump=info test.o | grep Foo
>     <1f>   DW_AT_name        : (indirect string, offset: 0x0): Foo<'\''>
> ```

Good catch, addressed this and added a test case for it.

osandov (approved):
I squashed your corner case fixes into the main commit. This is good to go now.

osandov:
Lol, immediately after merging this, another corner case occurred to me which I think worked before your fix for the escaped single quotes:

```
$ cat test.cpp
template <char C>
class Foo {};
Foo<'\\'> foo;
$ g++ -Wall -c -g test.cpp
$ readelf --debug-dump=info test.o | grep Foo   
    <1f>   DW_AT_name        : (indirect string, offset: 0x0): Foo<'\\'>
```

This test case fails:

```diff
diff --git a/tests/test_language_c.py b/tests/test_language_c.py
index d0527a05..2671521d 100644
--- a/tests/test_language_c.py
+++ b/tests/test_language_c.py
@@ -804,6 +804,7 @@ class TestLexer(TestCase):
             ("CharLiteralEdgeCase1", "<'<'>"),
             ("CharLiteralEdgeCase2", "<'>'>"),
             ("CharLiteralEdgeCase3", r"<'\''>"),
+            ("CharLiteralEdgeCase4", r"<'\\'>"),
         )
         self.assertEqual(
             [
```

PR #225, opened 2022-12-01
Add support for looking up types in namespaces

Svetlitski-FB:
Looking up objects in namespaces is already well-supported by `drgn`. These changes bring the same to functionality type lookup, so that `prog.type('struct A::B::C::MyType')` works in an analogous fashion to `prog['A::B::C::MyVar']`.

~I don't have test-cases at present because I was unsure if you'd rather tack on additional assertions to the existing namespace lookup test cases (e.g. [`TestScopes.test_namespaces_nested`](https://github.com/osandov/drgn/blob/845ac5fceb54c222d0b75d860eca8a015336c5a4/tests/test_dwarf.py#L5962-L6007)), or add entirely separate test cases. Just let me know which you prefer and I'll implement them accordingly.~

### DIFF
--- a/libdrgn/dwarf_info.c
+++ b/libdrgn/dwarf_info.c
@@ -6816,6 +6816,45 @@ drgn_type_from_dwarf_internal(struct drgn_debug_info *dbinfo,
 	return NULL;
 }
 
+static struct drgn_error *
+find_enclosing_namespace(struct drgn_namespace_dwarf_index *global_namespace,
+			 const char **name, size_t *name_len,
+			 struct drgn_namespace_dwarf_index **namespace_ret)
+{
+	*namespace_ret = global_namespace;
+	if (*name_len >= 2 && memcmp(*name, "::", 2) == 0) {
+		/* Explicit global namespace. */
+		*name_len -= 2;
+		*name += 2;
+	}
+
+	const char *template_parameters_start = memchr(*name, '<', *name_len);
+	ptrdiff_t searchable_len =
+		template_parameters_start ?
+				  template_parameters_start - *name:
+				  *name_len;
+	const char *colons;
+	while ((colons = memmem(*name, searchable_len, "::", 2))) {
+		struct drgn_dwarf_index_iterator it;
+		uint64_t ns_tag = DW_TAG_namespace;
+		struct drgn_error *err = drgn_dwarf_index_iterator_init(
+			&it, *namespace_ret, *name, colons - *name, &ns_tag, 1);
+		if (err)
+			return err;
+		struct drgn_dwarf_index_die *index_die =
+			drgn_dwarf_index_iterator_next(&it);
+		if (!index_die)
+			return &drgn_not_found;
+		*namespace_ret = index_die->namespace;
+		size_t chars_consumed = colons + 2 - *name;
+		searchable_len -= chars_consumed;
+		*name_len -= chars_consumed;
+		*name = colons + 2;
+	}
+
+	return NULL;
+}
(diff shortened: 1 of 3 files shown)


libdrgn/language_c.c
```diff
@@ -1700,7 +1700,7 @@ struct drgn_error *drgn_c_family_lexer_func(struct drgn_lexer *lexer,
 		if (isalpha(*p) || *p == '_') {
 			do {
 				p++;
-			} while (isalnum(*p) || *p == '_');
+			} while (isalnum(*p) || *p == '_' || (cpp && *p == ':'));
```
Svetlitski-FB, line 1703: Yes, I realize you'll probably hate this, but I did want to put it out there since it's ultimately a very simple way to solve the problem at hand.

osandov: Ha, yeah, I'm still not a fan. But I think we can do something similar to my first suggestion for template arguments (https://github.com/osandov/drgn/commit/a2140d2da639d53daca671d6174e553d74e4f876): add a `C_TOKEN_COLON` token, then modify `cpp_append_template_arguments_to_identifier()` to skip over any sequence of `C_TOKEN_IDENTIFIER` or `C_TOKEN_COLON` (and probably rename `cpp_append_template_arguments_to_identifier()` to something more general like `cpp_append_to_identifier()`).

libdrgn/dwarf_info.c
```diff
@@ -6786,6 +6786,40 @@ drgn_type_from_dwarf_internal(struct drgn_debug_info *dbinfo,
 	return NULL;
 }
 
+static struct drgn_error *
+find_enclosing_namespace(struct drgn_namespace_dwarf_index *global_namespace,
+			 struct drgn_namespace_dwarf_index *starting_namespace,
```
osandov, line 6791: Do you have plans to add cases for where `global_namespace` and `starting_namespace` are different? If not, I think let's just pass the global namespace and assume that they're the same.

osandov (changes requested):

libdrgn/dwarf_info.c
Svetlitski-FB: Not at the moment, was just planning on making this extensible in case we add the ability to search relative to a namespace (like via the scopes API you outlined), but sure we can just cross that bridge when we get there.

Svetlitski-FB:
Just realized this has a bug where it doesn't play well with types which have namespaced template parameters, for example: `prog.type('class MyGeneric<A::B::C>')` does not work as intended. Will fix.

Svetlitski-FB:
Bump on this, when you get a chance to review it.

libdrgn/dwarf_info.c
```diff
@@ -6816,6 +6816,45 @@ drgn_type_from_dwarf_internal(struct drgn_debug_info *dbinfo,
...
+		/* Explicit global namespace. */
+		*name_len -= 2;
+		*name += 2;
+	}
+
+	const char *template_parameters_start = memchr(*name, '<', *name_len);
+	ptrdiff_t searchable_len =
+		template_parameters_start ?
+				  ((uintptr_t)template_parameters_start) - ((uintptr_t)*name) :
+				  *name_len;
```
osandov, line 6835: Why are the casts to `uintptr_t` necessary here? Can't this be:

```suggestion
				  template_parameters_start - *name:
```

osandov (approved):
Thanks! Hopefully we can make this a little less hacky in the future, but this is a good compromise for now.

PR #227, opened 2022-12-02
libdrgn: Ensure booleans are treated as primitives in C++

Svetlitski-FB:
Booleans in C++ programs were not registering as primitives (i.e. `.primitive` was `None` on a boolean type). This has been remedied by treating `bool` as a keyword in C++, in much the same fashion that #158 added support for classes in C++.

### DIFF
--- a/libdrgn/build-aux/gen_c_keywords_inc_strswitch.py
+++ b/libdrgn/build-aux/gen_c_keywords_inc_strswitch.py
@@ -6,6 +6,7 @@ C_KEYWORDS = (
     "_Atomic",
     "_Bool",
     "_Complex",
+    "bool",
     "char",
     "class",
     "const",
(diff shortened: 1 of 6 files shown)


Svetlitski-FB:
The build error from clang in CI is:
```
In file included from ../../libdrgn/language_c.c:1655:
./c_keywords.inc:6:19: error: initializer overrides prior initialization of this subobject [-Werror,-Winitializer-overrides]
[C_TOKEN_BOOL] = "bool",
^~~~~~
./c_keywords.inc:4:19: note: previous initialization is here
[C_TOKEN_BOOL] = "_Bool",
```

This is unfortunate. I would suggest that we just have entirely separate lists of keywords for C and C++ to resolve this, but that won't help address this with regards to C23 support for the same issue (without having to make a new `struct drgn_language drgn_language_c23` that differs only in this respect), since it also makes `bool` a keyword .

Svetlitski-FB:
I could just add a separate `SPECIFIER_CPP_BOOL` and `C_TOKEN_CPP_BOOL` if that's alright with yout.

Svetlitski-FB:
Might revisit this later, but for now this is too minor to be worth pursuing.

PR #229, opened 2022-12-12
Add kernel module helpers

brenns10:
Add the drgn.helpers.linux.module module, which provides a (simple) helper to iterate over the module list, to lookup a module by name, helpers to retrieve kernel modules' layout, and finally, a quite useful helper which can lookup a kernel module by a memory address associated with it.

---

I'm guessing you won't be a big fan of the namedtuple class, but I thought it was the best way to return module layout information given that not all kernels have a `struct module_layout`.

### DIFF
--- a/drgn/helpers/linux/module.py
+++ b/drgn/helpers/linux/module.py
@@ -1,0 +1,226 @@
+# Copyright (c) 2022, Oracle and/or its affiliates.
+# SPDX-License-Identifier: LGPL-2.1-or-later
+
+"""
+Module
+------
+
+The ``drgn.helpers.linux.module`` module contains helpers for working with Linux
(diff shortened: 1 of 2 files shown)


brenns10:
While I'm really happy with these helpers, having now read through the proposed [module API](https://github.com/osandov/drgn/commits/modules), I can see that there's a fair amount of overlap in what will be provided via drgn itself through that API, and the helpers shown here. I don't think it's really worth merging these with the module API on the horizon, so in the effort of keeping the PR board cleaner, I'm closing this. However, the changes shown here (plus some additional code that implements kallsyms for kernel modules) are present in some of the branches described on my DWARFless Debugging issue #176.

PR #234, opened 2023-01-09
helpers/linux/fs: for_each_file: handle NULL files

brenns10:
It's allowable for task->files to be NULL. Handle this case gracefully by returning without yielding any files.

### DIFF
--- a/drgn/helpers/linux/fs.py
+++ b/drgn/helpers/linux/fs.py
@@ -339,7 +339,10 @@ def for_each_file(task: Object) -> Iterator[Tuple[int, Object]]:
     :param task: ``struct task_struct *``
     :return: Iterator of (fd, ``struct file *``) tuples.
     """
-    fdt = task.files.fdt.read_()
+    files = task.files.read_()
+    if not files:
+        return
+    fdt = files.fdt.read_()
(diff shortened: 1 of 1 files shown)


drgn/helpers/linux/fs.py
```diff
@@ -339,6 +339,8 @@ def for_each_file(task: Object) -> Iterator[Tuple[int, Object]]:
     :param task: ``struct task_struct *``
     :return: Iterator of (fd, ``struct file *``) tuples.
     """
+    if not task.files:
+        return
     fdt = task.files.fdt.read_()
```
osandov, line 344: Good call, I've had to work around this a few times, too.

We can avoid an extra `/proc/kcore` read (and I guess be slightly less racy) if we read `task.files` into a value object and use that:

```suggestion
    files = task.files.read_()
    if not files:
        return
    fdt = files.fdt.read_()
```

osandov (changes requested):

drgn/helpers/linux/fs.py
brenns10: Yep, good point, it's updated now :)

osandov (approved):
Thanks! I'll merge as soon as tests pass.

Edit: ignoring those that seem to be having Ubuntu issues

PR #235, opened 2023-01-11
Fix memory leak in `c_format_compound_object`

Svetlitski-FB:
Found via [CodeChecker](https://codechecker.readthedocs.io/en/latest/) static analysis.


osandov:
Wow, I'm impressed that static analysis was able to find that.

PR #236, opened 2023-01-13
cli: Separate the main function from the CLI

brenns10:
It's quite useful to run drgn's CLI from a user-provided script. One reason is saving the drgn core dump and debuginfo setup into a script to more easily start analysis. Another might be to replace drgn's debuginfo loading scheme with one customized to a different environment. Or, a script might want to handle errors by allowing an interactive user to examine the program themselves.

To serve all of these cases and more, separate out the logic of running the drgn CLI from the internal logic relating to argument parsing, choosing the correct target, and loading debuginfo. While this may seem like a simple case of code.interact, there are several important internal features drgn provides in its CLI: a custom displayhook, a custom readline completer, a history file, and of course the logic to import and include the proper drgn APIs and helpers for a core dump.

This means creating a run_cli() function and moving it out of the drgn.internal area.

### DIFF
--- a/drgn/cli.py
+++ b/drgn/cli.py
@@ -1,7 +1,8 @@
 # Copyright (c) Meta Platforms, Inc. and affiliates.
+# Copyright (c) 2023, Oracle and/or its affiliates.
 # SPDX-License-Identifier: LGPL-2.1-or-later
 
-"""drgn command line interface"""
+"""Functions for embedding the drgn CLI."""
 
 import argparse
 import builtins
@@ -10,12 +11,30 @@ import importlib
 import os
 import os.path
 import pkgutil
+import readline
 import runpy
 import shutil
 import sys
-from typing import Any, Dict
+from typing import Any, Callable, Dict, Optional
 
 import drgn
+from drgn.internal.rlcompleter import Completer
+
+__all__ = ("run_interactive", "version_header")
+
+
+def version_header() -> str:
+    """
+    Return the version header printed at the beginning of a drgn session.
+
+    The :func:`run_interactive()` function does not include this banner at the
+    beginning of an interactive session. Use this function to retrieve one line
+    of text to add to the beginning of the drgn banner, or print it before
+    calling :func:`run_interactive()`.
+    """
+    python_version = ".".join(str(v) for v in sys.version_info[:3])
+    libkdumpfile = f'with{"" if drgn._with_libkdumpfile else "out"} libkdumpfile'
+    return f"drgn {drgn.__version__} (using Python {python_version}, elfutils {drgn._elfutils_version}, {libkdumpfile})"
 
 
 def _identify_script(path: str) -> str:
@@ -48,7 +67,7 @@ def _identify_script(path: str) -> str:
     return "core" if e_type == ET_CORE else "elf"
 
 
-def displayhook(value: Any) -> None:
+def _displayhook(value: Any) -> None:
     if value is None:
         return
     setattr(builtins, "_", None)
@@ -71,10 +90,8 @@ def displayhook(value: Any) -> None:
     setattr(builtins, "_", value)
 
 
-def main() -> None:
-    python_version = ".".join(str(v) for v in sys.version_info[:3])
-    libkdumpfile = f'with{"" if drgn._with_libkdumpfile else "out"} libkdumpfile'
-    version = f"drgn {drgn.__version__} (using Python {python_version}, elfutils {drgn._elfutils_version}, {libkdumpfile})"
+def _main() -> None:
+    version = version_header()
     parser = argparse.ArgumentParser(prog="drgn", description="Programmable debugger")
 
     program_group = parser.add_argument_group(
@@ -198,7 +215,6 @@ def main() -> None:
                 prefix = f"\033[33m{prefix}\033[0m"
             missing_debug_info_warning = f"{prefix} {e}"
 
-    init_globals: Dict[str, Any] = {"prog": prog}
     if args.script:
         sys.argv = args.script
         script = args.script[0]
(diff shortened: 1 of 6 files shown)


drgn/cli.py
```diff
@@ -0,0 +1,151 @@
...
+            print("could not read history:", str(e), file=sys.stderr)
+
+    def write_history_file() -> None:
+        try:
+            readline.write_history_file(histfile)
+        except OSError as e:
+            if not quiet:
+                print("could not write history:", str(e), file=sys.stderr)
+
+    atexit.register(write_history_file)
```
osandov, line 119: We should probably replace these `atexit` handlers with something that will be called before returning from this function rather than whenever the caller exits. `contextlib.ExitStack` might be useful.

drgn/cli.py
```diff
@@ -0,0 +1,151 @@
...
+
+    readline.set_history_length(1000)
+    readline.parse_and_bind("tab: complete")
+    readline.set_completer(Completer(init_globals).complete)
+    atexit.register(lambda: readline.set_completer(None))
+
+    sys.displayhook = displayhook
```
osandov, line 126: We modify `sys.path` and `sys.displayhook` globally. Perhaps we should restore them before we return with a `contextlib.ExitStack` handler, as well.

The `readline` settings are another story. I don't see a good way to scope those changes to only affect this function.

I'm imagining a use case where the caller has its own REPL (e.g., @pfactum's [crush](https://codeberg.org/pf-kernel/crush)) and it wants to temporarily drop into the drgn REPL, then return to its own REPL. We can limit a lot of the global state we change, but maybe not `readline`...

drgn/internal/cli.py
```diff
@@ -4,18 +4,15 @@
...
 import os
 import os.path
 import pkgutil
 import runpy
-import shutil
 import sys
 from typing import Any, Dict
 
 import drgn
+from drgn.cli import drgn_version_header, run_cli
```
osandov, line 15: Rather than splitting this into two modules, I'd prefer to move it all to `drgn.cli` and make everything other than `run_interactive()` "private" (i.e., prefix it with "_").

osandov (changes requested):
I like the idea of supporting wrappers around interactive mode. Some comments below.

drgn/cli.py
```diff
@@ -0,0 +1,151 @@
...
+            sys.stdout.write(text)
+    sys.stdout.write("\n")
+    setattr(builtins, "_", value)
+
+
+def run_cli(
+    prog: drgn.Program,
+    additional_globals: Optional[Dict[str, Any]] = None,
+    banner: Optional[str] = None,
+    additional_banner: Optional[str] = None,
```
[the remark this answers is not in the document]
brenns10: Yeah, I think that's a definitely better API. I was struggling a bit with the design because I wanted to be able to preserve, modify, or completely alter the banner. This is much more flexible.

drgn/cli.py
```diff
@@ -0,0 +1,151 @@
...
+        else:
+            text = encoded.decode(sys.stdout.encoding, "strict")
+            sys.stdout.write(text)
+    sys.stdout.write("\n")
+    setattr(builtins, "_", value)
+
+
+def run_cli(
+    prog: drgn.Program,
+    additional_globals: Optional[Dict[str, Any]] = None,
```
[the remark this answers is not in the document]
brenns10: I actually initially had a `globals` and `additional_globals` which mirrored the banner args, but I decided against it because it was too cluttered.

I personally don't see the use case for too much flexibility here, because all you gain (vs `additional_globals`) is the ability to delete from the globals dict, or clear it altogether. Plus, it's a bit extra code to write for the common case of adding a few globals.

But I guess, flexibility should be preferred over brevity here. I wanted this API so I could do stuff beyond what the original design allowed, so I probably shouldn't assume to know how this will be used. So I'm on board with `globals_func` or some name along those lines.

drgn/cli.py
brenns10: Yeah, the `readline` stuff can't be entirely be avoided. I think the best bet here is to simply document what drgn clobbers with readline, and recommend that users save & restore their readline settings.

We use the history mechanism, so we should clear history prior to reading our file, and we should clear the history after writing our file. We should unbind the TAB key and set the completer to the system default. I think these are the two major readline features we use. If they're documented, then it's easy for a caller to call their history file persistence routine prior to drgn, and then reload their history file plus completer.

As far as an `ExitStack`... unless we're allowing the user to provide a context manager, I think it may be simplest to just wrap the `code.interact()` with a `try/finally`.

brenns10:
Sorry for the double push. Missed some `run_cli` references in the docs, and now that I looked at the diff, I realized I could reduce diff churn by keeping all the functions in the same order from `drgn/internal/cli.py`.

brenns10:
(Oof, and again since I accidentally broke `--quiet`, now I'm actually finished updating the PR)

drgn/cli.py
```diff
@@ -154,8 +164,6 @@ def main() -> None:
             )
         elif script_type == "elf":
             sys.exit(f"error: {args.script[0]} is a binary, not a drgn script")
-    else:
-        print(version, file=sys.stderr, flush=True)
```
osandov, line 158: If we can, I'd like to keep printing the version as early as possible, so that if something goes horribly wrong (e.g., segfaults from C) in the initial setup of loading debug info, the output copied and pasted into bug reports will still include the version number.

Maybe we can print it here like before, and have the `banner_func` in `_main` remove the version header? Kinda ugly, but I think it'll work.

drgn/cli.py
```diff
@@ -198,75 +209,125 @@ def main() -> None:
...
+
+    If you are using this function to embed within another application using
+    readline, then you'll need to take special care. While drgn tries to clean
+    up any modifications to global state made during this function, it cannot
+    save and restore the readline settings. In particular, drgn clears and uses
+    the readline history, and it binds a custom completer to the TAB key. You
+    should be sure to save your application's history and clear any custom
+    keybindings or completers prior to calling this function, and then re-read
+    your application's history and restore keybindings and completion functions
+    on return.
```
osandov, line 255: Let's put this in a Sphinx [note](https://sublime-and-sphinx-guide.readthedocs.io/en/latest/notes_warnings.html).

drgn/cli.py
```diff
@@ -198,75 +209,125 @@ def main() -> None:
...
+    :param prog: The program to insert into the CLI namespace as ``prog``
+    :param banner_func: If provided, this function is called with the banner
+      that would normally be printed, and is free to return a modified or new
+      banner instead.
+    :param globals_func: If provided, this function is called with the globals
+      dictionary that would normally be provided to the interactive session.
+      The function may return a new dictionary with any modifications.
+    :param quiet: When true, suppress any warnings related to non-fatal errors.
+    """
+    version = _version_header()
```
osandov, line 266: Nitpick: this is only used in one place in this function, so just put the call below where we use it.
```suggestion
```

osandov (changes requested):
Thanks, a handful of minor comments.

drgn/cli.py
```diff
@@ -198,75 +209,125 @@ def main() -> None:
...
+            banner = banner_func(banner)
+        if globals_func:
+            init_globals = globals_func(init_globals)
         code.interact(banner=banner, exitmsg="", local=init_globals)
+    finally:
+        if "" in sys.path:
+            sys.path.remove("")
+        sys.displayhook = old_displayhook
+        readline.parse_and_bind("tab: self-insert")
+        readline.set_completer(None)
```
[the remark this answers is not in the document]
brenns10: I do not know how I missed the `get_completer()` function -- I recall looking for something like that. I prefer restoring it to how it was, which is usually a clean slate anyway.

drgn/cli.py
brenns10: What about making the `version_banner()` function a public API, and including in the documentation that we don't include it in the banner of `run_interactive()`?

Then, the main function is free to print it as early as possible so it's still there in case of a crazy error. Users that care to replicate the exact startup look of drgn can simply call `version_banner()` and print it early themselves, or add it in their `banner_func()`. 

When you think about it, while the other parts of the banner printed by `run_interactive()` seem objectively useful each time you load the interactive shell (e.g. to answer "which helpers are imported?"), the version banner seems to be something that would only be useful the first time. If I had a program with a command that allowed my users to drop into the drgn shell, I'd probably want to remove the version banner anyway.

osandov: I like that idea.

brenns10:
Ok, should have resolved all the concerns.

I did make one other change, to use [`inspect.cleandoc()`](https://docs.python.org/3/library/inspect.html#inspect.cleandoc) on the banner, so that it can be nicely indented. I thought it made things more readable but you're welcome to veto it :)

osandov (approved):
Thanks! I edited the documentation a bit, and also added restoring the history length that we missed (and dropped the `inspect.cleandoc()`, I like the plain literal).

PR #237, opened 2023-01-13
Document the linux kernel object finder

brenns10:
drgn is able to lookup some special metadata for the Linux kernel, sometimes even without debuginfo. For users who may expect that Program.object() will only return objects corresponding to variables, these metadata are unexpected and can be quite useful. Regardless, they're currently undocumented. Add documentation under Advanced Usage for these, and reference it in the Program.object() docstring.

### DIFF
--- a/_drgn.pyi
+++ b/_drgn.pyi
@@ -185,6 +185,10 @@ class Program:
         """
         Get the object (variable, constant, or function) with the given name.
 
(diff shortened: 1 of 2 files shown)


osandov:
Thanks, it's great to have this documented! I made some edits and will merge it now.

PR #238, opened 2023-01-18
Non radix tree idr support rc1

imran-kn:
This changeset adds support for non-radix tree based idrs in drgn, so that drgn idr helpers can be used with kernels prior tha v4.11.
I have also added idr test cases separately. So far idr test cases were covered by radix tree test cases (since idr helpers were available only for radix-tree base idrs.

### DIFF
--- a/drgn/helpers/linux/idr.py
+++ b/drgn/helpers/linux/idr.py
@@ -7,20 +7,61 @@ IDR
 
 The ``drgn.helpers.linux.idr`` module provides helpers for working with the IDR
 data structure in :linux:`include/linux/idr.h`. An IDR provides a mapping from
-an ID to a pointer. This currently only supports Linux v4.11+; before this,
-IDRs were not based on radix trees.
+an ID to a pointer.
 """
 
+import operator
 from typing import Iterator, Tuple
 
-from _drgn import _linux_helper_idr_find as idr_find
-from drgn import Object
+from _drgn import _linux_helper_idr_find
+from drgn import NULL, IntegerLike, Object, cast, sizeof
 from drgn.helpers.linux.radixtree import radix_tree_for_each
 
 __all__ = (
     "idr_find",
     "idr_for_each",
 )
+
+_IDR_BITS = 8
+_IDR_MASK = (1 << _IDR_BITS) - 1
+
+
+def idr_find(idr: Object, id: IntegerLike) -> Object:
+    """
+    Look up the entry with the given ID in an IDR.
+
+    :param idr: ``struct idr *``
+    :param id: Entry ID.
+    :return: ``void *`` found entry, or ``NULL`` if not found.
+    """
+    # Since Linux kernel commit 0a835c4f090a ("Reimplement IDR and IDA using
+    # the radix tree") (in v4.11), IDRs are backed by radix trees. Before that,
+    # they are a separate data structure. The helper in libdrgn only handles
+    # the radix tree version.
+    if hasattr(idr, "idr_rt"):
+        return _linux_helper_idr_find(idr, id)
+    else:
+        prog = idr.prog_
+        id = operator.index(id)
+
+        if id < 0:
+            return NULL(prog, "void *")
+
+        p = idr.top.read_()
+        if not p:
+            return NULL(prog, "void *")
+
+        n = (p.layer.value_() + 1) * _IDR_BITS
+        MAX_IDR_SHIFT = sizeof(prog.type("int")) * 8 - 1
+        # Equivalent to id > idr_max(p->layer + 1) in the kernel.
+        if id >= (1 << min(n, MAX_IDR_SHIFT)):
+            return NULL(prog, "void *")
+
+        while n > 0 and p:
+            n -= _IDR_BITS
+            p = p.ary[(id >> n) & _IDR_MASK].read_()
+
+        return cast("void *", p)
 
 
 def idr_for_each(idr: Object) -> Iterator[Tuple[int, Object]]:
@@ -30,9 +71,29 @@ def idr_for_each(idr: Object) -> Iterator[Tuple[int, Object]]:
     :param idr: ``struct idr *``
     :return: Iterator of (index, ``void *``) tuples.
     """
+    # Since Linux kernel commit 0a835c4f090a ("Reimplement IDR and IDA using
+    # the radix tree") (in v4.11), IDRs are backed by radix trees.
     try:
-        base = idr.idr_base.value_()
+        idr_rt = idr.idr_rt
     except AttributeError:
-        base = 0
-    for index, entry in radix_tree_for_each(idr.idr_rt.address_of_()):
-        yield index + base, entry
+        voidp_type = idr.prog_.type("void *")
+
+        def aux(p: Object, id: int, n: int) -> Iterator[Tuple[int, Object]]:
+            p = p.read_()
+            if p:
+                if n == 0:
+                    yield id, cast(voidp_type, p)
+                else:
+                    n -= _IDR_BITS
+                    for child in p.ary:
+                        yield from aux(child, id, n)
+                        id += 1 << n
+
+        yield from aux(idr.top, 0, idr.layers.value_() * _IDR_BITS)
+    else:
+        try:
+            base = idr.idr_base.value_()
+        except AttributeError:
+            base = 0
+        for index, entry in radix_tree_for_each(idr_rt.address_of_()):
+            yield index + base, entry
--- a/tests/linux_kernel/kmod/drgn_test.c
+++ b/tests/linux_kernel/kmod/drgn_test.c
@@ -739,6 +739,68 @@ static void drgn_test_xarray_exit(void)
 #endif
 }
 
+// idr
+
+DEFINE_IDR(drgn_test_idr_empty);
+DEFINE_IDR(drgn_test_idr_one);
+DEFINE_IDR(drgn_test_idr_one_at_zero);
+DEFINE_IDR(drgn_test_idr_sparse);
+DEFINE_IDR(drgn_test_idr_dense);
+static unsigned long drgn_test_idr_ptrs[5] = {
+	0x12121212,
+	0x34343434,
+	0x5656565656565656,
+	0x1234567812345678,
+	0x1234567890abcdef,
+};
+
+static int drgn_test_idr_init(void)
+{
+	int ret, i;
+
+	ret = idr_alloc(&drgn_test_idr_one, (void *)0xdeadb00, 66, 67,
+			GFP_KERNEL);
+	if (ret < 0)
+		return ret;
+
+	ret = idr_alloc(&drgn_test_idr_one_at_zero, (void *)0x1234, 0, 1,
+			GFP_KERNEL);
+	if (ret < 0)
+		return ret;
+
+	ret = idr_alloc(&drgn_test_idr_sparse, (void *)0x1234, 1, 2,
+			GFP_KERNEL);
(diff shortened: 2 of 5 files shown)


drgn/helpers/linux/idr.py
```diff
@@ -7,21 +7,101 @@
...
+    for i in range(num_bits):
+        if (num & 1) == 1:
+            pos = i
+
+        num = num >> 1
+
+    return pos + 1
+
+
+def idr_find(idr: Object, idval: int) -> Object:
```
osandov, line 59: To match the original signature, `idval` should be named `id`, and it should be an `IntegerLike` so that `id` can be given as a drgn object:

```suggestion
def idr_find(idr: Object, id: IntegerLike) -> Object:
```

Then below to convert it to an `int`, you can do:

```python3
    id = operator.index(id)
```

You'll of course need:

```python3
import operator
from drgn import IntegerLike
```

drgn/helpers/linux/idr.py
```diff
@@ -7,21 +7,101 @@
...
+
+        hint = idr.hint.value_()
+
+        if hint:
+            prefix = Object(prog, "struct idr_layer", address=hint).prefix.value_()
+
+            if (idval & ~IDR_MASK) == prefix:
+                index = idval & IDR_MASK
+                p = Object(prog, "struct idr_layer", address=hint).ary.value_()[index]
+                return Object(prog, "void *", p)
```
osandov, line 80: Since this is just a fast path/optimization in the kernel and not required for correctness, I think we should just leave it out for simplicity.

```suggestion
```

drgn/helpers/linux/idr.py
```diff
@@ -7,21 +7,101 @@
...
+from drgn import NULL, Object, Program, sizeof
 from drgn.helpers.linux.radixtree import radix_tree_for_each
 
 __all__ = (
     "idr_find",
     "idr_for_each",
 )
 
+IDR_BITS = 8
+IDR_MASK = (1 << IDR_BITS) - 1
```
osandov, line 25: Let's indicate that these are internal:

```suggestion
_IDR_BITS = 8
_IDR_MASK = (1 << _IDR_BITS) - 1
```

drgn/helpers/linux/idr.py
```diff
@@ -30,9 +110,54 @@ def idr_for_each(idr: Object) -> Iterator[Tuple[int, Object]]:
...
+                    pa_idx = pa_idx - 1
+
+            return tmp_id, NULL(prog, "void *")
+
+        next_id = 0
+        next_id, entry = _idr_get_next(next_id)
+        while entry != NULL(prog, "void *"):
+            yield next_id, entry
+            next_id = next_id + 1
+            next_id, entry = _idr_get_next(next_id)
```
osandov, line 163: Doing this recursively is simpler and likely more efficient since it will require fewer reads from the core dump. Something like:

```suggestion
        voidp_type = idr.prog_.type("void *")

        def aux(p: Object, id: int, n: int) -> Iterator[Tuple[int, Object]]:
            p = p.read_()
            if p:
                if n == 0:
                    yield id, cast(voidp_type, p)
                else:
                    n -= IDR_BITS
                    for child in p.ary:
                        yield from aux(child, id, n)
                        id += 1 << n

        yield from aux(idr.top, 0, idr.layers.value_() * IDR_BITS)
```

tests/linux_kernel/kmod/drgn_test.c
```diff
@@ -739,6 +740,68 @@ static void drgn_test_xarray_exit(void)
...
+				              0x1234567812345678,
+					      0x1234567890abcdef};
+static int drgn_test_idr_init(void)
+{
+	int ret, i;
+
+	ret = idr_alloc(&drgn_test_idr_one, (void *)0xdeadb00, 66, 67,
+				GFP_KERNEL);
+	if (ret != 66)
+		return -ENOSPC;
```
osandov, line 761: In the kernel, it seems like the usual error checking pattern for `idr_alloc()` is:

```suggestion
	if (ret < 0)
		return ret;
```

tests/linux_kernel/kmod/drgn_test.c
```diff
@@ -739,6 +740,68 @@ static void drgn_test_xarray_exit(void)
...
+static void drgn_test_idr_exit(void)
+{
+	int i;
+	idr_remove(&drgn_test_idr_one, 66);
+	idr_remove(&drgn_test_idr_one_at_zero, 0);
+	idr_remove(&drgn_test_idr_sparse, 1);
+	idr_remove(&drgn_test_idr_sparse, 0x80);
+	idr_remove(&drgn_test_idr_sparse, 0xee);
+	for (i = 10; i < 15; i++)
+		idr_remove(&drgn_test_idr_dense, i);
```
osandov, line 802: Can you use `idr_destroy()`?

```suggestion
	idr_destroy(&drgn_test_idr_one);
	idr_destroy(&drgn_test_idr_one_at_zero);
	idr_destroy(&drgn_test_idr_sparse);
	idr_destroy(&drgn_test_idr_dense);
```

osandov (changes requested):
Thanks, some comments below! It's a bummer that this is only required for EOL kernels, but at least it's fairly simple.

imran-kn:
> Thanks, some comments below! It's a bummer that this is only required for EOL kernels, but at least it's fairly simple.

Yeah. Actually I am preparing some workqueue helpers and found out that some of test cases (those that need to traverse table of worker_pools) were failing. This was happening because worker_pools are arranged in an idr table and hence test case was failing for 4.9. One option was to skip these tests for kernels < 4.11 but as it was small work, I thought of adding it so that idr helpers and other helpers (which rely on idr, for example workqueue) can be used on these older kernels as well

drgn/helpers/linux/idr.py
```diff
@@ -7,21 +7,101 @@
...
+
+def idr_find(idr: Object, idval: int) -> Object:
+    """
+    Lookup entry with the given ID in an IDR.
+
+    :param idr: ``struct idr *``
+    :param idval: ``Entry ID``
+    :return: ``void *`` found entry, or NULL if not found.
+    """
+    if idr.prog_.type("struct idr").has_member("idr_rt"):
```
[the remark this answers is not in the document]
imran-kn: Done

drgn/helpers/linux/idr.py
imran-kn: Sure. I have incorporated these changes

drgn/helpers/linux/idr.py
imran-kn: Sounds fair to me. I have removed hint usage from the helper

drgn/helpers/linux/idr.py
imran-kn: Done

drgn/helpers/linux/idr.py
imran-kn: Indeed. Thanks for doing this.

tests/linux_kernel/kmod/drgn_test.c
imran-kn: Yeah. The thing I had in mind was that since I am allocating a single ID from a range consisting of 1 number only, I should explicitly check if I got that number as returned ID. But this should not be needed since idr allocation code would take care of range consisting of one number and would return error if it can't allocate that ID. So just checking "ret" will be enough

tests/linux_kernel/kmod/drgn_test.c
imran-kn: Sure. idr_destroy() is better option since all idrs being touched here are the test ones.

drgn/helpers/linux/idr.py
```diff
@@ -7,21 +7,89 @@
...
+
+
+def _max_idr_level(prog: Program) -> int:
+    return (_max_idr_shift(prog) + _IDR_BITS - 1) // _IDR_BITS
+
+
+def _round_up(number: int, multiple: int) -> int:
+    return -((-number) // multiple) * multiple
+
+
```
osandov, line 40: These don't seem to be used anymore.

```suggestion
```

drgn/helpers/linux/idr.py
```diff
@@ -7,21 +7,89 @@
...
+    for i in range(num_bits):
+        if (num & 1) == 1:
+            pos = i
+
+        num = num >> 1
+
+    return pos + 1
+
+
```
osandov, line 59: Also not used.

```suggestion
```

drgn/helpers/linux/idr.py
```diff
@@ -7,21 +7,89 @@
...
+    """
+    Lookup entry with the given ID in an IDR.
+
+    :param idr: ``struct idr *``
+    :param id: ``Entry ID``
+    :return: ``void *`` found entry, or NULL if not found.
+    """
+    id = operator.index(id)
+
+    if hasattr(idr, "idr_rt"):
```
osandov, line 70: We should add a comment mentioning the exact commit that made the change here and some additional explanation.

```suggestion
    # Since Linux kernel commit 0a835c4f090a ("Reimplement IDR and IDA using
    # the radix tree") (in v4.11), IDRs are backed by radix trees. Before that,
    # they are a separate data structure. The helper in libdrgn only handles
    # the radix tree version.
    if hasattr(idr, "idr_rt"):
```

drgn/helpers/linux/idr.py
```diff
@@ -30,9 +98,26 @@ def idr_for_each(idr: Object) -> Iterator[Tuple[int, Object]]:
     :param idr: ``struct idr *``
     :return: Iterator of (index, ``void *``) tuples.
     """
-    try:
-        base = idr.idr_base.value_()
-    except AttributeError:
-        base = 0
-    for index, entry in radix_tree_for_each(idr.idr_rt.address_of_()):
-        yield index + base, entry
+    if idr.prog_.type("struct idr").has_member("idr_rt"):
```
osandov, line 101: In this case, we can follow ["easier to ask for forgiveness than permission"](https://docs.python.org/3/glossary.html#term-EAFP) and assume that `idr_rt` exists, then handle if it doesn't. Something like:

```python3
    # Since Linux kernel commit 0a835c4f090a ("Reimplement IDR and IDA using
    # the radix tree") (in v4.11), IDRs are backed by radix trees.
    try:
        idr_rt = idr.idr_rt
    except AttributeError:
        ... Pre-4.11 version here ...
    else:
        try:
            base = idr.idr_base.value_()
        except AttributeError:
            base = 0
        for index, entry in radix_tree_for_each(idr_rt.address_of_()):
            yield index + base, entry
```

drgn/helpers/linux/idr.py
```diff
@@ -7,21 +7,89 @@
...
+        if id < 0:
+            return NULL(prog, "void *")
+
+        p = idr.top.read_()
+        if not p:
+            return NULL(prog, "void *")
+
+        n = (p.layer.value_() + 1) * _IDR_BITS
+        max_id = _idr_max(prog, p.layer.value_() + 1)
+        if n > max_id:
```
osandov, line 84: Just spotted this typo. This needs to be:

```suggestion
        if id > max_id:
```

I added a test case that catches this mistake in `test_idr_find_one()`: `self.assertIdentical(idr_find(root, 256 + 66), NULL(self.prog, "void *"))`

osandov (commented):
This was almost there, so I'm just going to clean up these last few nits myself and merge this as soon as tests pass. Thanks!

osandov (approved):

PR #239, opened 2023-01-18
Mark immutable attributes as `Final` in type stubs

Svetlitski-FB:
Object attributes which are not changed after the object's creation have been marked with the `Final` type attribute. The primary utility of this change is that some Python type checkers, [such as Pyre](https://pyre-check.org/docs/errors/#optional-attributes), rely on the presence of the `Final` attribute to determine whether `Optional` values have been validated as being present before their contents are accessed.

This is an issue I ran into myself while trying to typecheck a codebase which uses `drgn`'s scripting interface.

### DIFF
--- a/_drgn.pyi
+++ b/_drgn.pyi
@@ -13,6 +13,7 @@ import sys
 from typing import (
     Any,
     Callable,
+    ClassVar,
     Dict,
     Iterable,
     Iterator,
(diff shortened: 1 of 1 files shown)


osandov:
All of the ones you marked look right and I didn't find any you missed, either. Thanks!

PR #240, opened 2023-01-19
python: Implement equality comparison for `drgn.Type`

Svetlitski-FB:
I was baffled by the seemingly nonsensical results I was getting when comparing instances of `drgn.Type` in a script I wrote, and discovered that there currently is no meaningful support for these comparisons. This implementation does happen to rely on the fact that `drgn` caches types internally, but it's certainly more useful than the current status-quo of more or less always returning `False`.


Svetlitski-FB:
The CI error [is a `dnf` issue where a dependency isn't being found](https://download.copr.fedorainfracloud.org/results/@meta/drgn/fedora-eln-i386/05286129-python-drgn/builder-live.log.gz), it appears to be completely unrelated to these changes.

osandov:
This only works for deduplicated types, which excludes non-trivial structs, unions, classes, enums, and function types. See:

https://github.com/osandov/drgn/blob/9730deeb8575000de88d60d380700b57d7e31913/libdrgn/program.h#L86-L98

This means that the following is still false even with this change:

```pycon
>>> type1 = prog.struct_type("foo", 4, (drgn.TypeMember(prog.type("int"), name="bar"),))
>>> type2 = prog.struct_type("foo", 4, (drgn.TypeMember(prog.type("int"), name="bar"),))
>>> type1 == type2
False
```

Those types can't be deduplicated in general because drgn parses member types lazily from DWARF. Deduplicating would require parsing eagerly and doing a deep comparison, which would be unreasonable for large types (e.g., `struct task_struct` in the kernel fans out to an insane number of types).

We actually supported `==` before commit 7d7aa7bf7b070a3270bb99cae7827bb20e37629f by doing a deep comparison. But then people tried using it for type checking (e.g., `if arg.type_ != prog.type("struct task_struct *"): raise Blah`), which was way more expensive than people realized. That convinced me that `==` doesn't convey enough information about what the user wants: do they just want a sanity check that the type is probably what they were expecting (in which case they could compare `drgn.Type.type_name()`), or do they really want a 100% identical type.

So now the deep comparison is only implemented for unit testing: https://github.com/osandov/drgn/blob/9730deeb8575000de88d60d380700b57d7e31913/tests/__init__.py#L122-L133

I didn't have any great ideas for what `==` should mean that wouldn't be surprising to a large number of people (super expensive or inexact), so I punted the problem to users and hoped that they'd implement their own desired semantics in Python.

Svetlitski-FB:
That makes a lot of sense. My use case was relatively niche: I'm wrapping a function that takes as input a `drgn.Type` and returns a new `drgn.Type` with [`@functools.lru_cache`](https://docs.python.org/3/library/functools.html#functools.lru_cache) to avoid doing a lot of unnecessarily duplicated work. Keying on just the type name alone doesn't work for disambiguating since this is interfacing with C++, and therefore multiple types can have the same (unqualified) name but live in different namespaces. I _do_ have a [function `drgn.Type.fully_qualified_name()` on my `drgn` fork](https://github.com/Svetlitski-FB/drgn/commit/59993c96eba87bec76a551d888b585eb27213f0a) which I could use for this purpose, but due to the way it's implemented, it's quite expensive to check.

Basically, my use case is admittedly sort of weird in that I can tolerate false negatives for equality checks, and this behavior probably doesn't belong in core `drgn`, so I'll keep this on my fork. If I happen to come up with some cheap yet generic way of determining type equality, I'll open another PR.

osandov:
`drgn.Type` does have an undocumented `_ptr` attribute that is the address of the underlying `struct drgn_type`:

https://github.com/osandov/drgn/blob/48dd87da985a685ac85dcfb85e63d62f51a14b1f/libdrgn/python/type.c#L439-L442

https://github.com/osandov/drgn/blob/48dd87da985a685ac85dcfb85e63d62f51a14b1f/libdrgn/python/type.c#L48-L51

I don't consider that part of the public API, but I don't see any reason why I would remove it in the near future. Also note that if you create a `drgn.Program`, then delete it, then create another `drgn.Program`, the type pointer addresses could be "reused", but that probably doesn't matter for your usecase.

PR #243, opened 2023-01-23
Fix example for add_memory_segment

Tim---:
The parameters of the read_file function in the example are swapped. This matches the parameters order used by py_memory_read_fn.

### DIFF
--- a/docs/advanced_usage.rst
+++ b/docs/advanced_usage.rst
@@ -63,7 +63,7 @@ program "memory":
         file = open(dev, 'rb')
         size = file.seek(0, 2)
 
-        def read_file(address, count, physical, offset):
+        def read_file(address, count, offset, physical):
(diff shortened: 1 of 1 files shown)


osandov:
Good catch! It looks like I forgot to update this early on in commit c0bc72b0ea77f42e4c1e26fe4b990fb9504b35d1. Could you please add a `Fixes: c0bc72b0ea77 ("libdrgn: use splay tree for memory reader")` line to the commit message? It looks like the commit is also missing a [Signed-off-by](https://github.com/osandov/drgn/blob/main/CONTRIBUTING.rst#signing-off).

Tim---:
Oops, I read about signing-off commits, and forgot it immediately after.
It should be good now.

osandov:
Thanks! The RPM build failure looks totally unrelated.

PR #244, opened 2023-01-23
ARM32 MMU support

Tim---:
As noted, this is only valid for ARM32 processors without LPAE. Until recently I thought LPAE was the same as Aarch64 but it turns out that no... For these, I think the MMU should be identical or similar to aarch64 but I did not look into it.

I took the liberty of unrolling the loops because it's a 2-level MMU with different page handling. Also it allowed me to keep my sanity :).

### DIFF
--- a/libdrgn/arch_arm.c
+++ b/libdrgn/arch_arm.c
@@ -4,6 +4,7 @@
 #include <elf.h>
 
 #include "platform.h" // IWYU pragma: associated
+#include "program.h"
 
 /*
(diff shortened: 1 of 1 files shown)


osandov:
Just wanted to let you know that I haven't forgotten about this. I'm currently working on making drgn's testing setup support other architectures including ARM so that I can properly test this out and review it. I'm going to trying patching drgn's test kernels to support `/proc/kcore` on ARM.

PR #245, opened 2023-01-24
libdrgn: add s390x pagetable walk support

svens-s390:
Add support for walking s390x page tables. This supports up to 5 level page table walking and huge/large pages. In order to figure out the level of paging used, we read the first entry of the pgd, which is always mapped for lowcore access and use the level bits of the next page table. This is because drgn passes mm::pgd as pgtable argument to the walker function which doesn't contain the ASCE bits.

Signed-off-by: Sven Schnelle <svens@linux.ibm.com>

### DIFF
--- a/libdrgn/arch_s390x.c
+++ b/libdrgn/arch_s390x.c
@@ -237,6 +237,232 @@ out:
 	return err;
 }
 
+struct pgtable_data {
+	uint64_t entries[2048];
+	uint64_t addr;
+	int length;
+	int offset;
+};
+
+struct pgtable_iterator_s390x {
+	struct pgtable_iterator it;
+	struct pgtable_data pagetable[5];
+	int levels;
+};
+
+struct dat_level {
+	int bits;
+	int shift;
+	int entries;
+};
+
+static const struct dat_level dat_levels[] = {
+	{ .bits = 8,  .shift = 12, .entries = 256 },	/* PTE */
+	{ .bits = 11, .shift = 20, .entries = 2048 },	/* STE */
+	{ .bits = 11, .shift = 31, .entries = 2048 },	/* RTTE */
+	{ .bits = 11, .shift = 42, .entries = 2048 },	/* RSTE */
+	{ .bits = 11, .shift = 53, .entries = 2048 },	/* RFTE */
+};
+
+#define REGION_INVALID		0x20
+#define SEGMENT_INVALID		0x20
+#define PAGE_INVALID		0x400
+
+#define REGION_LARGE		0x400
+#define SEGMENT_LARGE		0x400
+#define PAGE_LARGE		0x800
+
+static struct drgn_error *
+linux_kernel_pgtable_iterator_create_s390x(struct drgn_program *prog,
+					  struct pgtable_iterator **ret)
+{
+	struct pgtable_iterator_s390x *it = malloc(sizeof(*it));
+	if (!it)
+		return &drgn_enomem;
+
+	*ret = &it->it;
+	return NULL;
+}
+
+static void
+linux_kernel_pgtable_iterator_destroy_s390x(struct pgtable_iterator *_it)
+{
+	free(container_of(_it, struct pgtable_iterator_s390x, it));
+}
+
+static void
+linux_kernel_pgtable_iterator_init_s390x(struct drgn_program *prog,
+					struct pgtable_iterator *_it)
+{
+	struct pgtable_iterator_s390x *it =
+		container_of(_it, struct pgtable_iterator_s390x, it);
+	for (int i = 0; i < 5; i++)
+		it->pagetable[i].addr = UINT64_C(-1);
+	it->levels = 3;
+}
+
+static bool
+entry_is_invalid(uint64_t entry, int level)
+{
+	switch (level) {
+	case 0:
+		return entry & PAGE_INVALID;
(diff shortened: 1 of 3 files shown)


osandov:
Thanks, I think doing it this way is a good compromise. If we need more general page table walking in the future (i.e., for anything other than Linux), then we can work something out. I'll take a closer look at the code this week.

svens-s390:
Did you had a chance to look at this?

osandov:
Just wanted to let you know that I haven't forgotten about this. I'm currently working on getting the testing setup to properly support s390x, then I'll come back to reviewing this.

libdrgn/arch_s390x.c
```diff
@@ -237,6 +237,223 @@ linux_kernel_get_initial_registers_s390x(const struct drgn_object *task_obj,
...
+{
+	struct pgtable_iterator_s390x *it =
+		container_of(_it, struct pgtable_iterator_s390x, it);
+	const uint64_t va = it->it.virt_addr;
+	uint64_t table = _it->pgtable & ~0xfffUL;
+	int levels = 2, length = 2048, offset = 0;
+	uint64_t entry;
+
+	struct drgn_error *err = drgn_program_read_memory(prog, &entry,
+							  table, 8, true);
```
osandov, line 399: This doesn't byte swap the value if the host is little-endian. `drgn_program_read_u64()` will do that for you. Also, the initial page table is a virtual address, so `physical` should be `false` (but from what I understand, the addresses are equal on s390x anyways, so this would be more for documentation/consistency).

```suggestion
	struct drgn_error *err = drgn_program_read_u64(prog, table, 8, false,
						       &entry);
```

Finally, could you please add a comment here explaining what we are doing here? I.e., getting the ASCE bits in a Linux kernel-specific way.

libdrgn/arch_s390x.c
```diff
@@ -249,7 +466,10 @@ const struct drgn_architecture_info arch_info_s390x = {
 	.linux_kernel_get_initial_registers =
 		linux_kernel_get_initial_registers_s390x,
 	.apply_elf_reloc = apply_elf_reloc_s390,
-
+	.linux_kernel_pgtable_iterator_create =  linux_kernel_pgtable_iterator_create_s390,
+	.linux_kernel_pgtable_iterator_destroy = linux_kernel_pgtable_iterator_destroy_s390,
+	.linux_kernel_pgtable_iterator_init = linux_kernel_pgtable_iterator_init_s390,
+	.linux_kernel_pgtable_iterator_next = linux_kernel_pgtable_iterator_next_s390,
```
osandov, line 472: Can we name these all `_s390x` for consistency with the other kernel callbacks?

```suggestion
	.linux_kernel_pgtable_iterator_create =  linux_kernel_pgtable_iterator_create_s390x,
	.linux_kernel_pgtable_iterator_destroy = linux_kernel_pgtable_iterator_destroy_s390x,
	.linux_kernel_pgtable_iterator_init = linux_kernel_pgtable_iterator_init_s390x,
	.linux_kernel_pgtable_iterator_next = linux_kernel_pgtable_iterator_next_s390x,
```

libdrgn/arch_s390x.c
```diff
@@ -237,6 +237,223 @@ linux_kernel_get_initial_registers_s390x(const struct drgn_object *task_obj,
...
+
+		index -= offset;
+		if (it->pagetable[level].addr != table ||
+		    it->pagetable[level].length != length ||
+		    it->pagetable[level].offset != offset) {
+			/*
+			 * It's only marginally more expensive to read 4096 bytes than 8
+			 * bytes, so we always read the full table.
+			 */
+			struct drgn_error *err = drgn_program_read_memory(prog,
```
osandov, line 421: This is shadowing `err` from above. You can reuse it.

```suggestion
			err = drgn_program_read_memory(prog,
```

(drgn doesn't enable `-Wshadow` yet, but I'd like to in the future.)

osandov (commented):
I rebased on main and enabled the memory management test cases with this patch:

```diff
diff --git a/tests/linux_kernel/__init__.py b/tests/linux_kernel/__init__.py
index 9a8c41e..1c76d9a 100644
--- a/tests/linux_kernel/__init__.py
+++ b/tests/linux_kernel/__init__.py
@@ -91,7 +91,7 @@ skip_unless_have_test_kmod = unittest.skipUnless(
     "DRGN_TEST_KMOD" in os.environ, "test requires drgn_test Linux kernel module"
 )
 
-HAVE_FULL_MM_SUPPORT = NORMALIZED_MACHINE_NAME in ("aarch64", "x86_64")
+HAVE_FULL_MM_SUPPORT = NORMALIZED_MACHINE_NAME in {"aarch64", "s390x", "x86_64"}
 
 skip_unless_have_full_mm_support = unittest.skipUnless(
     HAVE_FULL_MM_SUPPORT,
```

and the tests all pass, which is awesome, thank you! I left a handful of minor comments.

svens-s390:
I incorporated all the requested changes, fixed one bug and added one commit to enable the tests like you did.

svens-s390:
gentle ping?

osandov:
Hi, I didn't get to this earlier because I'm on paternity leave, but I managed to sneak in a few minutes to review this :) I made some trivial code style changes, and with that this looks good to me, thanks!

PR #247, opened 2023-01-26
Release the GIL during expensive, pure-C operations

ghost:
Indexing the debug information of large binaries can be very expensive, taking nearly a full minute on some large binaries despite `drgn` aggressively parallelizing this process. As such, it's useful for `drgn` to release the Python Global Interpreter Lock (GIL) while executing long-running sections of pure-C code, so that other Python threads are able to execute concurrently. `drgn` now releases the GIL while initializing `struct drgn_program` and while indexing debug information (both the initial top-level indexing as well as lazy-indexing triggered later by a call to `Program.type()`. These are by no means the only sites where `drgn` could release the GIL, but are the sites which kick off the heaviest pure-C computations.

### DIFF
--- a/libdrgn/python/program.c
+++ b/libdrgn/python/program.c
@@ -44,7 +44,9 @@ int Program_type_arg(Program *prog, PyObject *type_obj, bool can_be_none,
 		name = PyUnicode_AsUTF8(type_obj);
 		if (!name)
 			return -1;
+		Py_BEGIN_ALLOW_THREADS
 		err = drgn_program_find_type(&prog->prog, name, NULL, ret);
+		Py_END_ALLOW_THREADS
 		if (err) {
 			set_drgn_error(err);
 			return -1;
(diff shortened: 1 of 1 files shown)


osandov:
Thanks, I think this makes sense. Thankfully it looks like we're already calling `PyGILState_Ensure()` in all of the right places. The only thing that makes me nervous is that libdrgn itself isn't thread-safe; only one thread can access `prog` at a time. Since the GIL isn't protecting us from this anymore, we probably want to document this. Something like:

```diff
diff --git a/docs/api_reference.rst b/docs/api_reference.rst
index beff6e7..23a692e 100644
--- a/docs/api_reference.rst
+++ b/docs/api_reference.rst
@@ -135,3 +135,12 @@ CLI
 ---
 
 .. drgndoc:: cli
+
+Thread Safety
+-------------
+
+Only one thread at a time should access the same :class:`Program` (including
+:class:`Object`, :class:`Type`, :class:`StackTrace`, etc. from that
+:class:`Program`).
+
+It is safe to use different :class:`Program`\ s from concurrent threads.
```

osandov:
FYI, I ended up doing this with a different approach:

https://github.com/osandov/drgn/commit/06a825f31555217bebdd6da54549a24edd2631b8
https://github.com/osandov/drgn/commit/0ad19dc37b310057b65db7e5c36296352dd0bbae
https://github.com/osandov/drgn/commit/fa8207161853b936fc1dd3db98d45ea2f30a2723

PR #250, opened 2023-01-26
drgn.helpers.linux.slab: Handle SLUB freelist dereference without swab()

brenns10:
Linux commit torvalds/linux@1ad53d9fa3f6 ("slub: improve bit diffusion for freelist ptr obfuscation"), adds a swab() into the freelist_ptr() calculation. Unfortunately we can't detect this except by trial and error. The commit was added in 5.7 and has been included in stable backports, which means that pre-5.7 kernels which haven't recently been updated may still have this behavior. Add in logic to do the trial and error in order to support this oddity.

### DIFF
--- a/drgn/helpers/linux/slab.py
+++ b/drgn/helpers/linux/slab.py
@@ -17,7 +17,7 @@ Linux slab allocator.
 
 import operator
 from os import fsdecode
-from typing import Dict, Iterator, Optional, Set, Tuple, Union, overload
+from typing import Callable, Dict, Iterator, Optional, Set, Tuple, Union, overload
 
 from drgn import (
     NULL,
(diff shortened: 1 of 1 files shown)


sdimitro (approved):
Applied this patch on my local setup and running it against the crash dumps that I initially used to report the issue - the change works as expected.

brenns10:
As an update, looks like this works on Serapheim's core dump without that commit!

drgn/helpers/linux/slab.py
```diff
@@ -258,25 +258,59 @@ def __init__(self, slab_cache: Object) -> None:
             freelist_random = slab_cache.random.value_()
         except AttributeError:
 
-            def _freelist_dereference(ptr_addr: int) -> int:
+            def _simple_dereference(ptr_addr: int) -> int:
                 return self._prog.read_word(ptr_addr)
 
+            self._freelist_dereference = _simple_dereference
```
osandov, line 264: We can get rid of the trivial `_simple_dereference()` wrapper if we do this:

```suggestion
            self._freelist_dereference: Callable[[int], int] = self._prog.read_word
```

drgn/helpers/linux/slab.py
```diff
@@ -258,25 +258,59 @@ def __init__(self, slab_cache: Object) -> None:
...
+
         else:
             ulong_size = sizeof(self._prog.type("unsigned long"))
 
-            def _freelist_dereference(ptr_addr: int) -> int:
+            # Since Linux kernel commit 1ad53d9fa3f6 ("slub: improve bit
+            # diffusion for freelist ptr obfuscation"), a swab() was added to
```
osandov, line 270: It's not as meaningful since it was backported to stable, but it's still useful context to know the original kernel version this was changed in:

```suggestion
            # Since Linux kernel commit 1ad53d9fa3f6 ("slub: improve bit
            # diffusion for freelist ptr obfuscation") (in v5.7), a swab() was added to
```

drgn/helpers/linux/slab.py
```diff
@@ -258,25 +258,59 @@ def __init__(self, slab_cache: Object) -> None:
...
+            # the freelist dereferencing calculation. This commit is backported
+            # widely, but you still encounter some older stable kernels which
```
osandov, line 272: It might be useful to be more specific and say that it was backported to every stable kernel branch with freelist obfuscation, but that's up to you.

osandov (commented):
This is a really clever solution, great work. Some trivial comments to quickly resolve before merging.

brenns10:
All good improvements, I've included them now!

osandov:
Thanks for the quick turnaround! The previous run succeeded, so I'll just merge this now.

PR #253, opened 2023-01-27
fix sock_cgroup_ptr for v.5.15+

marxin:
Since Linux kernel commit 8520e224f547 ("bpf, cgroups: Fix cgroup v2 fallback on v1/v2 mixed mode") (in v5.15), the sock_cgroup_data contains directly cgroup member (of struct cgroup * type).


osandov:
Good catch, it looks like we don't have any test cases for `sock_cgroup_ptr()`. That probably wouldn't be too difficult to test now that we have some cgroup scaffolding in [tests/linux_kernel/helpers/test_cgroup.py](https://github.com/osandov/drgn/blob/main/tests/linux_kernel/helpers/test_cgroup.py).

PR #256, opened 2023-01-27
add sys.py to contrib

marxin:
This script mimics crash>sys and provides the following output:

```
CPUS             16
DATE             Fri Jan 27 20:26:24 2023
UPTIME           1 day, 7:29:37
LOAD AVERAGE     0.00, 0.00, 0.00
TASKS            317
NODENAME         tw
RELEASE          6.1.7-1-default
VERSION          #1 SMP PREEMPT_DYNAMIC Wed Jan 18 11:12:34 UTC 2023 (872045c)
MACHINE          x86_64
MEMORY           12.67 GiB
```

### DIFF
--- a/contrib/kernel_sys.py
+++ b/contrib/kernel_sys.py
@@ -1,0 +1,39 @@
+#!/usr/bin/env drgn
+# Copyright (c) SUSE Linux.
+# SPDX-License-Identifier: LGPL-2.1-or-later
+
+"""Display system information and configuration data."""
+
+from datetime import datetime
+from datetime import timedelta
+
+from drgn.helpers.common.format import number_in_binary_units
+from drgn.helpers.linux import for_each_online_cpu
+from drgn.helpers.linux.mm import totalram_pages
+from drgn.helpers.linux.pid import for_each_task
+from drgn.helpers.linux.sched import loadavg
+
+
+def print_line(key, value):
+    print(f"{key:<16} {value}")
+
+
+uts = prog["init_uts_ns"].name
+
+timekeeper = prog["shadow_timekeeper"]
+date = datetime.fromtimestamp(timekeeper.xtime_sec).strftime("%c")
+uptime = timedelta(seconds=timekeeper.ktime_sec.value_())
+load = ", ".join([f"{v:.2f}" for v in loadavg(prog)])
+totalram = (prog['PAGE_SIZE'] * totalram_pages(prog)).value_()
+
+
+print_line("CPUS", len(list(for_each_online_cpu(prog))))
+print_line("DATE", date)
+print_line("UPTIME", uptime)
+print_line("LOAD AVERAGE", load)
+print_line("TASKS", len(list(for_each_task(prog))))
+print_line("NODENAME", uts.nodename.string_().decode())
+print_line("RELEASE", uts.release.string_().decode())
+print_line("VERSION", uts.version.string_().decode())
+print_line("MACHINE", uts.machine.string_().decode())
+print_line("MEMORY", number_in_binary_units(totalram))


contrib/sys.py
```diff
@@ -0,0 +1,35 @@
...
+def print_line(key, value):
+    print(f"{key:<16} {value}")
+
+
+uts = prog["init_uts_ns"].name
+
+timekeeper = prog["shadow_timekeeper"]
+date = datetime.fromtimestamp(timekeeper.xtime_sec).strftime("%c")
+uptime = timedelta(seconds=timekeeper.ktime_sec.value_())
+avenrun = [f"{x.value_() / (1 << 11):.2f}" for x in prog["avenrun"]]
```
osandov, line 23: This would make a great helper if you're interested in adding that. One complication is that some versions of GCC (including the one used to build the vmtest kernel) generate bad debug info for `avenrun` that doesn't include the array length:

```
$ python3 -m vmtest.vm -k '*' python3 -Bm drgn
...
>>> prog['avenrun']
(unsigned long []){}
```

So you might want to explicitly iterate with something like:

```python3
avenrun = prog["avenrun"]
loadavg = [f"{avenrun[i].value_() / (1 << 11):.2f}" for i in range(3)]
```

contrib/sys.py
```diff
@@ -0,0 +1,35 @@
...
+totalram = (prog['PAGE_SIZE'] * prog['_totalram_pages'].counter).value_()
```
osandov, line 24: If you want to support kernels before 5.0, you'll need to handle https://github.com/torvalds/linux/commit/ca79b0c211af63fa3276f0e3fd7dd9ada2439839:

```
$ python3 -m vmtest.vm -k '4.9.*' python3 -Bm drgn ./contrib/sys.py
...
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/home/osandov/repos/drgn2/drgn/__main__.py", line 16, in <module>
    main()
  File "/home/osandov/repos/drgn2/drgn/internal/cli.py", line 209, in main
    runpy.run_path(script, init_globals=init_globals, run_name="__main__")
  File "<frozen runpy>", line 291, in run_path
  File "<frozen runpy>", line 98, in _run_module_code
  File "<frozen runpy>", line 88, in _run_code
  File "./contrib/sys.py", line 25, in <module>
    totalram = (prog['PAGE_SIZE'] * prog['_totalram_pages'].counter).value_()
                                    ~~~~^^^^^^^^^^^^^^^^^^^
KeyError: '_totalram_pages'
```

As usual, in `contrib`, it's up to you if want to deal with that.

osandov (commented):
Some optional comments to address if you want to. Other than that, it'd be great to give this a name other than `sys.py`. The name makes sense given that it's based on the crash `sys` command, but in Python land it might be confused with the [`sys`](https://docs.python.org/3/library/sys.html) module. Let me know if you can think of any better names.

contrib/sys.py
marxin: Works for me, let's do it as #265.

marxin:
> Let me know if you can think of any better names.

Agree with the name is misleading a bit. What about naming that `system.py`?

contrib/sys.py
osandov: I like that idea. There's a function in the kernel called `totalram_pages()`, so it makes sense to have the matching helper `drgn.helpers.linux.mm.totalram_pages()`.

osandov:
> Agree with the name is misleading a bit. What about naming that `system.py`?

Maybe `kernel_sys.py`? To keep the connection to crash's `sys` command without being too generic.

contrib/sys.py
marxin: Fine, let's do that in #271.

marxin:
> Maybe `kernel_sys.py`?

Works for me.

marxin:
This can be merged now, both prerequisites have been already merged. I've just tested the script both with `4.4` and `6.2`.

PR #260, opened 2023-01-30
Allow building just one Python release

brenns10:
I've had this change sitting around in my local tree for a bit and thought I'd share it. I frequently build & share "beta" development releases in wheel format, and I only build one Python release (Python 3.6 😢 ). This adds an optional parameter to the wheel building scripts to specify a Python version, which saves a fair bit of time. I don't know if anybody else does this but it may be a useful change to have?

### DIFF
--- a/scripts/build_manylinux_in_docker.sh
+++ b/scripts/build_manylinux_in_docker.sh
@@ -21,6 +21,12 @@ yum install -y \
 # https://github.com/pypa/manylinux/issues/731.
 ln -s /usr/share/aclocal/pkg.m4 /usr/local/share/aclocal/
 
+BUILD_ONLY_PYTHON=""
+if [ -n "${1:-}" ]; then
+	# Translate, e.g. 3.10 -> (3, 10)
+	BUILD_ONLY_PYTHON="$(echo "$1" | perl -pe 's/(\d+)\.(\d+)/(\1, \2)/')"
+fi
+
 # Install a recent version of elfutils instead of whatever is in the manylinux
 # image.
 elfutils_version=0.188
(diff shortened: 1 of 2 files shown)


scripts/build_manylinux_in_docker.sh
```diff
@@ -53,12 +59,20 @@ mkdir /tmp/drgn
 cd /tmp/drgn
 tar -xf "/io/$SDIST" --strip-components=1
 
-python_supported() {
-	"$1" -c 'import sys; sys.exit(sys.version_info < (3, 6))'
+build_for_python() {
+	if [ -n "$BUILD_ONLY_PYTHON" ]; then
+		# Build for selected Python release only
+		"$1" -c "import sys; sys.exit(sys.version_info[:2] != $BUILD_ONLY_PYTHON)"
+		return $?
```
osandov, line 66: Nit: this is redundant since it's the last command executed in the function:

```suggestion
```

scripts/build_manylinux_in_docker.sh
```diff
@@ -53,12 +59,20 @@ mkdir /tmp/drgn
...
+	else
+		# Build for all supported Pythons
+		"$1" -c 'import sys; sys.exit(sys.version_info < (3, 6))'
+		return $?
```
osandov, line 70: Same:

```suggestion
```

osandov (commented):
I don't see why not. I'll apply the nitpicks on my side and merge this.

brenns10:
Thanks! I had added the return statements because I thought I had observed the function not working without them. Clearly I was observing some other issue that was present in an earlier version of the changes, because it works now without the return statements :)

PR #265, opened 2023-02-03
Introduce loadavg helper.

marxin:
The helper function returns a list with load
averaged over the last 1, 5 and 15 minutes.

### DIFF
--- a/drgn/helpers/linux/sched.py
+++ b/drgn/helpers/linux/sched.py
@@ -9,14 +9,17 @@ The ``drgn.helpers.linux.sched`` module provides helpers for working with the
 Linux CPU scheduler.
 """
 
+from typing import Tuple
+
 from _drgn import (
     _linux_helper_idle_task as idle_task,
     _linux_helper_task_cpu as task_cpu,
 )
-from drgn import Object
+from drgn import Object, Program
 
 __all__ = (
     "idle_task",
+    "loadavg",
     "task_cpu",
     "task_state_to_char",
 )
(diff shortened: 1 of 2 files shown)


drgn/helpers/linux/system.py
```diff
@@ -0,0 +1,30 @@
...
+__all__ = ("avenrun",)
+
+
+def avenrun(prog: Program) -> List[float]:
+    """
+    Return system load averaged over 1, 5 and 15 minutes as
+    list of three float values.
+
+    >>> avenrun(prog)
+    (2.34, 0.442, 1.33)
```
osandov, line 25: With the current code, this would actually be:

```suggestion
    [2.34, 0.442, 1.33]
```

But seeing it this way, I think it would actually be better to return a `Tuple[float, float, float]`.

osandov (changes requested):
Thanks! I think this helper would make sense in `drgn.helpers.linux.sched`, so no need to add a new module for it. One comment about the return type below.

drgn/helpers/linux/system.py
marxin: Sure!

marxin:
> Thanks! I think this helper would make sense in `drgn.helpers.linux.sched`, so no need to add a new module for it.

Even better, thanks for the hint.

drgn/helpers/linux/sched.py
```diff
@@ -83,3 +86,17 @@ def task_state_to_char(task: Object) -> str:
         return "I"
     else:
         return char
+
+
+def avenrun(prog: Program) -> Tuple[float, float, float]:
```
osandov, line 91: How about we call this `loadavg`, since that name will be more familiar to users than the kernel-internal `avenrun` variable:

```suggestion
def loadavg(prog: Program) -> Tuple[float, float, float]:
```

(Sorry, I should've mentioned this on the last version.)

drgn/helpers/linux/sched.py
```diff
@@ -83,3 +86,17 @@ def task_state_to_char(task: Object) -> str:
...
+    """
+    Return system load averaged over 1, 5 and 15 minutes as
+    list of three float values.
```
osandov, line 94:
```suggestion
    tuple of three float values.
```

osandov (changes requested):

Issue #268, opened 2023-02-03
contrib/tcp_sock.py fails for older kernels (<= `v5.18`)


marxin:
I noticed it fails with:

```
python3 -m vmtest.vm -k '5.18.*' python3 -Bm drgn contrib/tcp_sock.py
...
Traceback (most recent call last):
  File "/usr/lib64/python3.10/runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib64/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "/home/marxin/Programming/drgn/drgn/__main__.py", line 16, in <module>
    _main()
  File "/home/marxin/Programming/drgn/drgn/cli.py", line 225, in _main
    runpy.run_path(script, init_globals={"prog": prog}, run_name="__main__")
  File "/usr/lib64/python3.10/runpy.py", line 289, in run_path
    return _run_module_code(code, init_globals, run_name,
  File "/usr/lib64/python3.10/runpy.py", line 96, in _run_module_code
    _run_code(code, mod_globals, init_globals,
  File "/usr/lib64/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "contrib/tcp_sock.py", line 104, in <module>
    _print_sk(sk)
  File "contrib/tcp_sock.py", line 58, in _print_sk
    tcp_state = TcpState(sk_tcpstate(sk))
  File "/home/marxin/Programming/drgn/drgn/helpers/linux/tcp.py", line 24, in sk_tcpstate
    return cast(sk.prog_["TCP_ESTABLISHED"].type_, sk.__sk_common.skc_state)
_drgn.FaultError: address is not mapped: 0x3fffffab
```

Looking at the code, it seems to me the hlist might be a nulls hlist (`inet_listen_hashbucket` contains a union):
https://github.com/torvalds/linux/blob/4b0986a3613c92f4ec1bdc7f60ec66fea135991f/include/net/inet_hashtables.h#L115-L117

@rdna

PR #269, opened 2023-02-03
Fix TCP listing contrib script

marxin:
In v5.18, it's actually hlist_nulls_head what is used for items of inet_hashinfo::listening_hash. With the change it works for all releases starting with v4.9. The version 4.4 is not supported right now (nulls head list is probably not known at the time).

Fixes: #268

### DIFF
--- a/contrib/tcp_sock.py
+++ b/contrib/tcp_sock.py
@@ -8,12 +8,12 @@ import ipaddress
 import socket
 import struct
 
-from drgn import cast, container_of
+from drgn import cast
 from drgn.helpers.common.type import enum_type_to_class
 from drgn.helpers.linux import (
     cgroup_path,
-    hlist_for_each,
     hlist_nulls_empty,
+    hlist_nulls_for_each_entry,
     sk_fullsock,
     sk_nulls_for_each,
     sk_tcpstate,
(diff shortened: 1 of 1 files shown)


marxin:
Example output:

```
$ M=leap S=tcp_sock.py ; scp contrib/$S root@$M:/tmp/ && ssh root@$M "uname -a && drgn -k /tmp/$S"
tcp_sock.py                                                                                                                                                                                                                                   100% 3001    15.2MB/s   00:00    
Linux leap 5.14.21-150500.37-default #1 SMP PREEMPT_DYNAMIC Mon Jan 16 13:26:08 UTC 2023 (5b4c916) x86_64 x86_64 x86_64 GNU/Linux
LISTEN                                           [::]:22                                         [::]:0      /system.slice/sshd.service
LISTEN                                        0.0.0.0:22                                      0.0.0.0:0      /system.slice/sshd.service
LISTEN                                      127.0.0.1:25                                      0.0.0.0:0      /system.slice/postfix.service
LISTEN                                          [::1]:25                                         [::]:0      /system.slice/postfix.service
ESTABLISHED                            192.168.122.17:22                                192.168.122.1:53202  /system.slice/sshd.service
```

marxin:
> The version 4.4 is not supported right now (nulls head list is probably not known at the time).

Apparently, you removed `4.4` from the testing matrix. Note `4.9` is also EOL since Jan 2023.

osandov (approved):
It looks like this was changed in 5.5 by https://github.com/torvalds/linux/commit/8dbd76e79a16b45b2ccb01d2f2e08dbf64e71e40, and that change was also backported to stable kernels (e.g., https://github.com/gregkh/linux/commit/0a0ee9f2d5c0076e6d45d40a1b50c0411edef40a). So technically this won't work on stable kernels without the fix, but I don't think that matters much in practice.

I added this context to the commit message and will merge now. Thanks!

PR #271, opened 2023-02-07
mm: introduce new helper function `totalram_pages`

marxin:


tests/linux_kernel/helpers/test_mm.py
```diff
@@ -291,3 +292,7 @@ def test_environ(self):
             proc_environ = f.read().split(b"\0")[:-1]
         task = find_task(self.prog, os.getpid())
         self.assertEqual(environ(task), proc_environ)
+
+    @skip_unless_have_full_mm_support
+    def test_totalram_pages(self):
+        self.assertTrue(totalram_pages(self.prog) > 1000)
```
osandov, line 298: I believe that the "MemTotal" line in `/proc/meminfo` comes from `totalram_pages()` in the kernel (see `meminfo_proc_show()` and `si_meminfo()`. So this test case can parse that file and compare the exact value.

tests/linux_kernel/helpers/test_mm.py
```diff
@@ -291,3 +292,7 @@ def test_environ(self):
             proc_environ = f.read().split(b"\0")[:-1]
         task = find_task(self.prog, os.getpid())
         self.assertEqual(environ(task), proc_environ)
+
+    @skip_unless_have_full_mm_support
```
osandov, line 296: This helper doesn't use virtual address translation, so it can run without full MM support.

```suggestion
```

osandov (changes requested):

drgn/helpers/linux/mm.py
```diff
@@ -1093,3 +1094,17 @@ def environ(task: Object) -> List[bytes]:
...
+    Return the total number of RAM memory pages.
+    """
+
+    if "_totalram_pages" in prog:
+        # The variable is present since Linux kernel commit ca79b0c211af63fa32
+        # ("mm: convert totalram_pages and totalhigh_pages variables
+        # to atomic") (in v5.0).
+        return prog["_totalram_pages"].counter.value_()
+    else:
+        return prog["totalram_pages"].value_()
```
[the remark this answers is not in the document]
marxin: Heh :D Thanks for the provided link that explains `EAFP`. Btw. do you make the Commit suggestions directly in the browser, or somehow else?

tests/linux_kernel/helpers/test_mm.py
marxin: Yep, a nice suggestion.

drgn/helpers/linux/mm.py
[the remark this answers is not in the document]
osandov: For trivial suggestions, I do them directly in the browser, but usually I end up copying it into Vim and then copying back. Not the most convenient workflow, but it's good enough :)

PR #275, opened 2023-02-10
tests: test contrib scripts

marxin:
I know the `contrib` folder is low barrier place where one can place semi-working scripts. On the other hand, it's handy to have a control script tested with all supported kernel versions.

What do you think? We may run `TestContrib` only conditionally (based on an env. variable)?
Related to: #272.


marxin:
This should be done properly as part of #283.

PR #277, opened 2023-02-16
docs: use jiffies_64 global variables in the examples

marxin:
I think the jiffies global variable was removed some time ago and I would recommend using a valid variable name so that users can run the mentioned code snippets.

### DIFF
--- a/_drgn.pyi
+++ b/_drgn.pyi
@@ -113,8 +113,8 @@ class Program:
         arbitrarily. In this case, the :meth:`variable()`, :meth:`constant()`,
         :meth:`function()`, or :meth:`object()` methods can be used instead.
 
-        >>> prog['jiffies']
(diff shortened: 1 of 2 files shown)


osandov:
Thanks for reporting this. I believe what happened is that the way that the `jiffies` variable is defined on x86-64 was changed in https://github.com/torvalds/linux/commit/d8ad6d39c35d2b44b3d48b787df7f3359381dcbf. The way it's defined now (and always has been on other architectures) via a linker script means that it doesn't appear in debug info.

I suspect that more kernel developers are familiar with `jiffies` than `jiffies_64`, so instead of updating the documentation, I added a special case to make `jiffies` available again:
https://github.com/osandov/drgn/commit/18a8f69ad8448f16a85cc694d7127b14612b39a7.

marxin:
Thank you for the fix, I welcome it!

PR #279, opened 2023-02-22
Expand DW_AT_upper_bound to include other types of int

jgkamat:
GCC appears to use data8 at -1 when reporting zero length arrays, this patch adds support and a test for that behavior.

dwarf_info.c: Remove check for sdata on quirk for array length == 0

I also added a test to avoid this case regressing in the future

See https://github.com/facebookexperimental/object-introspection/pull/77#issuecomment-1439156748

### DIFF
--- a/libdrgn/dwarf_info.c
+++ b/libdrgn/dwarf_info.c
@@ -6405,11 +6405,11 @@ static struct drgn_error *subrange_length(Dwarf_Die *die,
 
 	dimension->is_complete = true;
 	/*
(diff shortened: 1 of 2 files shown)


osandov:
Could you please mention that this new case only happens in C++ mode in the commit message? Other than that and the lint error, this looks good to me.

osandov:
The Packit failures are my fault, but other than that tests passed, so this is good to go, thanks!

PR #284, opened 2023-02-27
Fix output for of lsmod.py

marxin:
Unify it with what lsmod(8) does.


osandov:
Thanks for the clean fix!

PR #292, opened 2023-03-17
python: Make Type.name and Type.tag always exist

brenns10:
This is a quick drive-by commit to raise a discussion. One incredibly common use case of Types in Linux helpers is to distinguish between the different versions of structures in different kernel versions. It's quite frustrating to write perfectly good code like this (taken from an in-development helper):

    if obj.type_.name == "int":
        pass
    elif obj.type_.type_name() == "atomic_t *":
        pass

And find that even though the intent of this code is clear, it would raise "AttributeError: pointer type does not have a name". Of course, the code could be reordered to avoid this issue. But it seems confusing that type presence or absence is used in this way. Further, some Type fields have a strange "tristate" property, where they could be either absent, present but None, or have a value -- all depending on what kind of Type it is.

In my opinion, that sort of subtlety is probably more correct, but in reality is a difficult API to work with. The Type class already exposes the necessary information to infer the details of some of these tristate properties (e.g. Type.is_complete(), or Type.kind). Consumer code which wants to know those things should directly query those attributes, and not have to tiptoe around unexpected AttributeErrors which could pop up at any time.

This commit only touches the name and tag fields, but the pattern exists for the size, length, is_signed, byteorder fields. I didn't update all fields because I really just want to have the discussion first. Also, I feel that some of these fields (length, is_signed, byteorder) are pretty justifiable to be absent: if you're accessing the byteorder field of a structure type, you're clearly doing something wrong. However, with the name-related fields, it seems less clear.

---

As I wrote this commit message I did realize that `type_name()` is probably the correct API for doing these sorts of type checks. It's just quite unclear. I think it would be worth discussing this more, but I don't necessarily think this change is the only correct answer.

### DIFF
--- a/_drgn.pyi
+++ b/_drgn.pyi
@@ -1815,16 +1815,16 @@ class Type:
     language: Final[Language]
     """Programming language of this type."""
 
-    name: Final[str]
+    name: Final[Optional[str]]
     """
-    Name of this type. This is present for integer, boolean, floating-point,
-    and typedef types.
+    Name of this type. This exists for integer, boolean, floating-point,
+    and typedef types, and is ``None`` otherwise.
     """
 
     tag: Final[Optional[str]]
     """
-    Tag of this type, or ``None`` if this is an anonymous type. This is present
-    for structure, union, class, and enumerated types.
+    Tag of this type, or ``None`` if this is an anonymous type. This is also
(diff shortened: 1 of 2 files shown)


brenns10:
Looking back on this one and I disagree with my own comments now... Closing it :)

PR #301, opened 2023-06-01
Add Python 3.12 to CI

brenns10:
Use the "hyphen release" syntax to specify the latest version between 3.12.0-beta.1 and 3.12. It's unclear whether this will actually work to bring in the real releases of 3.12 when they happen, but I think the only way to find out would be to try it out.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,20 @@
 name: CI
 
-on: [push, pull_request]
+on:
+  workflow_dispatch:
+    inputs:
+      test_all_python_versions:
+        description: "Run tests on all Python versions"
+        type: boolean
+        default: false
+        required: true
+  workflow_call:
+    inputs:
+      test_all_python_versions:
+        description: "Run tests on all Python versions"
+        type: boolean
+        default: false
+        required: true
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
@@ -13,7 +27,9 @@ jobs:
     runs-on: ubuntu-20.04
     strategy:
       matrix:
-        python-version: ['3.11', '3.10', '3.9', '3.8', '3.7', '3.6']
+        python-version: ${{ inputs.test_all_python_versions
+                            && fromJSON('["3.12", "3.11", "3.10", "3.9", "3.8", "3.7", "3.6"]')
+                            || fromJSON('["3.11", "3.6"]')}}
         cc: [gcc, clang]
       fail-fast: false
     env:
@@ -25,6 +41,7 @@ jobs:
         uses: actions/setup-python@v4
         with:
           python-version: ${{ matrix.python-version }}
+          allow-prereleases: true
       - name: Install dependencies
         run: |
           sudo apt-get update
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -1,0 +1,16 @@
+name: Pull Request CI
+
+on:
+  pull_request:
+    types:
+      - opened
+      - synchronize
+      - reopened
+      - labeled
+
+jobs:
+  test:
+    uses: ./.github/workflows/ci.yml
+    if: ${{ github.event.action != 'labeled' || github.event.label.name == 'test-all-python-versions' }}
+    with:
+      test_all_python_versions: ${{ contains(github.event.pull_request.labels.*.name, 'test-all-python-versions') }}
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -1,0 +1,12 @@
+name: Main Branch Push CI
+
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+  test:
+    uses: ./.github/workflows/ci.yml
+    with:
+      test_all_python_versions: true
--- a/.github/workflows/vmtest-build.yml
+++ b/.github/workflows/vmtest-build.yml
@@ -36,3 +36,8 @@ jobs:
           name: kernel-build-logs
           path: build/vmtest/kbuild/*.log
           if-no-files-found: ignore
+  test:
+    needs: build
+    uses: ./.github/workflows/ci.yml
+    with:
+      test_all_python_versions: true


osandov:
The test matrix is getting pretty big now (as evidenced by the rate limits). I wonder if we should pare down the number of Python versions we test by default. Maybe something like: only test the oldest supported release, newest supported release, and the next prerelease by default, but have a way to run all releases manually (if we suspect that the change is that sensitive to the Python version or before cutting a release of drgn). Assuming that's even possible, what do you think of that idea? Would we be sacrificing too much test coverage? I can't think of a time off the top of my head that a version in the middle caught anything that the oldest or latest didn't catch.

brenns10:
I agree that it's tough to come up with an example of a Python breakage that would appear in one version and not the next. For day-to-day CI, I think trimming the Python versions to the oldest, the current release, and the current pre-release makes perfect sense.

Just brainstorming other ideas to mitigate the rate limiting:

1. We can restrict CI to run on pull requests only, rather than on branch push. I definitely have CI run regularly when I don't need it, simply because I pushed a branch.
2. We can add [workflow_dispatch](https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow) as a trigger to allow us to trigger CI on branches we push where we _do_ want CI, even if we don't have a PR open. That way any issues with (1) above are resolved.
3. We could try to use the same `workflow_dispatch` trigger for manually running a more extensive CI against _every_ Python version prior to a release. But that may require duplicating the `ci.yaml` file which seems like a pain.
4. Is there some other place to host the VM images so we won't get rate limited? Some other HTTP server or Git LFS?

The last thing is that there's always the option to look into alternative CI options. The "rate limit" thing is beyond silly: we're downloading the same artifacts over and over again. I run Gitlab CI internally for Oracle Linux + UEK, and I just have my VM images (and test vmcores) sitting on the host which does the CI jobs! No downloads and no rate limits. Obviously that means there's a dedicated CI host which does cost money.

brenns10:
Ok, so I do actually have a solution for this that doesn't involve simply cutting out kernels or Python versions. It looks feasible to me, I can start trying to implement it on my repo to see whether I can get it to work.

We can use a [Github Actions Cache](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows) to cache the latest vmtest kernels, and a list of kernel releases. The basic idea here is that you can define a directory that contains files that are cached, and you give the cache a name and a key. Cached directories can't be modified after creation, but you can create a new one with a different key.

So we can do the following:

1. Update the `vmtest-build` job to create a directory containing all the latest kernels & compilers which would be used by the CI step. Then use the `actions/cache/save` step to create a cache which is keyed by the date of the run, a monday, e.g. `vmtest-ci-YYYY-MM-DD`
2. Update the `ci` job to have an additional step. It would compute the key for the Monday on or prior to the current date, and output this as an environment variable. The `test` steps would use this variable to specify the cache for the test, and they would read the manifest and tar.zstd files from there. The test steps would not need to communicate with the Github API, so long as they found a valid cache. If there was no cache hit, they could continue using the current logic.
3. Tests could also fall back to a previous cache if the new one hasn't been generated yet. This would also facilitate cache maintenance, e.g. by deleting a cache and then re-running the vmtest-build job.

Ideally, this system would avoid pretty much all of the github requests related to downloading kernels during a CI step. I have to assume that it would make each test a bit slower, because github would block each test fetching the cache, rather than doing it in a background thread. However, the benefit of avoiding rate limits seems worth it.

Let me know what you think and I can try to come up with a draft.

osandov:
As long as local runs still work the same way and removing the background download doesn't slow things down too much, I'm okay with this.

Alternatively, I also like your earlier ideas. Extending those to something like this sounds ideal to me:

* Pull requests only run the oldest and newest Python version by default
* We can add a label like `test-all-python-versions` to the pull request to test on the rest
* Branch pushes don't run tests by default
* The weekly vmtest build also triggers a run of all Python versions when it completes to cover the rare bugs like the one you found
* Maybe we can even do something similar for kernel flavors: a `test-all-kernel-flavors` label, and the weekly run also tests all flavors

What do you think?

brenns10:
This makes sense, I can at least tackle the first four in this pull request. I was thinking it makes sense to have an exception for pushing the branch `main` - that should trigger all tests on all Python versions, right? But any other branch should skip the tests altogether.

I won't tackle the caching in this one... while I really like the idea, there's a lot of moving parts and it will need some tinkering around on my fork's Github actions in order to test it out.

brenns10:
I'll test this on a separate branch, otherwise I'll continue spamming this with force pushes and actions, sorry for the noise...

brenns10:
Ok, I can't access the labels to apply one named `run_all_python`, but if that were applied, then the CI tests would be running on all pythons.

As it is, the pull request just does 3.6, 3.11, and 3.12 beta.

The CI workflow now simply triggers the `vmtest` workflow, which takes as input `run_all_python`. The `vmtest` is also triggered by the `push` workflow, which only triggers it on pushes to `main`, and includes all Python versions in that case. Finally, the `vmtest` workflow can be manually triggered once this gets merged to `main`.

.github/workflows/vmtest.yml
```diff
@@ -0,0 +1,54 @@
...
+
+jobs:
+  test:
+    # We're stuck on Ubuntu 20.04 as long as we want to keep testing on Python
+    # 3.6 due to actions/setup-python#544.
+    runs-on: ubuntu-20.04
+    strategy:
+      matrix:
+        python-version: ${{ inputs.run_all_python
+                            && fromJSON('["3.12.0-beta.1 - 3.12", "3.11", "3.10", "3.9", "3.8", "3.7", "3.6"]')
```
osandov, line 31: I found a nicer way to specify 3.12: https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#allow-pre-releases. E.g.:

```yaml
on: push

jobs:
  test:
    strategy:
      matrix:
        python-version:
          - 3.12
          - 3.11
    steps:
      - uses: actions/checkout@v3
      - name: Set up Python ${{ matrix.python-version }}
        uses: actions/setup-python@v4
        with:
          python-version: ${{ matrix.python-version }}
          allow-prereleases: true
      - run: python3 --version
```

P.S. I love this `fromJSON` trick, nice one.

.github/workflows/vmtest.yml
```diff
@@ -0,0 +1,54 @@
+name: VM Test
```
osandov, line 1: I don't know that vmtest is the most accurate name for this (especially if we move the lint job here). How about we keep this main logic in `ci.yml`, then have a new `pull_request.yml` that does the call.

osandov:
I played around with labels in actions and ended up with this test workflow:

```yaml
on:
  pull_request:
    types:
      - opened
      - synchronize
      - reopened
      - labeled

jobs:
  test:
    runs-on: ubuntu-latest
    if: ${{ github.event.action != 'labeled' || github.event.label.name == 'test-all-python-versions' }}
    strategy:
      matrix:
        python-version: ${{ contains(github.event.pull_request.labels.*.name, 'test-all-python-versions') && 
                            fromJSON('["3.12", "3.11"]') ||
                            fromJSON('["3.11"]') }}
    steps:
      - uses: actions/checkout@v3
      - name: Set up Python ${{ matrix.python-version }}
        uses: actions/setup-python@v4
        with:
          python-version: ${{ matrix.python-version }}
          allow-prereleases: true
      - run: |
          echo event_name = ${{ github.event_name }} event action = ${{ github.event.action }} label name = ${{ github.event.label.name }}
          python3 --version
```

That runs all Python versions if the PR was created with the label or if that specific label was added later

.github/workflows/vmtest.yml
brenns10: I wish I could take credit for it, but there's a stack overflow post somewhere in my browser history that gave me the idea.

brenns10:
Thanks for the review! I'm PTO til Tuesday and will turn this around then, if that's ok. Looks like all straightforward changes.

osandov:
Of course! Enjoy your PTO.

.github/workflows/ci.yml
```diff
@@ -1,41 +1,13 @@
...
-        run: python setup.py --version
-      - name: Check with mypy
-        run: pre-commit run --all-files mypy
-      - name: Build and test with ${{ matrix.cc }}
-        run: CONFIGURE_FLAGS="--enable-compiler-warnings=error" python setup.py test -K
+    uses: ./.github/workflows/vmtest.yml
+    with:
+      run_all_python: ${{ contains(github.event.pull_request.labels.*.name, 'run-all-python') }}
 
   lint:
```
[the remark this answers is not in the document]
brenns10: As you suggested elsewhere, lint now remains part of `ci.yaml`, so it will be triggered wherever tests run.

.github/workflows/vmtest.yml
brenns10: Applied the "allow-prereleases: true" that you suggested. I remember reading that in the docs, but I guess I worried that it would allow a prerelease patch version of an already released Python version. For example, 3.11.5-beta even though 3.11.4 is the latest.

But I don't think Python does those sort of patch-level prereleases, so we should be good.

brenns10:
With your suggestions the diff is now much cleaner. FYI I don't believe I have permissions to add labels to PRs. I think only you as maintainer can do that. I don't think it's necessarily a problem, but thought I'd let you know.

.github/workflows/pull_request.yml
```diff
@@ -0,0 +1,17 @@
...
+      - reopened
+      - labeled
+
+jobs:
+  test:
+    uses: ./.github/workflows/ci.yml
+    if: ${{ github.event.action != 'labeled' || github.event.label.name == 'test-all-python-versions' }}
+    with:
+      test_all_python_versions: ${{ contains(github.event.pull_request.labels.*.name, 'test-all-python-versions') }}
+
```
osandov, line 17: Stray newline.

.github/workflows/push.yml
```diff
@@ -0,0 +1,12 @@
+name: Trigger VM Tests on Push
```
osandov, line 1:
```suggestion
name: Main Branch Push CI
```

osandov (commented):
Some super trivial cosmetic stuff, but this looks awesome, thank you!

PR #302, opened 2023-06-02
Add btrfs helpers to contrib

boryas:
btrfs.py is a pile of helpers Omar gave me once that I have cherished ever since. If it is already in the repo, happy to refer to it otherwise.

btrfs_tree_mod_log.py is a script I wrote to do a tree mod log search and rewind while debugging a btrfs issue.

### DIFF
--- a/contrib/btrfs_tree.py
+++ b/contrib/btrfs_tree.py
@@ -1,0 +1,432 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# SPDX-License-Identifier: LGPL-2.1-or-later
+
+import enum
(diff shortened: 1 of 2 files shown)


osandov (changes requested):
Other than the inline comment, this just needs the boring minimal requirements from https://github.com/osandov/drgn/tree/main/contrib#contributing-to-contrib: your `Signed-off-by` and copyright/spdx boilerplate:

```
# Copyright (c) Meta Platforms, Inc. and affiliates.
# SPDX-License-Identifier: LGPL-2.1-or-later
```

PR #306, opened 2023-06-18
drgn.helpers.linux.slab: add support for retrieving objects exclusive…

wenjianhn:
…ly from per-node lists

The current implementation of for_each_allocated_object() is slow as it iterates through every physical page.

This commit adds the ability to retrieve objects exclusively from the per-node partial lists, greatly improving efficiency when searching for the source of vfs caches of dying cgroups or millions of negative dentries.


osandov:
Thanks for the PR! This returns a subset of allocated objects, correct? I.e., it will omit those which come from a full slab? My gut feeling is that this should be a separate helper with an explanation of this subtlety. My naming suggestion is `slab_cache_for_each_partial_slab_object()`, which is really long, so let me know if you have a better idea.

Also, how about the per-CPU partial lists? Those should be fairly quick to access, too, right?

wenjianhn:
Yes, it does. A separate helper was used  in our environment.

slab_cache_for_each_partial_slab_object() is good enough since chatgpt
is able to understand the meaning of it :)

Please review the v2 patch that includes per-CPU partial lists.

wenjianhn:
FYI, the V2 patch helped us find out that https://github.com/spring-projects/spring-framework/issues/27092 was the root cause of 3 million dentries. 

Here is the script that I cooked a few minutes ago:

```Python
#!/usr/bin/env drgn

from drgn.helpers.common.format import escape_ascii_string

from drgn.helpers.linux.slab import (
    for_each_slab_cache,
    slab_cache_for_each_partial_slab_object,
)

from drgn.helpers.linux.fs import dentry_path

# docker ps  |grep container_name
# f7958b87468f        ...       6 hours ago         Up 6 hours  ...

container_id = "f7958b87468f"

for slab in for_each_slab_cache(prog):
    name = escape_ascii_string(slab.name.string_(), escape_backslash=True)
    if name.startswith("dentry") and container_id in name:
        for (i, d) in enumerate(
                slab_cache_for_each_partial_slab_object(slab,
                                                        "struct dentry")):
            if (i < 10000):
                print(dentry_path(d))
        break
```

wenjianhn:
@osandov gentle ping

PR #308, opened 2023-06-22
script/ps: enhanced ps script

cosmologistPiyush:
Enhanced the `ps.py` script and made it more modular to support the following options:
	[-a | --active] thread on each CPU
	[-c | --childs] processes info
	[--cpus] Ready/running proccesses for the given cpu
	[-d | --detailed] information (in addition to standard)
	[--hierarchy] of all parent processes
	[-k | --kthread] information on the given kernel threads
	[-u | --uthread] information on the given user-space threads
	[-t | --threads] information on the given threads

These options make the scipt more versatile and useful.

### DIFF
--- a/contrib/ps.py
+++ b/contrib/ps.py
@@ -2,16 +2,189 @@
 # Copyright (c) Meta Platforms, Inc. and affiliates.
 # SPDX-License-Identifier: LGPL-2.1-or-later
 
-"""A simplified implementation of ps(1) using drgn"""
+"""An implementation of ps(1) using drgn"""
 
-from drgn.helpers.common.format import number_in_binary_units
-from drgn.helpers.linux.mm import totalram_pages
-from drgn.helpers.linux.percpu import percpu_counter_sum
+import sys
+from argparse import ArgumentParser
+from collections import OrderedDict
+
+from drgn import ProgramFlags
+from drgn.helpers.linux.cpumask import for_each_online_cpu
+from drgn.helpers.linux.list import list_for_each_entry, list_count_nodes
+from drgn.helpers.linux.mm import cmdline, totalram_pages
+from drgn.helpers.linux.percpu import per_cpu, percpu_counter_sum
 from drgn.helpers.linux.pid import for_each_task
 from drgn.helpers.linux.sched import task_cpu, task_state_to_char
 
-
 PAGE_SIZE = prog["PAGE_SIZE"].value_()
+
+
+def get_number_of_children(task):
+    """
+    Returns number of children of a given task
+    """
+    return list_count_nodes(task.children.address_of_())
+
+
+def get_cmd(task):
+    """
+    Return the commandline arguments of a given task
+    """
+    return b" ".join(cmdline(task)).decode()
+
+
+def parse_cmdline_args(args):
+    """
+    Command line argument parser
+    """
+    parser = ArgumentParser(prog="drgn ps",
+                            description="Report process status infromation")
+    group = parser.add_mutually_exclusive_group()
+
+    group.add_argument("-a", "--active", action="store_true", default=False,
+                        help="Print active thread on each CPU "
+                        "(data may be inconsistent for live kernel)")
+
+    group.add_argument("-c", "--children", nargs="*", default=None, type=int,
+                        help="Print data about children of the given process(es)")
+
+    group.add_argument("--cpus", nargs="*", default=None, type=int,
+                        help="Ready/running processes for given CPUs")
+
+    parser.add_argument("-d", "--detailed", action="store_true", default=False,
+                        help="Print additional details about the threads")
+
+    group.add_argument("--hierarchy", metavar="PID", nargs="+", default=None,
+                        help="Print parent hierarchy")
+
+    group.add_argument("-k", "--kthread", action='store_true', default=False,
+                        help="Print kernel threads only")
+
+    group.add_argument("-t", "--threads", nargs="+", default=None, type=int,
+                        help="Print detailed information of a given threads")
+
+    group.add_argument("-u", "--uthread", action="store_true", default=False,
+                        help="Print userspace threads only")
+
+    cmd_opts = parser.parse_args(args)
+
+    return cmd_opts
+
+
+def get_rss(task):
+    """
+    Returns the Resident Set Size
+    """
+    try:
+        return PAGE_SIZE * sum([percpu_counter_sum(x) for x in task.mm.rss_stat])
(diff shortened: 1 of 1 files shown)


sourabhjains:
@osandov Any review/comments on this PR?

contrib/ps.py
```diff
@@ -2,23 +2,282 @@
...
+    try:
+        return PAGE_SIZE * sum([percpu_counter_sum(x) for x in task.mm.rss_stat])
+    except (AttributeError, TypeError):
+        return PAGE_SIZE * sum([x.counter for x in task.mm.rss_stat.count])
+
+
+def print_active_tasks():
+    """ Function to print active task on each CPU """
+
+    if prog.flags == (pf.IS_LINUX_KERNEL | pf.IS_LIVE):
```
osandov, line 174: You can just check:

```suggestion
    if prog.flags & pf.IS_LIVE:
```

contrib/ps.py
```diff
@@ -2,23 +2,282 @@
...
+from drgn import ProgramFlags as pf
+from drgn.helpers.linux.cpumask import for_each_online_cpu
+from drgn.helpers.linux.list import list_for_each_entry
+from drgn.helpers.linux.mm import cmdline, totalram_pages
+from drgn.helpers.linux.percpu import per_cpu, percpu_counter_sum
 from drgn.helpers.linux.pid import for_each_task
 from drgn.helpers.linux.sched import task_cpu, task_state_to_char
 
 
+global cmd_opts
```
osandov, line 22: I'd prefer if this was passed around as needed instead of a global.

contrib/ps.py
```diff
@@ -2,23 +2,282 @@
...
+
+def parse_cmdline_args(args):
+    """ Command line argument parser """
+
+    parser = ArgumentParser(prog="drgn ps",
+                            description="Report process status infromation")
+
+    parser.add_argument("-a", "--active", action="store_true", default=False,
+                        help="Print active thread on each CPU" +
+                        "(data may be inconsistent for live kernel)")
```
osandov, line 76: This is missing a space, and you can let the string literals be concatenated without `+`:

```suggestion
                        help="Print active thread on each CPU "
                        "(data may be inconsistent for live kernel)")
```

contrib/ps.py
```diff
@@ -2,23 +2,282 @@
...
+
+    parser.add_argument("-c", "--childs", nargs="*", default=None, type=int,
+                        help="Print data about child processes")
```
osandov, line 79: `children` instead of `childs`, make it more clear that the expected argument is a PID by setting `metavar`, and clarify the help:

```suggestion
    parser.add_argument("-c", "--children", metavar="PID", nargs="*", default=None, type=int,
                        help="Print children of the given process(es)")
```

Also, what does it mean to use `-c` with no arguments?

contrib/ps.py
```diff
@@ -2,23 +2,282 @@
...
+
+    parser.add_argument("--cpus", nargs="*", default=None, type=int,
+                        help="Ready/running processes for given CPUs")
```
osandov, line 82: Similar here, what does it mean to use `--cpus` with no arguments?

contrib/ps.py
```diff
@@ -2,23 +2,282 @@
...
+
+    parser.add_argument("-d", "--detailed", action="store_true", default=False,
+                        help="Print additional details about the threads")
+
+    parser.add_argument("--hierarchy", nargs="+", default=None,
+                        help="Print parent hierarchy")
```
osandov, line 88: Similar to above:

```suggestion
    parser.add_argument("--hierarchy", metavar="PID", nargs="+", default=None,
                        help="Print parent hierarchy of given process(es)")
```

osandov (changes requested):
These are really great additions! I normally wouldn't nitpick so much in `contrib`, but since these are such useful pieces of information, I think it makes sense to make them clean examples for others.

contrib/ps.py
```diff
@@ -2,23 +2,282 @@
 # Copyright (c) Meta Platforms, Inc. and affiliates.
 # SPDX-License-Identifier: LGPL-2.1-or-later
 
-"""A simplified implementation of ps(1) using drgn"""
+""" A simplified implementation of ps(1) using drgn """
```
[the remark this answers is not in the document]
sourabhjains: Sure we will update the doc string.

contrib/ps.py
```diff
@@ -2,23 +2,282 @@
...
-from drgn.helpers.common.format import number_in_binary_units
-from drgn.helpers.linux.mm import totalram_pages
-from drgn.helpers.linux.percpu import percpu_counter_sum
+import sys
+from argparse import ArgumentParser
+from collections import OrderedDict
+
+
+""" from drgn.helpers.common.format import number_in_binary_units """
+from drgn import ProgramFlags as pf
```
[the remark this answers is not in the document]
sourabhjains: Agree.

contrib/ps.py
```diff
@@ -2,23 +2,282 @@
...
 
 
+def get_number_of_children(task):
+    """ Returns number of children of a given task """
+
+    n = 0
+    for _ in list_for_each_entry("struct task_struct",
+            task.children.address_of_(), "sibling"):
+        n += 1
+    return n
```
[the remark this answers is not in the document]
sourabhjains: Sure we will replace above code with the new API.

contrib/ps.py
```diff
@@ -2,23 +2,282 @@
...
+        options += 1
+    if cmd_opts.hierarchy:
+        options += 1
+    if cmd_opts.kthread or cmd_opts.uthread or cmd_opts.threads or cmd_opts.detailed:
+        options += 1
+
+    if options >= 2:
+        parser.error("'-k', '-u', '-t' or '-d' don't go along with either of " +
+        "'-a', '-c' '-h'. Neither do any of '-a', '-c' or '-h' " +
+        "go with each other")
```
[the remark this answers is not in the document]
sourabhjains: Sure.

contrib/ps.py
cosmologistPiyush: `-c` without any arguments will print the children of all the tasks.

contrib/ps.py
cosmologistPiyush: Same as `-c`, even '--cpus' without any arguments will print tasks that are running or ready on all CPUs.

sourabhjains:
Thanks for accepting the patch.

Looking forward to contribute more scripts.

PR #309, opened 2023-06-26
contrib: Add ptdrgn.py script to run interactively via ptpython

brenns10:
I finally got around to getting Drgn working with [ptpython](https://github.com/prompt-toolkit/ptpython) and it's pretty glorious. One particular item that's great is that it has a good multi-line code editor, so you can write multi-line loops, etc, and get automatic indentation, and you can go back and fix mistakes prior to the current line. I added in C syntax highlighting for Drgn objects + types, and I'd like to at some point get it to do completion for object members.

The actual code doing it was just going to live in `~/bin/ptdrgn`, but I figured I could at least share it. It's full of hacks and implementation details, but this is the `contrib` directory, so maaaybe it's ok?

### DIFF
--- a/contrib/ptdrgn.py
+++ b/contrib/ptdrgn.py
@@ -1,0 +1,176 @@
+# Copyright (c) 2023, Oracle and/or its affiliates.
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# SPDX-License-Identifier: LGPL-2.1-or-later
+"""
+Drgn CLI, but using ptpython rather than the standard code.interact()
+
+NOTE: this is definitely a bit of a hack, using implementation details of Drgn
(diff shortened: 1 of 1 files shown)


brenns10:
image

(also pictured: drgn running with CTF support)

brenns10:
It turns out that the Object completions were already there, but they're not easy to find since the first completion is usually `Object.absent_`. I added a hacky completion wrapper which reorders Object attribute completions to prioritize the ones which came from the program, rather than the default ones in drgn.

brenns10:
Here it is for a kmem_cache:
image
And here it is for list_head -- the reordering is very important here or you would not be able to see the struct members easily:
image

osandov:
Eventually we can make this less hacky with something like #319, but this is really cool and there's no reason it can't go in `contrib`.

PR #311, opened 2023-06-27
Treat >64 bit integers as arrays for now.

pcc:
A number of common kernel data structures contain 128-bit integer fields, such as thread.uw.fpsimd_state.vregs in struct task_struct on arm64, and the presence of these fields prevents the whole data structure from being printed. Work around this by treating them as byte arrays for now so that at least the rest of the data structure can be printed properly.

### DIFF
--- a/libdrgn/type.c
+++ b/libdrgn/type.c
@@ -397,6 +397,20 @@ struct drgn_error *drgn_int_type_create(struct drgn_program *prog,
 {
 	struct drgn_error *err;
 
+	/*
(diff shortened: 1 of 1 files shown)


pcc:
I see that this is causing some tests to fail because they are expecting the correct type.

I guess a better solution would be to exclude unprintable fields from the output, but I didn't get a chance to look into how to do that.

osandov:
I played around with some more workarounds for this, but in the end I bit the bullet and implemented enough support for big integers to store them and print them.

PR #315, opened 2023-06-29
Add DRGN_PROGRAM_IS_LOCAL flag

pcc:
Currently, when drgn is used to debug a running program, we assume it to be running on the local machine. However, with remote debugging, this will no longer be the case. To accommodate remote debugging, introduce a flag DRGN_PROGRAM_IS_LOCAL, and use it to decide whether to use /sys/module.

### DIFF
--- a/_drgn.pyi
+++ b/_drgn.pyi
@@ -837,6 +837,11 @@ class ProgramFlags(enum.Flag):
     kernel or a running process).
     """
 
+    IS_LOCAL = ...
+    """
+    The program is running on the local machine.
+    """
+
 class FindObjectFlags(enum.Flag):
     """
     ``FindObjectFlags`` are flags for :meth:`Program.object()`. These can be
--- a/libdrgn/drgn.h.in
+++ b/libdrgn/drgn.h.in
@@ -490,6 +490,8 @@ enum drgn_program_flags {
 	DRGN_PROGRAM_IS_LINUX_KERNEL = (1 << 0),
 	/** The program is currently running. */
 	DRGN_PROGRAM_IS_LIVE = (1 << 1),
+	/** The program is running on the local machine. */
+	DRGN_PROGRAM_IS_LOCAL = (1 << 2),
 };
 
 /**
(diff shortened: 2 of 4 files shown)


osandov:
I probably won't be able to look at the remainder of your pull requests until next week, but I'll start with this one since it's easy :) I have been meaning to add something like this flag, so I like the idea. I wonder if we should expand the definition so that:

`(prog->flags & (DRGN_PROGRAM_IS_LIVE | DRGN_PROGRAM_IS_LOCAL)) == (DRGN_PROGRAM_IS_LIVE | DRGN_PROGRAM_IS_LOCAL)` means that the program is currently running on the local machine, and
`(prog->flags & (DRGN_PROGRAM_IS_LIVE | DRGN_PROGRAM_IS_LOCAL)) == DRGN_PROGRAM_IS_LOCAL` means that the program is a core dump on the local machine. Then, `DRGN_PROGRAM_IS_LOCAL` could be used to make decisions for cases like the discussion in #312: if the core dump is local, then it's optimal to do big batched reads, but if the program is remote, then we should optimize for minimizing the amount of data that we transfer.

What do you think?

pcc:
It may make sense to do the opposite: `(prog->flags & DRGN_PROGRAM_IS_LOCAL) == 0` should activate code that usually does fewer round trips while transferring a similar amount of data. This is because with remote debugging, latency plays a big role for these small reads.

A good example of how we can achieve fewer round trips would be to cache page tables even between reads, and use the local page table cache as a "speculative cache". We would read all levels of the page table at once, together with the data that we intend to read, using the addresses stored in the cached PTEs. If it turned out that any of the PTEs had changed, we can repeat the transfer one level at a time. In this way we can read by virtual address in a single round trip under ideal conditions.

But it's possible that this would be faster for local debugging as well, as it would mean fewer syscalls (assuming we would use something like readv on the core dump), so we may not want to make this conditional at all.

For the time being, I think it makes sense to only use this flag to control functional behavior and not for optimizations, unless it's shown that different code paths are faster depending on local/remote.

osandov:
Agreed that for now we can limit this only controlling functional behavior. Pulled, thanks.

PR #316, opened 2023-06-30
Add VMCOREINFO to Linux special objects

brenns10:


brenns10:
(No difference except rebasing on `main` so I can rebase my upcoming symbol PR on top of it)

libdrgn/linux_kernel_object_find.inc.strswitch
```diff
@@ -33,6 +33,10 @@ struct drgn_error *linux_kernel_object_find(const char *name, size_t name_len,
 			if (flags & DRGN_FIND_OBJECT_CONSTANT)
 				return linux_kernel_get_vmemmap(prog, ret);
 			break;
+		@case "VMCOREINFO"@
```
osandov, line 36: Nit: these are sorted lexicographically, could you please move this to after `UTS_RELEASE`?

libdrgn/program.h
```diff
@@ -168,6 +168,10 @@ struct drgn_program {
 		bool pgtable_l5_enabled;
 		/** PAGE_SHIFT of the kernel (derived from PAGE_SIZE). */
 		int page_shift;
+
+		/** The original vmcoreinfo data, to expose as an object */
+		char *data;
+		size_t data_size;
```
osandov, line 174: What do you think of naming this `raw` and `raw_size`?

osandov (changes requested):
Some style nits, but this looks good, thanks!

libdrgn/program.h
brenns10: That makes more sense because it's already inside `vmcoreinfo` struct, so it is clearer to understand what it contains. Data is very generic and is usually a bad variable name :)

osandov (approved):

PR #338, opened 2023-07-08
libdrgn: add support for remote debugging via OpenOCD

pcc:
OpenOCD may be used to establish a JTAG or SWD connection to a remote
debugging target. This patch adds the necessary support to drgn for
debugging a remote target using OpenOCD.

A new memory backend is added that issues memory transactions using
OpenOCD's Tcl interface. With remote debugging using direct physical
memory accesses there is no need for the kernel to have core dump
support. Instead, we use the debugging information in the vmlinux
file, together with information available in the physical memory of
the target. A new arch-specific hook is added for figuring out the
information in vmcoreinfo using the vmlinux file and the physical load
address of the kernel, which must be supplied by the user.

With this patch, drgn can also act as a debugger for microcontroller
firmware and other bare-metal programs that run without an MMU, using
the same OpenOCD memory backend used for kernel debugging.

(Since this has dependencies on my other PRs, I included their commits in this PR.)

### DIFF
--- a/drgn/cli.py
+++ b/drgn/cli.py
@@ -101,6 +101,9 @@ def _main() -> None:
         "-k", "--kernel", action="store_true", help="debug the running kernel (default)"
     )
     program_group.add_argument(
+        "--openocd", metavar="PATH", type=str, help="debug the given kernel on a remote target using OpenOCD"
+    )
+    program_group.add_argument(
         "-c", "--core", metavar="PATH", type=str, help="debug the given core dump"
     )
     program_group.add_argument(
@@ -135,6 +138,23 @@ def _main() -> None:
         action="store_const",
         const={},
         help="don't load any debugging symbols that were not explicitly added with -s",
+    )
+
+    openocd_group = parser.add_argument_group(title="OpenOCD options")
+    openocd_group.add_argument(
+        "--openocd-host", metavar="HOST", type=str, default="localhost", help="hostname of OpenOCD server"
+    )
+    openocd_group.add_argument(
+        "--openocd-port", metavar="PORT", type=str, default="6666", help="TCL port number of OpenOCD server"
+    )
+    openocd_group.add_argument(
+        "--openocd-addr", metavar="ADDR", type=lambda x: int(x,0), help="debug kernel at given physical address"
+    )
+    openocd_group.add_argument(
+        "--openocd-nommu", action="store_true", help="debug a program that does not use the MMU"
+    )
+    openocd_group.add_argument(
+        "--openocd-tap", metavar="TAP", type=str, help="debug kernel using given TAP"
     )
 
     parser.add_argument(
@@ -182,6 +202,16 @@ def _main() -> None:
             prog.set_core_dump(args.core)
         elif args.pid is not None:
             prog.set_pid(args.pid or os.getpid())
+        elif args.openocd:
+            if args.openocd_tap is None:
+              print("error: --openocd-tap required when using OpenOCD", file=sys.stderr)
+              sys.exit(1)
+            if args.openocd_addr is None and args.openocd_nommu is None:
+              print("error: --openocd-addr or --openocd-nommu required when using OpenOCD", file=sys.stderr)
+              sys.exit(1)
+            prog.set_openocd(vmlinux=args.openocd, host=args.openocd_host,
+                             port=args.openocd_port, tap=args.openocd_tap,
+                             mmu=not args.openocd_nommu, paddr=args.openocd_addr or 0)
         else:
             prog.set_kernel()
     except PermissionError as e:
--- a/libdrgn/arch_aarch64.c
+++ b/libdrgn/arch_aarch64.c
@@ -466,6 +466,12 @@ linux_kernel_pgtable_iterator_next_aarch64(struct drgn_program *prog,
 	}
 }
 
+static uint64_t untagged_addr_aarch64(uint64_t addr)
+{
+	/* Apply TBI by sign extending bit 55 into bits 56-63. */
+	return (((int64_t)addr) << 8) >> 8;
+}
+
 const struct drgn_architecture_info arch_info_aarch64 = {
 	.name = "AArch64",
 	.arch = DRGN_ARCH_AARCH64,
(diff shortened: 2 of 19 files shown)


pcc:
Looks like some of the RPM builders failed, but I didn't have much luck finding the actual build failure amongst the various log files under the Details link. Am I missing anything?

osandov:
The logs are showing "No matching package to install: 'libkdumpfile-devel'", which is also happening on the main branch. So something appears to be busted with Fedora ELN, unrelated to your change.

pcc:
Hi @osandov, do you think you will have a chance to look at this (or any of my other PRs) soon? I'm still actively using them for my kernel development work. I tried rebasing them onto the current main branch and ran into some merge conflicts, so for now I'm sticking with an older version of drgn. It looks like #312 and #315 still apply cleanly; if those go in I can try to rebase the rest.

osandov:
Thanks for the ping, I'll make some time to look at your PRs this week.

@brenns10 since you've been looking at recovering the vmcoreinfo from core dumps lacking virtual address information, would you mind taking a look at that part of the top commit of this PR and sharing how that compares to what you've been doing?

brenns10:
Yeah definitely, I'll take a look -- probably tomorrow morning.

drgn/cli.py
```diff
@@ -182,6 +202,16 @@ def _main() -> None:
             prog.set_core_dump(args.core)
         elif args.pid is not None:
             prog.set_pid(args.pid or os.getpid())
+        elif args.openocd:
+            if args.openocd_tap is None:
+              print("error: --openocd-tap required when using OpenOCD", file=sys.stderr)
+              sys.exit(1)
```
brenns10, line 208: Tiny suggestion, Python allows you to shorten this:
```python
              sys.exit("error: --openocd-tap required when using OpenOCD")
```
Non-int args are printed to stderr and an error code of 1 is returned.

libdrgn/arch_aarch64.c
```diff
@@ -408,7 +409,7 @@ linux_kernel_pgtable_iterator_next_aarch64(struct drgn_program *prog,
 	int level;
 	uint16_t num_entries = it->last_level_num_entries;
 	uint64_t table = it->it.pgtable;
-	bool table_physical = false;
+	bool table_physical = it->it.pgtable_phys;
```
brenns10, line 412: I haven't delved deep into the pgtable code, so I was not aware that `table_physical` was here for pretty much all of them and that it would be that easy to properly support a physical page table base.

I had a hack where I fooled drgn into using a physical `swapper_pg_dir` address. But your way is the proper way, I like it a lot!

libdrgn/arch_aarch64.c
```diff
@@ -466,6 +467,141 @@ linux_kernel_pgtable_iterator_next_aarch64(struct drgn_program *prog,
 	}
 }
 
+static struct drgn_error *
+linux_kernel_init_vmcoreinfo_from_phys_aarch64(struct drgn_program *prog,
+					       Elf *vmlinux, GElf_Ehdr *ehdr,
+					       uint64_t text_pa)
```
brenns10, line 473: Having `text_pa` makes sense for OpenOCD targets. But for other "bare metal" cases (e.g. linux core dumps without proper virtual address ranges provided), it may be that you can compute `text_pa` by some other data in the vmcoreinfo note. I don't know if there's one function signature that could cover all the necessary data that an arch-specific hook would need...

We may find ourselves needing to take a page out of crash's book, which allows you to use `-m key=value` to specify arch-specific key-value data to help it make sense of the target. This would be require more work, but it would give architectures the flexibility to handle all sorts of interesting cases.

brenns10 (commented):
I only reviewed the top commit as mentioned, though I took a brief look at the whole series to get the context as best I could. I'd be happy to go through more code and have more of a reviewing angle -- here I was discussing more architecture rather than nitty gritty code review.

I think the approach here for physical page table base addresses is much better than my hacks for handling missing data. I think it would be best if we could try to avoid fiddling with the vmcoreinfo structure and lying about being a Linux kernel, but that could be something we work on down the line.

Incidentally @pcc what sort of test setup are you using? I have a Raspberry Pi 4B that I've (somewhat) used with OpenOCD and some bare metal programs before. I might try to give it a go on that, once I dust everything off :)

libdrgn/arch_aarch64.c
```diff
@@ -466,6 +467,141 @@ linux_kernel_pgtable_iterator_next_aarch64(struct drgn_program *prog,
...
+{
```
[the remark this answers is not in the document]
pcc: Yes, I think it would be worth trying to remove dependencies on Linux-isms such as vmcoreinfo as drgn starts to become useful for debugging bare-metal programs other than the Linux kernel. I personally don't think it needs to be done immediately though.

drgn/cli.py
pcc: Ack, I'll take care of that when rebasing.

libdrgn/arch_aarch64.c
pcc: Longer term I was thinking that we might want to try removing the `pgtable_phys` flag and have all page table iterator accesses go via physical memory. That should simplify the code a bit. This seems like it should also work with core dumps because the physical address can be calculated from the program headers in the core dump.

libdrgn/arch_aarch64.c
pcc: > I don't know if there's one function signature that could cover all the necessary data that an arch-specific hook would need...

To me it seems that there should be different APIs for "starting from a vmcoreinfo note" and "starting from a physical address". In each case the information discovery mechanism is very different. For example, initialization from PA doesn't even require the kernel to be configured with `CONFIG_CRASH_CORE`.

> We may find ourselves needing to take a page out of crash's book, which allows you to use -m key=value to specify arch-specific key-value data to help it make sense of the target. This would be require more work, but it would give architectures the flexibility to handle all sorts of interesting cases.

Possibly. I suppose that another option would be to get any information we need added to the vmlinux file in some form, and then we can read it from there. The page size in the arm64 kernel header is one example. You might also have noticed changes in the code to tolerate missing information (i.e. changes to `linux_kernel.c` to tolerate missing `nr_section_roots`).

pcc:
> Incidentally @pcc what sort of test setup are you using? I have a Raspberry Pi 4B that I've (somewhat) used with OpenOCD and some bare metal programs before. I might try to give it a go on that, once I dust everything off :)

Unfortunately my target is proprietary. I was thinking about getting a Raspberry Pi or something so that I can verify that this all works on publicly available hardware, but if you want to give it a go I'd be happy to help.

PR #339, opened 2023-07-12
helpers: support `dmesg` with RHEL 7

pfactum:
RHEL 7 kernel still uses `struct log *` for a structured kernel log instead of `struct printk_log *`, so lets try to support it.

Tested on:

* `3.10.0-229`
* `3.10.0-1160.80.1`
* `3.10.0-1160.83.1`

This should have no impact on existing supported cases.


drgn/helpers/linux/printk.py
```diff
@@ -158,7 +158,12 @@ def add_record(current_id: int) -> None:
 
 
 def _get_printk_records_structured(prog: Program) -> List[PrintkRecord]:
-    printk_logp_type = prog.type("struct printk_log *")
+    try:
+        printk_logp_type = prog.type("struct printk_log *")
+    except LookupError:
+        # RHEL 7
```
osandov, line 164: Could you please add a more specific reference to the kernel commit that changed this? I just added this guideline to `CONTRIBUTING` for future reference: https://github.com/osandov/drgn/blob/main/CONTRIBUTING.rst#linux-kernel-helpers

```suggestion
        # Before Linux kernel commit 62e32ac3505a ("printk: rename struct log
        # to struct printk_log") (in v3.11), records were "struct log" instead
        # of "struct printk_log".
```

osandov (changes requested):
Thanks! I don't have a RHEL installation handy, so I'll trust your testing. One nit below, but this looks fine otherwise.

drgn/helpers/linux/printk.py
pfactum: Sure, thanks for checking this. I've just pushed the change.

PR #340, opened 2023-07-17
helpers/linux/mm: support RHEL 7 compound_head implementation

pfactum:
Kernel upstream commit 1d798ca3f1 changed this, but RHEL 7 still has got old implementation, which breaks `find_containing_slab_cache()` helper

### DIFF
--- a/drgn/helpers/linux/mm.py
+++ b/drgn/helpers/linux/mm.py
@@ -588,9 +588,25 @@ def PageCompound(page: Object) -> bool:
     :param page: ``struct page *``
     """
     page = page.read_()
-    return bool(
-        (page.flags & (1 << page.prog_["PG_head"])) or (page.compound_head.read_() & 1)
-    )
+    # Since Linux kernel commit 1d798ca3f164 ("mm: make compound_head()
+    # robust") (in v4.4), PG_head is always defined, and a tail page has the
+    # least significant bit of compound_head set. Before that, there is no
+    # compound_head (and no fake head pages). Instead, if
+    # CONFIG_PAGEFLAGS_EXTENDED=y, then PG_head and PG_tail are defined.
+    # Otherwise, there is only PG_compound, and PG_reclaim is set for tail
+    # pages and clear for head pages.
+    try:
+        PG_head = page.prog_["PG_head"]
+    except KeyError:
+        return bool(page.flags & (1 << page.prog_["PG_compound"]))
+    else:
+        flags = page.flags.read_()
+        if flags & (1 << PG_head):
+            return True
+        try:
+            return bool(page.compound_head.read_() & 1)
+        except AttributeError:
+            return bool(flags & (1 << page.prog_["PG_tail"]))
 
 
 # HugeTLB Vmemmap Optimization (HVO) creates "fake" head pages that are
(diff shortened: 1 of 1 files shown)


drgn/helpers/linux/mm.py
```diff
@@ -637,12 +637,20 @@ def compound_head(page: Object) -> Object:
     :return: ``struct page *``
     """
     page = page.read_()
-    head = page.compound_head.read_()
+    try:
+        head = page.compound_head.read_()
+    except AttributeError:
+        # RHEL 7
+        head = page.first_page.address_
```
osandov, line 644: Looking at [`compound_head()` in v4.3](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/include/linux/mm.h?h=v4.3#n429), shouldn't this be:

```suggestion
    try:
        head = page.compound_head.read_()
    except AttributeError:
        # Before Linux kernel commit 1d798ca3f164 ("mm: make compound_head()
        # robust") (in v4.4), the head page is in page->first_page, and there
        # are no fake head pages.
        return page.first_page.read_() if PageTail(page) else page
```

I don't have a kernel that old available, could you please test it out? If it works, feel free to update the PR.

drgn/helpers/linux/mm.py
```diff
@@ -637,12 +637,20 @@ def compound_head(page: Object) -> Object:
...
     if head & 1:
         return cast(page.type_, head - 1)
     # Handle fake head pages (see _page_is_fake_head()).
     if page.flags & (1 << page.prog_["PG_head"]):
-        head = page[1].compound_head.read_()
+        try:
+            head = page[1].compound_head.read_()
+        except AttributeError:
+            # RHEL 7
+            head = page[1].first_page.address_
```
osandov, line 653: My suggestion above removes the need for this change, since we'll never get to this case if `compound_head` didn't exist in the first place.

osandov (changes requested):

drgn/helpers/linux/mm.py
pfactum: This does work, although i had to also modify `PageTail()` helper in a similar way. Forced-pushed now, please check. Thank you.

pfactum: But this does not work on a RHEL 7 s390x vmcore once `slab_object_info()` is called:

```
  File "/home/onatalen/work/src/crush/crush/helpers/common.py", line 173, in demangle
    obj_info = slab_object_info(ctx.prog, addr)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/drgn/helpers/linux/slab.py", line 534, in slab_object_info
    return _get_slab_cache_helper(slab_cache).object_info(page, slab, addr)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/drgn/helpers/linux/slab.py", line 428, in _get_slab_cache_helper
    return type(slab_cache)
           ^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/drgn/helpers/linux/slab.py", line 244, in __init__
    self._slab_cache_size = slab_cache.size.value_()
                            ^^^^^^^^^^^^^^^^^^^^^^^^
_drgn.FaultError: could not read memory from kdump: Excluded page: 0x30038
```

osandov:
I pushed another attempt at this to https://github.com/osandov/drgn/tree/rhel-7-compound_head. It tries to handle the `PageCompound()`/`PageHead()`/`PageTail()` implementations from kernel 4.4. Please test that and let me know if it works for you.

pfactum:
@osandov Hi. Sorry for the delay. I've just tried https://github.com/osandov/drgn/tree/rhel-7-compound_head and it seems to work with both x86 and s390x vmcores I have for RHEL 7.

osandov (approved):
Thanks for testing. I just pushed that version and will merge it once tests pass.

PR #342, opened 2023-08-07
Type checking for helpers

brenns10:
As we discussed on Slack, here's a neat implementation I threw together. It's not suitable to merge but I thought it was a neat approach. This approach uses the docstrings, which seems to be the way that Drgn does it. An alternative would be to leverage [typing.Annotated](https://docs.python.org/3/library/typing.html#typing.Annotated) but I think that would look ugly, confuse some type checkers, and probably not be easy to integrate into Sphinx.

Missing: tests, also checking for lists / tuples / dicts. I suppose it would also be possible to type check generators if we wanted, but that's a questionable idea.

### DIFF
--- a/drgn/helpers/common/type.py
+++ b/drgn/helpers/common/type.py
@@ -10,12 +10,18 @@ with types in ways that aren't provided by the core drgn library.
 """
 
 import enum
+import inspect
+import os
+import re
 import typing
-from typing import Container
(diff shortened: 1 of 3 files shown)


brenns10:
This was a neat concept, but more trouble than it's worth. It never got used in drgn-tools either. Closing.

PR #343, opened 2023-08-09
Handle zero pc in x86_64 fallback unwind

inwardvessel:
Sometimes a stack trace cannot be unwound because of pc being zero and we end up with a single empty frame: `0: 0x0`. In these cases, we can assume a NULL function pointer was called and then modify RIP and RSP so that they are associated with the frame of the "caller of the NULL function pointer". This allows the stack to be unwound, revealing every frame below the NULL call. A new function is added to perform this manipulation of registers, and is then used within the x86_64 fallback unwinder where attempting to get registers from the frame pointer fails and pc is found to be zero.

### DIFF
--- a/libdrgn/arch_x86_64.c
+++ b/libdrgn/arch_x86_64.c
@@ -243,12 +243,57 @@ get_registers_from_frame_pointer(struct drgn_program *prog,
 	return NULL;
 }
 
+// Unwind from a call instruction, assuming that nothing else has been changed
+// since.
+static struct drgn_error *unwind_call(struct drgn_program *prog,
+				      struct drgn_register_state *regs,
+				      struct drgn_register_state **ret)
+{
+	struct drgn_error *err;
+
+	struct optional_uint64 rsp =
+		drgn_register_state_get_u64(prog, regs, rsp);
+	if (!rsp.has_value)
+		return &drgn_stop;
+
+	// Read the return address from the top of the stack.
+	uint64_t ret_addr;
+	err = drgn_program_read_u64(prog, rsp.value, false, &ret_addr);
+	if (err) {
+		if (err->code == DRGN_ERROR_FAULT) {
+			drgn_error_destroy(err);
+			err = &drgn_stop;
+		}
+		return err;
+	}
+
+	// Most of the registers are unchanged.
+	struct drgn_register_state *tmp = drgn_register_state_dup(regs);
+	if (!tmp)
+		return &drgn_enomem;
+
+	// The PC and rip are the return address we just read.
+	drgn_register_state_set_pc(prog, tmp, ret_addr);
+	drgn_register_state_set_from_u64(prog, tmp, rip, ret_addr);
+	// rsp is after the saved return address.
+	drgn_register_state_set_from_u64(prog, tmp, rsp, rsp.value + 8);
+	*ret = tmp;
+	return NULL;
+}
+
 static struct drgn_error *
 fallback_unwind_x86_64(struct drgn_program *prog,
 		       struct drgn_register_state *regs,
 		       struct drgn_register_state **ret)
 {
 	struct drgn_error *err;
+
+	// If the program counter is 0, it's likely that a NULL function pointer
+	// was called. Assume that the only thing we need to unwind is a single
+	// call instruction.
+	struct optional_uint64 pc = drgn_register_state_get_pc(regs);
+	if (pc.has_value && pc.value == 0)
+		return unwind_call(prog, regs, ret);
 
 	struct optional_uint64 rbp =
 		drgn_register_state_get_u64(prog, regs, rbp);
--- a/libdrgn/register_state.c
+++ b/libdrgn/register_state.c
@@ -12,11 +12,16 @@
 	&_state->buf[_state->regs_size];		\
 })
 
+static inline uint32_t drgn_register_state_bitset_size(uint16_t num_regs)
+{
+	return ((uint32_t)num_regs + CHAR_BIT + 1) / CHAR_BIT;
+}
+
 struct drgn_register_state *drgn_register_state_create_impl(uint32_t regs_size,
 							    uint16_t num_regs,
 							    bool interrupted)
 {
-	uint32_t bitset_size = ((uint32_t)num_regs + CHAR_BIT + 1) / CHAR_BIT;
+	uint32_t bitset_size = drgn_register_state_bitset_size(num_regs);
 	size_t size;
 	struct drgn_register_state *regs;
 	if (__builtin_add_overflow(regs_size, bitset_size, &size) ||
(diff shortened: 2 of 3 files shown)


libdrgn/register_state.c
```diff
@@ -31,6 +31,27 @@ struct drgn_register_state *drgn_register_state_create_impl(uint32_t regs_size,
 	return regs;
 }
 
+struct drgn_register_state *drgn_register_state_dup(
+		const struct drgn_register_state *regs)
+{
+	uint32_t bitset_size = ((uint32_t)regs->num_regs + CHAR_BIT + 1) / CHAR_BIT;
+	size_t size = 0;
```
osandov, line 38: No need to initialize this, it's overwritten in a couple of lines.
```suggestion
	size_t size;
```

libdrgn/register_state.c
```diff
@@ -31,6 +31,27 @@ struct drgn_register_state *drgn_register_state_create_impl(uint32_t regs_size,
...
+	    !(ret = malloc(size)))
+		return NULL;
+
+	ret->module = regs->module;
+	ret->regs_size = regs->regs_size;
+	ret->num_regs = regs->num_regs;
+	ret->interrupted = regs->interrupted;
+	ret->_pc = regs->_pc;
+	ret->_cfa = regs->_cfa;
+	memcpy(ret->buf, regs->buf, regs->regs_size);
```
osandov, line 51: This can be a single `memcpy`:

```suggestion
	memcpy(ret, regs, size);
```

libdrgn/arch_x86_64.c
```diff
@@ -258,9 +306,20 @@ fallback_unwind_x86_64(struct drgn_program *prog,
...
+
 			drgn_error_destroy(err);
 			err = &drgn_stop;
+
+			/*
+			 * Check if a NULL function pointer was called
+			 * and adjust registers so unwinding is possible.
+			 */
+			pc = drgn_register_state_get_pc(regs);
+			if (pc.has_value && pc.value == 0)
```
osandov, line 319: I think it'd be better to check this NULL call case before trying to use the frame pointer. (If frame pointers are enabled, then after a call of NULL, the frame pointer will still point to the caller of the function that called NULL, and thus we'll skip the function that called NULL, which is the one we care about.)

libdrgn/arch_x86_64.c
```diff
@@ -243,6 +243,54 @@ get_registers_from_frame_pointer(struct drgn_program *prog,
...
+	if (!rsp.has_value)
+		return &drgn_stop;
+
+	/*
+	 * Read RSP which contains the return address to be used
+	 * as the new RIP value.
+	 */
+	uint64_t postcall_rip;
+	err = drgn_program_read_memory(prog, &postcall_rip, rsp.value,
+		sizeof(postcall_rip), false);
```
osandov, line 272: This won't properly byte swap if you're debugging a x86-64 core dump from a big-endian host. Instead:

```suggestion
	err = drgn_program_read_u64(prog, rsp.value, false, &postcall_rip);
```

libdrgn/arch_x86_64.c
```diff
@@ -243,6 +243,54 @@ get_registers_from_frame_pointer(struct drgn_program *prog,
...
+	if (err) {
+		if (err->code == DRGN_ERROR_FAULT) {
+			drgn_error_destroy(err);
+			err = &drgn_stop;
+		}
+		return err;
+	}
+
+	struct drgn_register_state *tmp = drgn_register_state_dup(regs);
```
osandov, line 281: Need to check that the allocation succeeded.

```suggestion
	struct drgn_register_state *tmp = drgn_register_state_dup(regs);
	if (!tmp)
		return &drgn_enomem;
```

osandov (changes requested):
Thanks for the fix! A few minor comments. Could you please also split the new `drgn_register_dup()` function into its own commit?

inwardvessel:
made all requested changes.

libdrgn/register_state.c
```diff
@@ -31,6 +31,21 @@ struct drgn_register_state *drgn_register_state_create_impl(uint32_t regs_size,
 	return regs;
 }
 
+struct drgn_register_state *drgn_register_state_dup(
+		const struct drgn_register_state *regs)
```
osandov, line 35: Style nit: let's format it like this:

```suggestion
struct drgn_register_state *
drgn_register_state_dup(const struct drgn_register_state *regs)
```

libdrgn/register_state.h
```diff
@@ -130,6 +130,9 @@ struct drgn_register_state *drgn_register_state_create_impl(uint32_t regs_size,
 					DRGN_REGISTER_NUMBER(last_reg) + 1,	\
 					interrupted)
 
+struct drgn_register_state *drgn_register_state_dup(
+		const struct drgn_register_state *regs);
```
osandov, line 134: Same style nit as above, and let's document this:

```suggestion
/**
 * Create a copy of a @ref drgn_register_state.
 *
 * @return New register state on success, @c NULL on failure to allocate memory.
 */
struct drgn_register_state *
drgn_register_state_dup(const struct drgn_register_state *regs);
```

libdrgn/arch_x86_64.c
```diff
@@ -243,12 +243,68 @@ get_registers_from_frame_pointer(struct drgn_program *prog,
...
 		       struct drgn_register_state **ret)
 {
 	struct drgn_error *err;
+	struct optional_uint64 pc;
+
+	/*
+	 * Check if a NULL function pointer was called
+	 * and adjust registers so unwinding is possible.
+	 */
+	pc = drgn_register_state_get_pc(regs);
```
osandov, line 305: Style nit, and let's expand on the comment a little bit:

```suggestion
	/*
	 * If the program counter is 0, it's likely that a NULL function pointer
	 * was called. Assume that the only thing we need to unwind is a single
	 * call instruction.
	 */
	struct optional_uint64 pc = drgn_register_state_get_pc(regs);
```

libdrgn/arch_x86_64.c
```diff
@@ -243,12 +243,68 @@ get_registers_from_frame_pointer(struct drgn_program *prog,
...
+ * so that RIP points to the caller frame post-call instruction and RSP is
+ * increased to point to the top of the caller frame.
+ * This can be used in cases where, for example, a NULL function pointer
+ * was called and as a result the RIP (program counter) value of zero prevents
+ * any unwinding.
+ */
+static struct drgn_error *
+unwind_call(struct drgn_program *prog,
+		       struct drgn_register_state *regs,
+		       struct drgn_register_state **ret)
```
osandov, line 257: Style:
```suggestion
static struct drgn_error *unwind_call(struct drgn_program *prog,
				      struct drgn_register_state *regs,
				      struct drgn_register_state **ret)
```

osandov (changes requested):
A few more small style things, but this looks good functionally. Could you please also add a bit more explanation to the commit messages? It'd be nice to have some background about how you found this issue and what it fixes for you.

inwardvessel:
requested changes made and expanded commit messages.

osandov:
I made some small style/comment tweaks, so this is good to go. Thanks, this is a great addition! I wonder if we'll want something similar for other architectures.

PR #347, opened 2023-08-17
Add the ability to run drgn against the live kernel as non-root user

brenns10:
(As described in Slack)

`/proc/kcore` can only be opened by root, so live debugging has been restricted to running drgn as root. But running drgn as root has downsides:

1. The `PYTHONPATH` does not include your home directory
2. Python may drop `pyc` files as root (this can be avoided with `-B`)
3. All files and other actions your script might take are done by root.

It would be very nice to be able to run Drgn as your current user, and only use escalated privileges to open `/proc/kcore`. One approach would be to drop privileges after creating the program, which improves numbers 2 and 3. Unfortunately the PYTHONPATH, PATH, USER, and HOME environment variables would be unchanged, so much of the standard library would still believe you are root (e.g. `getpass.getuser()`). This can be worked around, but it's just a bit messy.

Instead, we can run a helper program with sudo that will transmit the opened file descriptor back to Drgn via a Unix socket.

To do this, we add `Program.set_fd()` and `program_from_fd()`  which behave just like `set_core_dump()`, etc, except they take the FD rather than the path.

We add a helper script to do the "unix socket magic", and then we tie it all together in the CLI code. When the CLI code runs against the kernel, if it is not running as root, it will use the helper function to get the file descriptor. If the CLI is already running as root, nothing changes.

Running as non-root breaks drgn's module & section iterators, since drgn expects to be able to read data out of `/sys/modules` -- unfortunately some of it has `400` permissions and can only be read by root. Thankfully, that code is just an optimized code path. We can disable that if we're running as a non-root user.

Putting this all together, we have a way to run drgn as non-root user, and have all the features we've come to expect!

### DIFF
--- a/_drgn.pyi
+++ b/_drgn.pyi
@@ -451,7 +451,7 @@ class Program:
             return an :class:`Object`.
         """
         ...
-    def set_core_dump(self, path: Path) -> None:
+    def set_core_dump(self, path: Union[Path, int]) -> None:
         """
         Set the program to a core dump.
 
@@ -459,7 +459,7 @@ class Program:
         mapped executable and libraries. It does not load any debugging
         symbols; see :meth:`load_default_debug_info()`.
 
-        :param path: Core dump file path.
+        :param path: Core dump file path or open file descriptor.
         """
         ...
     def set_kernel(self) -> None:
@@ -888,12 +888,12 @@ def filename_matches(haystack: Optional[str], needle: Optional[str]) -> bool:
     """
     ...
 
-def program_from_core_dump(path: Path) -> Program:
+def program_from_core_dump(path: Union[Path, int]) -> Program:
     """
     Create a :class:`Program` from a core dump file. The type of program (e.g.,
     userspace or kernel) is determined automatically.
 
-    :param path: Core dump file path.
+    :param path: Core dump file path or open file descriptor.
     """
     ...
 
--- a/drgn/cli.py
+++ b/drgn/cli.py
@@ -20,6 +20,7 @@ from typing import Any, Callable, Dict, Optional
 
 import drgn
 from drgn.internal.rlcompleter import Completer
+from drgn.internal.sudohelper import open_via_sudo
 
 __all__ = ("run_interactive", "version_header")
 
@@ -264,7 +265,10 @@ def _main() -> None:
         elif args.pid is not None:
             prog.set_pid(args.pid or os.getpid())
         else:
-            prog.set_kernel()
+            try:
+                prog.set_kernel()
+            except PermissionError:
+                prog.set_core_dump(open_via_sudo("/proc/kcore", os.O_RDONLY))
     except PermissionError as e:
         print(e, file=sys.stderr)
         if args.pid is not None:
--- a/drgn/internal/sudohelper.py
+++ b/drgn/internal/sudohelper.py
@@ -1,0 +1,71 @@
+# Copyright (c) Stephen Brennan <stephen@brennan.io>
+# SPDX-License-Identifier: LGPL-2.1-or-later
+
+"""Helper for opening a file as root and transmitting it via unix socket"""
+import array
+import os
+from pathlib import Path
+import pickle
+import socket
+import subprocess
+import sys
+import tempfile
+from typing import Union
+
+
+def open_via_sudo(
+    path: Union[Path, str],
+    flags: int,
+    mode: int = 0o777,
+) -> int:
+    """Implements os.open() using sudo to get permissions"""
+    # Currently does not support dir_fd argument
+    with tempfile.TemporaryDirectory() as td:
+        sockpath = Path(td) / "sock"
+        with socket.socket(socket.AF_UNIX, socket.SOCK_DGRAM) as sock:
+            sock.bind(str(sockpath))
+            subprocess.check_call(
+                [
+                    "sudo",
+                    sys.executable,
+                    "-B",
+                    __file__,
+                    sockpath,
+                    path,
+                    str(flags),
+                    str(mode),
+                ],
+            )
+            fds = array.array("i")
+            msg, ancdata, flags, addr = sock.recvmsg(
+                4096, socket.CMSG_SPACE(fds.itemsize)
+            )
+            for level, typ, data in ancdata:
+                if level == socket.SOL_SOCKET and typ == socket.SCM_RIGHTS:
+                    data = data[: fds.itemsize]
+                    fds.frombytes(data)
+                    return fds[0]
+            raise pickle.loads(msg)
+
+
+def main() -> None:
+    sockpath = sys.argv[1]
+    filename = sys.argv[2]
+    flags = int(sys.argv[3])
+    mode = int(sys.argv[4])
+
+    sock = socket.socket(socket.AF_UNIX, socket.SOCK_DGRAM)
+    sock.connect(sockpath)
+    try:
+        fd = os.open(filename, flags, mode)
+        fds = array.array("i", [fd])
+        sock.sendmsg(
+            [b"success"],
+            [(socket.SOL_SOCKET, socket.SCM_RIGHTS, fds)],
+        )
+    except Exception as e:
+        sock.sendmsg([pickle.dumps(e)])
+
+
+if __name__ == "__main__":
+    main()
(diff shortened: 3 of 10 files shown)


brenns10:
(Sorry for the force pushes, my pre-commit hooks weren't setup on my non-work laptop)

libdrgn/program.c
```diff
@@ -584,6 +577,37 @@ drgn_program_set_core_dump(struct drgn_program *prog, const char *path)
...
+drgn_program_set_fd(struct drgn_program *prog, int fd)
+{
+	struct drgn_error *err;
+	char path[sizeof("/proc/self/fd/") + 20];
+
+	err = drgn_program_check_initialized(prog);
+	if (err)
+		return err;
+
+	snprintf(path, sizeof(path), "/proc/self/fd/%d", fd);
```
osandov, line 590: This path is only used for error reporting, right?

This is how we do this statically-sized buffer for `printf()` in other places:

```suggestion
	err = drgn_program_check_initialized(prog);
	if (err)
		return err;

	#define FORMAT "/proc/self/fd/%d"
	char path[sizeof(FORMAT) - sizeof("%d") + max_decimal_length(int) + 1];
	snprintf(path, sizeof(path), FORMAT, fd);
	#undef FORMAT
```

osandov (commented):
I haven't looked at the actual sudohelper in detail yet, but I wanted to get the comments on the rest out before I head out for the weekend. Thanks!

libdrgn/program.c
brenns10: Nice, I just manually did the max decimal length of uint64_t and added one. Should have known you had a macro for it... :P

_drgn.pyi
```diff
@@ -462,6 +462,16 @@ class Program:
         :param path: Core dump file path.
         """
         ...
+    def set_fd(self, fd: int) -> None:
```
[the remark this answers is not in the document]
brenns10: The only thing that strikes me about it is that the name makes it sounds like `/proc/kcore` is not allowed. But that's still consistent with `set_core_dump()`, which allows `/proc/kcore` as well.

And I see what you mean... I hadn't considered the possible interpretations of that name with respect to debugging live userspace processes. I'll go with `set_core_dump_fd()`.

libdrgn/linux_kernel.c
```diff
@@ -1550,11 +1550,21 @@ report_kernel_modules(struct drgn_debug_info_load_state *load,
...
+	 * kernel debugging session, if the program was created based on a file
+	 * descriptor for /proc/kcore passed to us via Unix socket by a task
+	 * with root permissions.
+	 *
+	 * This fast path can be disabled via an environment variable for
+	 * testing.
 	 */
 	bool use_sys_module = false;
-	if (prog->flags & DRGN_PROGRAM_IS_LIVE) {
+	if ((prog->flags & DRGN_PROGRAM_IS_LIVE) && getuid() == 0) {
```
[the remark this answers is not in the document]
brenns10: It's definitely feasible, I was even able to preserve the speedup for using `/sys/module/*/notes` which at least seems to be reliably 444 permissions.

brenns10:
What I'm learning is that `git rebase` does not run pre-commit hooks. Sorry for the noise.

osandov (commented):
I only looked at `sudohelper` this time around, so I haven't looked at your other updates yet.

drgn/internal/sudohelper.py
```diff
@@ -0,0 +1,126 @@
...
+                    str(flags),
+                    str(mode),
+                ],
+                check=False,
+            )
+            if proc.returncode == 0:
+                return recv_fd(sock)
+            else:
+                recv_error(sock)
+                assert False  # No return from recv_error
```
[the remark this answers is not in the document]
brenns10: This is much more straightforward, thanks!

libdrgn/linux_kernel.c
```diff
@@ -825,14 +827,40 @@ struct kernel_module_section_iterator {
...
+						  struct kernel_module_iterator *kmod_it)
 {
 	struct drgn_error *err;
 
+	it->sections_dir = NULL;
+	it->i = 0;
+	it->name = NULL;
+	/* it->nsections = mod->sect_attrs->nsections */
+	err = drgn_object_member(&kmod_it->tmp1, &kmod_it->mod,
+					"sect_attrs");
```
osandov, line 840: I think there's some wonky indentation in this function.

libdrgn/linux_kernel.c
```diff
@@ -1570,8 +1589,11 @@ report_kernel_modules(struct drgn_debug_info_load_state *load,
 	/*
 	 * If we're debugging the running kernel, we can use
 	 * /sys/module/$module/notes and /sys/module/$module/sections instead of
+	 * getting the equivalent information from the core dump. But this is
```
osandov, line 1592: Seems to be a stray line.

```suggestion
```

drgn/cli.py
```diff
@@ -263,8 +264,10 @@ def _main() -> None:
             prog.set_core_dump(args.core)
         elif args.pid is not None:
             prog.set_pid(args.pid or os.getpid())
-        else:
+        elif os.geteuid() == 0:
             prog.set_kernel()
+        else:
+            prog.set_core_dump_fd(open_via_sudo("/proc/kcore", os.O_RDONLY))
```
osandov, line 270: Again instead of checking the UID, how about:

```suggestion
        else:
            try:
                prog.set_kernel()
            except PermissionError:
                prog.set_core_dump_fd(open_via_sudo("/proc/kcore", os.O_RDONLY))
```

drgn/internal/sudohelper.py
```diff
@@ -0,0 +1,88 @@
...
+        assert typ == socket.SCM_RIGHTS
+        data = data[: fds.itemsize]
+        fds.frombytes(data)
+        return fds[0]
+    else:
+        try:
+            exc = pickle.loads(msg)
+        except Exception:
+            exc = Exception(msg.decode())
+        raise exc
```
osandov, line 33: Instead of asserting the level and type, let's scan through the list and look for one that matches what we want, and fail if we don't find one. Also, we might as well blow up if we can't unpickle the message, because `msg.decode()` likely won't work either.

```suggestion
    fds = array.array("i")
    msg, ancdata, flags, addr = sock.recvmsg(4096, socket.CMSG_SPACE(fds.itemsize))
    for level, typ, data in ancdata:
        if level == socket.SOL_SOCKET and typ == socket.SCM_RIGHTS:
            data = data[: fds.itemsize]
            fds.frombytes(data)
            return fds[0]
    raise pickle.loads(msg)
```

At this point, this is small enough that I'd prefer to shove it in `open_via_sudo()` rather than its own function.

drgn/internal/sudohelper.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+def send_fd(sock: socket.socket, fd: int) -> None:
+    """Send a file descriptor on the socket"""
+    # Borrows heavily from the Python docs for sock.sendmsg()
+    fds = array.array("i", [fd])
+    sock.sendmsg(
+        [b"success"],
+        [(socket.SOL_SOCKET, socket.SCM_RIGHTS, fds)],
+    )
```
osandov, line 43: Similarly here, this is small enough to go directly in `main()` instead of a separate function.

osandov (changes requested):
This looks awesome, just a few more tweaks.

_drgn.pyi
```diff
@@ -462,6 +462,16 @@ class Program:
         :param path: Core dump file path.
         """
         ...
+    def set_core_dump_fd(self, fd: int) -> None:
```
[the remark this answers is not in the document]
brenns10: This is a very elegant way to do it! Thanks :)

brenns10:
Thanks for the excellent review feedback, I'm so excited about this now! Try `python -m contrib.ptdrgn` for extra fun, since I did rebase this :D

osandov (approved):
There was still some funky indentation in `kernel_module_section_iterator_init_no_sys_module`, so I went ahead and fixed that and added "and program_from_core_dump()" to the commit subject for the first commit. I'll merge this now, thanks! This is a really nice feature.

brenns10:
Thanks for the quick cleanup, sorry I missed it! I'm already using this on my development machines so glad it can be in `main` now :)

PR #352, opened 2023-08-28
libdrgn: add drgn_error_catch()

brenns10:
A common enough pattern is to catch errors by their code, and then destroy them and handle them in some way. Add a helper to handle this. The helper always destroys the error and always NULLs the error pointer (which may otherwise be left to point at a stale error).

A less common short-cut is to directly compare error values to global error variables. In particular, "err == &drgn_not_found" should really be replaced with a catch statement that catches all DRGN_ERROR_LOOKUP. However, the others (drgn_stop, drgn_line_wrap, and drgn_enomem) are specialized enough that they probably don't merit this treatment.

Apply the new helper to the whole codebase.


brenns10:
I still think this is a good idea but will need to be redone given all the changes that have happened in main. I'll pick up this idea at a later date. Keeping the branch around for reference though.

PR #353, opened 2023-09-04
drgn.helpers.linux.net: add some room helpers

mina:
They were useful in my debugging of an issue. I thought I'd contribute them.


drgn/helpers/linux/net.py
```diff
@@ -260,3 +266,88 @@ def skb_shinfo(skb: Object) -> Object:
...
+def skb_headlen(skb: Object) -> Object:
+    return skb.len - skb.data_len
+
+
+def skb_headroom(skb: Object) -> Object:
+    return skb.data - skb.head
+
+
+def skb_is_nonlinear(skb: Object) -> Object:
+    return skb.data_len
```
osandov, line 280: I think it would make more sense for this one to return `bool`:

```suggestion
def skb_is_nonlinear(skb: Object) -> bool:
    return bool(skb.data_len)
```

drgn/helpers/linux/net.py
```diff
@@ -260,3 +266,88 @@ def skb_shinfo(skb: Object) -> Object:
...
     else:
         return cast("struct skb_shared_info *", skb.end)
+
+
+def skb_headlen(skb: Object) -> Object:
+    return skb.len - skb.data_len
+
+
+def skb_headroom(skb: Object) -> Object:
+    return skb.data - skb.head
```
osandov, line 276: In the kernel, `skb_headroom()` returns `unsigned int`, but this returns `ptrdiff_t`. If we want to be extra accurate, we should cast this here:

```suggestion
    return cast("unsigned int", skb.data - skb.head)
```

drgn/helpers/linux/net.py
```diff
@@ -260,3 +266,88 @@ def skb_shinfo(skb: Object) -> Object:
...
+
+
+def skb_is_nonlinear(skb: Object) -> Object:
+    return skb.data_len
+
+
+def skb_tailroom(skb: Object) -> IntegerLike:
+    if skb_is_nonlinear(skb):
+        return 0
+    return skb.end - skb.tail
```
osandov, line 286: Similarly, this one is `int` in the kernel, so it could be:

```suggestion
def skb_tailroom(skb: Object) -> Object:
    if skb_is_nonlinear(skb):
        return Object(prog, "int", 0)
    return cast("int", skb.end - skb.tail)
```

There doesn't seem to be much logic in how the return types were picked in the kernel, but we might as well copy those choices for the rest, too.

drgn/helpers/linux/net.py
```diff
@@ -260,3 +266,88 @@ def skb_shinfo(skb: Object) -> Object:
...
+    has_mac = skb_mac_header_was_set(skb)
+    has_trans = skb_transport_header_was_set(skb)
+
+    print(
+        "skb len=%u headroom=%u headlen=%u tailroom=%u\n"
+        "mac=(%d,%d) net=(%d,%d) trans=%d\n"
+        "shinfo(txflags=%u nr_frags=%u gso(size=%hu type=%u segs=%hu))\n"
+        "csum(0x%x ip_summed=%u complete_sw=%u valid=%u level=%u)\n"
+        "hash(0x%x sw=%u l4=%u) proto=0x%04x pkttype=%u iif=%d\n"
+        % (
```
osandov, line 326: I'd prefer an f-string or at least `str.format()` over `%` formatting.

osandov (changes requested):
Thanks for these! A few comments here, and a couple of more generic requests in line with [`CONTRIBUTING.rst`](https://github.com/osandov/drgn/blob/main/CONTRIBUTING.rst#linux-kernel-helpers):

- Could you please add docstrings to these helpers?
- Could you please add test cases? The test I added for `skb_shinfo()` in 470b2e02de32cac564c198378747f1d84e28e3d8 might be good inspiration. At the very least we should call them from a test case without looking at the result to make sure they don't blow up, but bonus points if you can craft SKBs to meaningfully test these.

PR #354, opened 2023-09-05
Add more pre-commit hooks

brenns10:
The "pre-commit-hooks" repo contains a few hooks that enforce some nice, common sense guidelines. The "vermin" hook checks compatibility with a supplied minimum Python version requirement, both checking the syntax and standard library usage. It's very helpful for identifying incompatible code before getting to the CI steps.

---

I don't believe I proposed these hooks in my initial pull request, but I've been using them in some of my other projects and find them quite helpful.

The pre-commit-hooks are a bunch of fast, common-sense checks that I've never really had cause issues, but they can catch some obvious things. Some of them would overlap with Black for the Python code, but they will apply to non-python files as well.

The vermin hook is the main reason I submitted this. It has prevented me from breaking my compatibility promises several times. It does occasionally require some `# novermin` comments, but it's super nice to have a fairly reliable way to verify that you're using stdlib APIs which are compatible down to 3.6. There are a few vmtest scripts which actually aren't but I don't think there was really any expectation that the scripts which manage kernels, compilers, and root filesystems, etc. are going to hold to the same compatibility guarantees as the rest of the library.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,23 +1,44 @@
+exclude: ^contrib/
 repos:
 -   repo: https://github.com/pycqa/isort
     rev: 5.12.0
     hooks:
     -   id: isort
         name: isort (python)
-        exclude: ^contrib/
 -   repo: https://github.com/psf/black
     rev: 23.7.0
     hooks:
     -   id: black
-        exclude: ^contrib/
 -   repo: https://github.com/pycqa/flake8
     rev: 6.0.0
     hooks:
     -   id: flake8
-        exclude: ^contrib/
 -   repo: https://github.com/pre-commit/mirrors-mypy
     rev: v0.931
     hooks:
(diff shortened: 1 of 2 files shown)


osandov (changes requested):
These all seem reasonable. My only request is that we exclude `contrib` from all of these checks -- I'd still prefer for the only requirements for `contrib` to be the bare minimum legal ones.

brenns10:
Would you like me to exclude `contrib` from the existing Black, mypy, isort, and flake8 checks as well?

osandov:
FYI, I added `rootfsbuild.py` back to `vermin` in https://github.com/osandov/drgn/commit/9d7f70aeaa08578566c77c0d068b77c168eee4ae, and opened netromdk/vermin#221 to see if it'd be possible to get rid of the `novermin` comments.

PR #355, opened 2023-09-11
Improvements to opening /proc/kcore via sudo

osandov:
A couple of followups to #347:

- Check whether sudo is installed in the first place.
- Make the sudo prompt include why we're asking for sudo.

### DIFF
--- a/drgn/internal/sudohelper.py
+++ b/drgn/internal/sudohelper.py
@@ -27,6 +27,8 @@ def open_via_sudo(
             subprocess.check_call(
                 [
                     "sudo",
+                    "-p",
+                    f"[sudo] password for %p to open {path}: ",
                     sys.executable,
                     "-B",
                     __file__,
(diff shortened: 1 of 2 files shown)


osandov:
@brenns10 mind taking a look at a couple of usability suggestions for the sudo opener?

drgn/internal/sudohelper.py
```diff
@@ -27,6 +27,8 @@ def open_via_sudo(
             subprocess.check_call(
                 [
                     "sudo",
+                    "-p",
+                    f"[sudo] password for %p to open {path}: ",
```
brenns10, line 31: Today I learned about `sudo -p` :)  Turns out it's present even in OL7/RHEL7 so it's been around for a while and should work on most distros.

brenns10 (approved):
Looks good to me!

drgn/internal/sudohelper.py
osandov: I hadn't even considered that, but it looks like it goes way back to sudo-project/sudo@337897e2d83489b4427a90e3dba20fad28d3fe88 in 1995! So I think we're covered ;)

PR #359, opened 2023-10-17
contrib: Add a sample IPython shell with magic commands

brenns10:
Here's the sample IPython shell code, edited down to be standalone. The docstring should contain everything you need to know, and here's a little demo:

image

Sadly, the "ERROR: root:Invalid alias" thing can't be avoided, unless you fully disable logging, which I didn't want to do. But as you can see, you get a pretty standard Ipython shell, with the ability to run custom commands as %command. And, if the command is not found in the globals, then you can actually run it as "command" without the prefix. And IPython always gives you a range of other magic commands, as well as shell command access with `!`.

The toy codebase I stripped it from actually had two CLIs:

1. The IPython one shown here
2. The standard drgn CLI, but with an added global: `crash()`. The `crash()` global swapped into a new CLI, which would allow you to execute the commands shown here, or swap back to the drgn CLI. I would have added all of the fun bash-isms and piping features of crash, but I ended up coming to the conclusion that reimplementing Crash on top of drgn didn't seem like a great use of my time.

That's why I didn't just write the commands as IPython magic class / magic functions. Also because I don't like the IPython magic API, but that's a separate story.

### DIFF
--- a/contrib/ipyshell.py
+++ b/contrib/ipyshell.py
@@ -1,0 +1,288 @@
+#!/usr/bin/env python3
+"""
+A sample IPython-based system for running Drgn with magic commands.
+
+You'll need IPython installed:
+
+    pip install ipython
+
+Make sure to run this from the root of the repository:
+
+    python -m contrib.ipyshell -c /proc/kcore -s vmlinux
+
+Once launched, you're in an IPython shell with the usual prog object and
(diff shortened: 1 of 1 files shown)


brenns10:
Cleaning up: #367 is better than this, closing.

PR #363, opened 2023-10-26
Bitops improvements

brenns10:
In the spirit of (incrementally) reducing the verbosity of drgn, I noticed that the bitops APIs could use a bit of improvement.

```python
>>> b = prog["boot_cpu_data"]
>>> # x86_capability is an array of __u32 ...
>>> bitfield = cast("unsigned long *", b.x86_capability[0].address_of_())
>>> bits = list(for_each_set_bit(bitfield, 8 * sizeof(b.x86_capability)))
```

First, the documentation indicated that they take an `unsigned long *`, so I assumed I needed to cast my array to a pointer, which is a pretty verbose operation. But reading the code, I could see that it worked perfectly fine on arrays, and also on pointers/arrays to any unsigned integer type. (And later, I noticed even the tests use an arrray rather than pointer). So this PR updates that documentation and explicitly tests a wider range of integer types (just to be sure). That way future users would know they can provide the bitfield directly without the casting.

Second, it seems like sized array types are relatively common in the kernel for bit fields (e.g. `cpuinfo_x86.x86_capability` [link](https://elixir.bootlin.com/linux/latest/A/ident/cpuinfo_x86)), and for these, we can infer the size parameter. So this PR makes it optional to provide that parameter.

Now with the clearer documentation and optional parameter, it's a lot easier:

```python
>>> bits = list(for_each_set_bit(prog["boot_cpu_data"].x86_capability))
```


drgn/helpers/linux/bitops.py
```diff
@@ -24,7 +24,8 @@ def for_each_set_bit(bitmap: Object, size: IntegerLike) -> Iterator[int]:
     """
     Iterate over all set (one) bits in a bitmap.
 
-    :param bitmap: ``unsigned long *``
+    :param bitmap: pointer to, or array of, unsigned integer type,
```
osandov, line 29: I'm all for documenting that this also works with an array. I think I didn't originally because if you're expecting the usual C rules, you'd expect arrays to work transparently as pointers. But it doesn't hurt to make it explicit.

I am slightly nervous about adding the any unsigned integer part, because on big endian systems, the bit numbering actually depends on the type. E.g., bit 1 in an 8-bit `unsigned char` is bit 57 in a 64-bit `unsigned long`. (Fun fact, this was one of my first kernel bugs: torvalds/linux@2fe1d55134fce05c17ea118a2e37a4af771887bc.) The test cases didn't catch it because `MockProgramTestCase` always creates a little-endian program.

`x86_capability` is sketchy. This always works for it since it's obviously little endian-only. My concern is that if there is another bitmap doing the same `u32` shenanigans but also used on big endian systems, it'd be a bug to not cast it to `unsigned long *`. But maybe this doesn't matter in practice, since those shenanigans only make sense in little endian-only code anyways.

I'm only weakly opposed, so let me know what you think.

drgn/helpers/linux/bitops.py
```diff
@@ -20,15 +20,23 @@
...
+    bitmap: Object, size: Optional[IntegerLike] = None
+) -> Iterator[int]:
     """
     Iterate over all set (one) bits in a bitmap.
 
     :param bitmap: pointer to, or array of, unsigned integer type,
         EG: ``unsigned long *``
-    :param size: Size of *bitmap* in bits.
+    :param size: Size of *bitmap* in bits. When *bitmap* is a sized array type
+        (EG: ``unsigned long[2]``), this can be omitted.
```
osandov, line 32: It can be a bug to assume that the whole array is being used. See ee0b8efc303329bdb52f25c9acfef108f5c40ddc.

I still like this change for 99% of cases, so to hopefully avoid the buggy case, I'd change this wording to say that it defaults to the size of the array in bits, rather than saying the size can be omitted.

osandov (changes requested):
FWIW, this slightly less verbose version is equivalent to your cast:

```python
cast("unsigned long *", prog["boot_cpu_data"].x86_capability)
```

But I like these improvements anyways, modulo my endianness concerns in the comments.

drgn/helpers/linux/bitops.py
brenns10: Wow, this is definitely one of those "hindsight is 20/20" situations, it's obvious that big endian systems would be a problem but I 100% did not think of it.

If there were a quick and clean way to resolve the underlying integer type and verify whether it's an `unsigned long`, then I'd advocate for a throwing a warning in there, e.g. something like the following pseudo-code:

```python
if underlying_type(bitmap.type_) == "unsigned long" and not (bitmap.prog_.platform.flags & PlatformFlags.IS_LITTLE_ENDIAN):
    warnings.warn("...")
```

Of course, there could be about a dozen typedefs in between, so I wouldn't want to throw all that type wrangling and warning code in, especially given that `test_bit()` is intuitively supposed to be fast.

I agree that we shouldn't advertise this as "works with any unsigned integer type". What if I put a disclaimer in the module docstring, something like:

> The bitops helpers operate on arrays of, or pointers to `unsigned long`. For little-endian architectures, they can operate equivalently with any unsigned integer type, because bit number assignments will be the same with any alignment. However, for big-endian architectures, this does not hold true, and accessing bit fields using anything other than an `unsigned long` is dangerous.

I could keep the test code but guard it with a test of the program flags, so that the tests will pass in any future case with a big-endian system.

PR #366, opened 2023-11-06
libdrgn: accept flattened KDUMP format

ptesarik:
Support for the flattened KDUMP format was added in libkdumpfile-0.5.3.

### DIFF
--- a/libdrgn/program.c
+++ b/libdrgn/program.c
@@ -24,6 +24,7 @@
 #include "helpers.h"
 #include "io.h"
 #include "language.h"
+#include "log.h"
 #include "linux_kernel.h"
 #include "memory_reader.h"
 #include "minmax.h"
@@ -222,16 +223,17 @@ has_kdump_signature(struct drgn_program *prog, const char *path, bool *ret)
 	ssize_t r = pread_all(prog->core_fd, signature, sizeof(signature), 0);
 	if (r < 0)
 		return drgn_error_create_os("pread", errno, path);
+	*ret = false;
 	if (r >= FLATTENED_SIG_LEN
 	    && memcmp(signature, FLATTENED_SIGNATURE, FLATTENED_SIG_LEN) == 0) {
-		return drgn_error_create(DRGN_ERROR_INVALID_ARGUMENT,
-					 "the given file is in the makedumpfile flattened "
-					 "format, which drgn does not support; use "
-					 "'makedumpfile -R newfile <oldfile' to reassemble "
-					 "the file into a format drgn can read");
-	}
-	*ret = (r >= KDUMP_SIG_LEN
-	        && memcmp(signature, KDUMP_SIGNATURE, KDUMP_SIG_LEN) == 0);
+		drgn_log_warning(prog,
+				 "the given file is in the makedumpfile flattened "
+				 "format; if open fails or is too slow, reassemble "
+				 "it with 'makedumpfile -R newfile <oldfile'");
+		*ret = true;
+	} else if (r >= KDUMP_SIG_LEN
+		   && memcmp(signature, KDUMP_SIGNATURE, KDUMP_SIG_LEN) == 0)
+		*ret = true;
 	return NULL;
 }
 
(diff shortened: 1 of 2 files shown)


libdrgn/program.c
```diff
@@ -222,15 +222,9 @@ static struct drgn_error *has_kdump_signature(const char *path, int fd,
...
-					 "format, which drgn does not support; use "
-					 "'makedumpfile -R newfile <oldfile' to reassemble "
-					 "the file into a format drgn can read");
-	}
-	*ret = (r >= KDUMP_SIG_LEN
+	*ret = (r >= FLATTENED_SIG_LEN
+		&& memcmp(signature, FLATTENED_SIGNATURE, FLATTENED_SIG_LEN) == 0) ||
+	       (r >= KDUMP_SIG_LEN
 	        && memcmp(signature, KDUMP_SIGNATURE, KDUMP_SIG_LEN) == 0);
 	return NULL;
```
brenns10, line 229: This seems great, thanks for the work adding the support into libkdumpfile! This will definitely be a usability improvement, though I think for the larger vmcores, it will still be helpful to use makedumpfile to reassemble the flattened format.

I see that you added support in libkdumpfile 0.5.3. Does libkdumpfile have any runtime feature-detection or version detection? Obviously we can check at compile time, and that would work well enough for the PyPI wheels, since they're distributed with a bundled libkdumpfile.

But for distribution packages, it would be nice to be able to check what version we're actually using, and if we're using earlier versions, fall back to the error message (and maybe add the minimum libkdumpfile version into that message text).

The other comment I have, and I'm not sure that this will be universally agreed upon, is that I'd like Drgn to emit some sort of warning when opening a flattened vmcore. My experience with crash's flattened vmcore support has been good, but it does spend a while indexing larger vmcores at startup, and it doesn't bother to give any indication to the user that they could cut back on the startup time by using `makedumpfile -R` just once. I would like to avoid that with drgn, so users know they have the option to reassemble the vmcore.

ptesarik: …
> But for distribution packages, it would be nice to be able to check what version we're actually using, and if we're using earlier versions, fall back to the error message (and maybe add the minimum libkdumpfile version into that message text).

I am not a big fan of version checks, because a newer feature may be backported to an older version. I prefer the AFF approach: “Ask for forgiveness, not for permission.” I would just try to open the file and if that fails with `KDUMP_ERR_ NOTIMPL`, the flattened format is not supported.

> The other comment I have, and I'm not sure that this will be universally agreed upon, is that I'd like Drgn to emit some sort of warning when opening a flattened vmcore. My experience with crash's flattened vmcore support has been good, but it does spend a while indexing larger vmcores at startup, and it doesn't bother to give any indication to the user that they could cut back on the startup time by using `makedumpfile -R` just once. I would like to avoid that with drgn, so users know they have the option to reassemble the vmcore.

Agreed. The flattened file format is such that the whole file must be scanned first to know where to find data. For a flattened dump, the `file.description` attribute starts with `Flattened `. However, this may not be available early enough. IIUC you would like to emit a warning _before_ libkdumpfile spends a lot of time opening the file, so users can press CTRL+C, rearrange the file and run again. There are no progress callbacks in libkdumpfile, but since drgn checks the signature anyway, I can keep the message there.

brenns10: > I am not a big fan of version checks, because a newer feature may be backported to an older version. I prefer the AFF approach: “Ask for forgiveness, not for permission.” I would just try to open the file and if that fails with KDUMP_ERR_ NOTIMPL, the flattened format is not supported.

Agreed in principle! Half of the problems in my life are caused by backported commits :P

In practice many times you can get away with a version check, but I understand wanting to avoid it.

The way your v2 is implemented works for me. Essentially, when we open a flattened vmcore, log the message, and if it fails, the user has a good idea of why.

ptesarik:
Um. OK, let me add a commit.

brenns10 (approved):
Looks even nicer in the libkdumpfile-0.5.2 case! Thanks for the update.
image

osandov:
Thanks for adding support to libkdumpfile and for this change!

PR #367, opened 2023-11-09
Commands - dot-prefixed, stdlib only

brenns10:
This is another stab at the idea of commands in drgn.

Rather than using IPython like the first PR I shared, this one is based on the stdlib only. It integrates into the existing drgn CLI via a subclass of `InteractiveConsole`. Commands are simply a callable, since I didn't want to get to opinionated with a class-based solution, or even using argparse.

Commands in the CLI are prefixed by a `.` . No valid Python statement starts with that, so this shouldn't break any existing behavior. They are able to return results just like normal expressions, and they can access the locals too.

The `drgn.cli` module has a registry of commands, which can be added to using the `@command("name")` decorator. In order to ensure that user/third-party commands can be loaded, I've added a `drgn.command.v1` entry point. Drgn automatically loads modules from that entry point prior to starting the CLI. (This is notable from a security standpoint. However, if you're running drgn as root, you would have needed to install that entry point as root as well.)

Just like globals and everything else, commands can be customized in `run_interactive()`.

I've added three useful commands into the default set. These aren't the only ones I would be interested in adding, just a demonstration.

- `.x` - execute a script (basically a wrapper on execscript)
- `.let` - set a variable equal to a drgn object
- `.contrib` - run a contrib script by name

I also added support to the `ptdrgn.py` script because I use that constantly.

In any case, I hope that the code is pretty straightforward. I think this is the most simple and direct way I could come up with to do commands in drgn.


brenns10:
I fixed the Python 3.6-3.9 mypy failure which was present in the previous revision, but of course since I updated the pre-commit, you won't actually get a mypy run on 3.6 now. I don't have permissions to add the label to the PR, but it would be good to have CI run against all Python versions, just to verify. Of course I tested it locally on 3.9 and it worked.

I also rebased this on main. I even checked this in combination with the #364 changes and they don't conflict. With those changes, it's pretty impressive how much less verbose the command line is!

brenns10:
Very funny, I didn't realize that when Github Actions runs, it does an automatic merge with the `main` branch and then runs tests. I got failures from #364 because I guess the run happened in the brief window between the merge and the fix :)

brenns10:
Since you mentioned it Wednesday, I thought I'd update the command pull request here. If you wanted to just assemble your own system from the idea (overriding `InteractiveInterpreter.runsource()`) that's absolutely fine, but if you wanted these patches in some form, now they're at least rebased. I also updated the completer code so that it would tab-complete the command names, which is pretty convenient.

The actual commands themselves from this pull request aren't too important. Though I do really like the `let` command, I would be using it constantly.

I do think the major missing piece here is that, if we were going to have a command system, there should be a separate page in the documentation with a description of each command. Ideally it could be somehow generated from the docstring of each command function, but presented as something else. I'm not sure how that would work yet.

PR #371, opened 2023-11-27
Update pre-commit hooks, dropping support for development on 3.6-3.7

brenns10:
Reflecting a bit of our discussion at Plumbers, I think this won't be too controversial. I've gone ahead and updated the pre-commit hooks to their latest versions, and handled any breakages there (see commit message for details). Of course, this blitzes past several tool versions which drop support for Python 3.6. So I've updated the CI to skip pre-commit/mypy on 3.6 / 3.7, and I've updated the contributing guide to explicitly state a policy on Python versions: 3.6+ is supported for building, running, & testing, but development tooling requires a non-EOL Python version.

Let me know your feelings on this, I think it will help alleviate some, but not all, of our woes with Python 3.6 support.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,14 +45,21 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
           allow-prereleases: true
+      - name: Check Python version for pre-commit
+        # Only run pre-commit / mypy on upstream supported Python versions
+        run: |
+          if [[ "${{ matrix.python-version }}" =~ ^3\.([89]|[0-9][0-9])$ ]]; then
+              echo USE_PRE_COMMIT=1 >> $GITHUB_ENV
+          fi
       - name: Install dependencies
         run: |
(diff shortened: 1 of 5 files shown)


osandov:
This seems like a good compromise.

For #364, I'd like to use some typing features that are only available in mypy versions that no longer support 3.6 and 3.7. This would break drgn library users that are type checking on 3.6 or 3.7. If there even are any, I don't feel too bad about it. What do you think?

brenns10:
I obviously can't speak for all users, but I don't know of any use of type checkers on the Python 3.6/3.7 code. The docs we have for setting up development environments specify 3.8+ for the development python version. So nobody at Oracle is running mypy on those versions.

As far as I'm concerned if this helps your type annotation work, then that's even better.

osandov:
Merged, thanks! This will make the annotations for #364 SO much easier. I'm catching up after Thanksgiving and a couple of sick days, but I'll have an update for that soon.

PR #372, opened 2023-11-29
drgn.helpers.linux.mm: have cmdline() and environ() return None for k…

pcc:
…ernel tasks

Sometimes I need to enumerate all tasks in the system and print their command lines, among other information. This fails for kernel tasks (i.e. tasks without mm):

  File "<console>", line 2, in <module>
  File "[...]/drgn/helpers/linux/mm.py", line 1276, in cmdline
    arg_start = mm.arg_start.value_()
                ^^^^^^^^^^^^^^^^^^^^^
_drgn.FaultError: address is not mapped: 0x140

It seems reasonable for this not to cause an exception but rather to return a value that the caller can conveniently recover from (consider also the behavior of `cat /proc/2/cmdline`). Therefore, change cmdline() to return None for kernel tasks. Do the same for environ() for consistency.

### DIFF
--- a/drgn/helpers/linux/mm.py
+++ b/drgn/helpers/linux/mm.py
@@ -1320,9 +1320,9 @@ def access_remote_vm(mm: Object, address: IntegerLike, size: IntegerLike) -> byt
     return _linux_helper_read_vm(mm.prog_, mm.pgd, address, size)
 
 
-def cmdline(task: Object) -> List[bytes]:
+def cmdline(task: Object) -> Optional[List[bytes]]:
     """
-    Get the list of command line arguments of a task.
+    Get the list of command line arguments of a task, or ``None`` for kernel tasks.
 
     >>> cmdline(find_task(prog, 1495216))
     [b'vim', b'drgn/helpers/linux/mm.py']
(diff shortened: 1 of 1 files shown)


drgn/helpers/linux/mm.py
```diff
@@ -1335,14 +1335,16 @@ def cmdline(task: Object) -> List[bytes]:
     :param task: ``struct task_struct *``
     """
     mm = task.mm.read_()
+    if mm.value_() == 0:
```
osandov, line 1338: Slightly preferred:

```suggestion
    if not mm:
```

osandov (commented):
This seems reasonable. I'm debating whether I'd prefer returning an empty list rather than `None`. That would be more backwards compatible (e.g., type-checked scripts wouldn't start getting errors now that `cmdline()` can return `None`), and still in the spirit of `/proc/2/cmdline` and `/proc/2/environ` being empty files. I could be swayed either way.

pcc:
I also considered returning an empty list. `None` seemed more accurate though, since it is possible for a userspace task to have an empty environment or (before commit dcd46d897adb70d63e025f175a00a89797d31a43) an empty argv. My gut feeling is that Python type checking is unlikely to be used in the sort of ad-hoc scripts that use drgn, so I wouldn't expect this to be likely to break anything.

drgn/helpers/linux/mm.py
```diff
@@ -1320,9 +1320,9 @@ def access_remote_vm(mm: Object, address: IntegerLike, size: IntegerLike) -> byt
     return _linux_helper_read_vm(mm.prog_, mm.pgd, address, size)
 
 
-def cmdline(task: Object) -> List[bytes]:
+def cmdline(task: Object) -> Optional[List[bytes]]:
     """
-    Get the list of command line arguments of a task.
+    Get the list of command line arguments of a task, or None for kernel tasks.
```
osandov, line 1325: One more typographical thing:

```suggestion
    Get the list of command line arguments of a task, or ``None`` for kernel tasks.
```

osandov (changes requested):
Yeah, distinguishing between kernel threads and truly empty seems like a good enough reason for this, so please update with the nitpicks and I'll merge this, thanks!

PR #376, opened 2023-12-05
drgn.helpers.linux.stackdepot: add stack_depot_fetch() helper 

pcc:
This helper returns a stack trace from the stack depot.

### DIFF
--- a/drgn/helpers/linux/stackdepot.py
+++ b/drgn/helpers/linux/stackdepot.py
@@ -1,0 +1,47 @@
+# Copyright (c) Google LLC
+# SPDX-License-Identifier: LGPL-2.1-or-later
+
+"""
+Stack Depot
+-----------
+
+The ``drgn.helpers.linux.stackdepot`` module provides helpers for working with
+the stack depot implementation used by KASAN and other kernel debugging tools.
+"""
+
+from typing import Optional
+
+from drgn import Object, StackTrace, cast, reinterpret
+
+__all__ = ("stack_depot_fetch",)
+
+
+def stack_depot_fetch(handle: Object) -> Optional[StackTrace]:
+    """
+    Returns a stack trace for the given stack handle.
+
+    :param handle: ``depot_stack_handle_t``
+    :return: The stack trace, or ``None`` if not available.
+    """
+    prog = handle.prog_
+    handle_parts = reinterpret("union handle_parts", handle)
+
+    # Renamed in Linux kernel commit 961c949b012f ("lib/stackdepot: rename slab
+    # to pool") (in v6.3).
+    try:
+        pool = prog["stack_pools"][handle_parts.pool_index]
+    except KeyError:
+        pool = prog["stack_slabs"][handle_parts.slabindex]
+
+    if not pool:
+        return None
+
+    # This has remained the same since the stack depot was introduced in
+    # commit cd11016e5f52 ("mm, kasan: stackdepot implementation. Enable
+    # stackdepot for SLAB"), when it was known as STACK_ALLOC_ALIGN.
+    DEPOT_STACK_ALIGN = 4
+
+    record = cast(
+        "struct stack_record *", pool + (handle_parts.offset << DEPOT_STACK_ALIGN)
+    )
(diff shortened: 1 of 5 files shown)


pcc:
(I encountered #378 while trying to use this with a value type.)

libdrgn/stack_trace.c
```diff
@@ -1130,6 +1130,37 @@ drgn_program_stack_trace(struct drgn_program *prog, uint32_t tid,
...
+	size_t trace_capacity = pcs_size;
+
+	if (!trace)
+		return &drgn_enomem;
+
+	trace->prog = prog;
+	trace->num_frames = 0;
+	for (size_t i = 0; i != pcs_size; ++i) {
+		struct drgn_register_state *regs =
+			drgn_register_state_create_impl(0, 0, false);
```
osandov, line 1149: Setting `interrupted` to false means that this PC is the return address of some call, not the instruction itself that trapped or whatever. Often, the top frame is interrupted (e.g., in a core dump), but not always (e.g., for a scheduled out task). We probably want this interface to allow controlling the `interrupted` flag for the top frame.

Do you know whether stack frames in the stack depot store a return address or an interrupted address in the top frame?

drgn/helpers/linux/stackdepot.py
```diff
@@ -0,0 +1,47 @@
...
+    """
+    Returns a stack trace for the given stack handle.
+
+    :param handle: ``depot_stack_handle_t``
+    :return: The stack trace, or ``None`` if not available.
+    """
+    prog = handle.prog_
+    handle_parts = reinterpret("union handle_parts", handle)
+
+    # Renamed in commit 961c949b012f ("lib/stackdepot: rename slab to pool").
```
osandov, line 29: [This should include](https://github.com/osandov/drgn/blob/main/CONTRIBUTING.rst#linux-kernel-helpers) the kernel version, too.
```suggestion
    # Renamed in Linux kernel commit 961c949b012f ("lib/stackdepot: rename slab
    # to pool") (in v6.3).
```

drgn/helpers/linux/stackdepot.py
```diff
@@ -0,0 +1,47 @@
...
+    # This has remained the same since the stack depot was introduced in
+    # commit cd11016e5f52 ("mm, kasan: stackdepot implementation. Enable
+    # stackdepot for SLAB"), when it was known as STACK_ALLOC_ALIGN.
+    DEPOT_STACK_ALIGN = 4
+
+    record = cast(
+        "struct stack_record *", pool + (handle_parts.offset << DEPOT_STACK_ALIGN)
+    )
+    trace = list(map(lambda x: record.entries[x], range(record.size)))
+    return prog.stack_trace_from_pcs(trace)
```
osandov, line 47: A list comprehension would be more idiomatic here:
```suggestion
    return prog.stack_trace_from_pcs([record.entries[x] for x in range(record.size)])
```

osandov (changes requested):
Happy new year! I didn't know about the stack depot, so this is cool. Some comments below.

P.S. I also like this as a first step towards an interface for custom stack unwinders that could set any registers.

libdrgn/stack_trace.c
pcc: > Do you know whether stack frames in the stack depot store a return address or an interrupted address in the top frame?

At least for the current users of stack depot I think they're all return addresses. E.g. almost every caller of `stack_depot_save` or `__stack_depot_save` calls `stack_trace_save` to collect the stack trace (the exception is KMSAN which uses the stack depot to store partial stack traces in a non-standard format), which calls `arch_stack_walk` with a null `pt_regs` argument, which only ever passes return addresses to the callback (at least the arm64 and x86 implementations do).

> We probably want this interface to allow controlling the interrupted flag for the top frame.

Is it worth doing that now? Personally I'd wait until there is a user.

osandov: Fair enough, it'll be easy enough to add later.

osandov (approved):
Thanks, this looks great.

osandov:
I just pushed a couple of test cases for this, too.

PR #379, opened 2023-12-06
x86_64: mm: support numa migration pte entry

### LINKED ISSUE #370: Can't access user address
Here is a vma extracted from the smaps file. It is evident that this particular VMA contains 1942012kB of anonymous memory.

```
700000000-8003b7000 rw-p 00000000 00:00 0
Size:            4198108 kB
KernelPageSize:        4 kB
MMUPageSize:           4 kB
Rss:             1942012 kB
Pss:             1942012 kB
Shared_Clean:          0 kB
Shared_Dirty:          0 kB
Private_Clean:         0 kB
Private_Dirty:   1942012 kB
Referenced:      1942012 kB
Anonymous:       1942012 kB
LazyFree:              0 kB
AnonHugePages:         0 kB
ShmemPmdMapped:        0 kB
Shared_Hugetlb:        0 kB
Private_Hugetlb:       0 kB
Swap:                  0 kB
SwapPss:               0 kB
Locked:                0 kB
ProtectionKey:         0
VmFlags: rd wr mr mw me ac sd
```

Now, using drgn 0.0.24, we can scan each address within this VMA and count the number of addresses that can be successfully accessed.

```
task = find_task(prog, 2023533)
addr = 0x700000000
count = 0
while addr < 0x8003b7000:
	try:
		data = access_process_vm(task, addr, 4096)
		count += 1
	except:
		#print(hex(addr))
		pass
	addr += 4096

print(count)
print(count * 4096 / 1024/1024)
```

The result indicates a count of 20436, and the total accessible size is 79.828125MB, which is significantly smaller than the actual size.

I don't know how to solve this...

hao-lee:
The PRESENT bit alone does not indicate whether a page actually exists. The PRESENT bit is also cleared for numa migration pte entries, even though the page is present. Therefore, to determine if a page truly exists, we need to check pte and convert it to phys in the same way the kernel does.

Fix #370

### DIFF
--- a/libdrgn/arch_x86_64.c
+++ b/libdrgn/arch_x86_64.c
@@ -574,18 +574,49 @@ linux_kernel_pgtable_iterator_init_x86_64(struct drgn_program *prog,
 	memset(it->index, 0xff, sizeof(it->index));
 }
 
+// pte flags
+static const uint64_t _PAGE_PRESENT = 1 << 0;
+static const uint64_t _PAGE_PROTNONE = 1 << 8;
+
+// TODO: PTE_PFN_MASK may be different because of CONFIG_DYNAMIC_PHYSICAL_MASK
+static const uint64_t PTE_PFN_MASK = UINT64_C(0xffffffffff000);
+static const int PAGE_SHIFT = 12;
+
+static inline bool __pte_needs_invert(uint64_t val)
+{
+	return val && !(val & _PAGE_PRESENT);
+}
+
+static inline uint64_t protnone_mask(uint64_t val)
+{
+	return __pte_needs_invert(val) ?  ~0ull : 0;
+}
+
+static inline uint64_t pte_flags(uint64_t pte)
+{
+	return pte & ((1 << PAGE_SHIFT) - 1);
+}
+
+static inline int pte_present(uint64_t pte)
+{
+	return pte_flags(pte) & (_PAGE_PRESENT | _PAGE_PROTNONE);
+}
+
+static inline uint64_t pte_phys(uint64_t pte)
+{
+	pte ^= protnone_mask(pte);
+	return pte & PTE_PFN_MASK;
+}
+
 static struct drgn_error *
 linux_kernel_pgtable_iterator_next_x86_64(struct drgn_program *prog,
 					  struct pgtable_iterator *_it,
 					  uint64_t *virt_addr_ret,
 					  uint64_t *phys_addr_ret)
 {
-	static const int PAGE_SHIFT = 12;
 	static const int PGTABLE_SHIFT = 9;
 	static const int PGTABLE_MASK = (1 << PGTABLE_SHIFT) - 1;
-	static const uint64_t PRESENT = 0x1;
 	static const uint64_t PSE = 0x80; // a.k.a. huge page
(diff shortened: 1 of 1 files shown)


libdrgn/arch_x86_64.c
```diff
@@ -623,14 +654,14 @@ linux_kernel_pgtable_iterator_next_x86_64(struct drgn_program *prog,
 			uint64_t entry = it->table[level][it->index[level]++];
 			if (bswap)
 				entry = bswap_64(entry);
-			table = entry & ADDRESS_MASK;
-			if (!(entry & PRESENT) || (entry & PSE) || level == 0) {
+			table = entry & PTE_PFN_MASK;
+			if (!(entry & _PAGE_PRESENT) || (entry & PSE) || level == 0) {
```
osandov, line 658: Can huge pages also have numa migration entries? `pmd_present()` in the kernel checks `_PAGE_PROTNONE`, so maybe 2MB huge pages can? If so, this check probably needs to be slightly different (and depend on the level).

libdrgn/arch_x86_64.c
```diff
@@ -623,14 +654,14 @@ linux_kernel_pgtable_iterator_next_x86_64(struct drgn_program *prog,
...
 				uint64_t mask = (UINT64_C(1) <<
 						 (PAGE_SHIFT +
 						  PGTABLE_SHIFT * level)) - 1;
 				*virt_addr_ret = virt_addr & ~mask;
-				if (entry & PRESENT)
-					*phys_addr_ret = table & ~mask;
+				if (pte_present(entry))
+					*phys_addr_ret = pte_phys(entry);
```
osandov, line 664: This will also get entries for higher levels (for huge pages or unmapped levels), so the `pte_*` names are not completely accurate. The functions themselves probably also need to depend on the level as noted above.

libdrgn/arch_x86_64.c
```diff
@@ -574,18 +574,49 @@ linux_kernel_pgtable_iterator_init_x86_64(struct drgn_program *prog,
...
+static const int PAGE_SHIFT = 12;
+
+static inline bool __pte_needs_invert(uint64_t val)
+{
+	return val && !(val & _PAGE_PRESENT);
+}
+
+static inline uint64_t protnone_mask(uint64_t val)
+{
+	return __pte_needs_invert(val) ?  ~0ull : 0;
```
osandov, line 592: In theory, `~0ull` could be a different size than `uint64_t`, so I'd feel better about:

```suggestion
	return __pte_needs_invert(val) ? -1 : 0;
```

libdrgn/arch_x86_64.c
```diff
@@ -574,18 +574,49 @@ linux_kernel_pgtable_iterator_init_x86_64(struct drgn_program *prog,
...
+}
+
+static inline uint64_t pte_flags(uint64_t pte)
+{
+	return pte & ((1 << PAGE_SHIFT) - 1);
+}
```
osandov, line 598: I'd prefer to fold these small helpers into their callers.

osandov (changes requested):
This was a great find, thank you! I left a few comments.

PR #383, opened 2024-02-19
helpers: linux: fs: use mount rbtree instead of list

morbidrsa:
Starting with kernel v6.8 the mnt_list int struct mnt_namespace git converted to a rb-tree.

Reflect this change in drgn's for_each_mount() helper.

### DIFF
--- a/drgn/helpers/linux/fs.py
+++ b/drgn/helpers/linux/fs.py
@@ -21,6 +21,7 @@ from drgn.helpers.linux.list import (
     list_for_each_entry,
     list_for_each_entry_reverse,
 )
+from drgn.helpers.linux.rbtree import rbtree_inorder_for_each_entry
 
 __all__ = (
     "path_lookup",
@@ -310,7 +311,13 @@ def for_each_mount(
         dst = os.fsencode(dst)
     if fstype:
         fstype = os.fsencode(fstype)
-    for mnt in list_for_each_entry("struct mount", ns.list.address_of_(), "mnt_list"):
+    try:
+        mounts = list_for_each_entry("struct mount", ns.list.address_of_(), "mnt_list")
+    except AttributeError:
+        mounts = rbtree_inorder_for_each_entry(
+            "struct mount", ns.mounts.address_of_(), "mnt_node"
+        )
+    for mnt in mounts:
         if (
             (src is None or mount_src(mnt) == src)
             and (dst is None or mount_dst(mnt) == dst)


morbidrsa:
Sure will send an update soon

osandov:
Here's another example I just had to do: https://github.com/osandov/drgn/commit/718da449efc2d09a9ebf9748a30b1f80d29cc200. Let me know if you need any more pointers or if you'd like me to take it from here.

morbidrsa:
> Here's another example I just had to do: [718da44](https://github.com/osandov/drgn/commit/718da449efc2d09a9ebf9748a30b1f80d29cc200). Let me know if you need any more pointers or if you'd like me to take it from here.

Hi Omar, no there's no problem, I just didn't have time to do the update. I'm stuck debugging a btrfs issue.

drgn/helpers/linux/fs.py
```diff
@@ -310,7 +311,13 @@ def for_each_mount(
         dst = os.fsencode(dst)
     if fstype:
         fstype = os.fsencode(fstype)
-    for mnt in list_for_each_entry("struct mount", ns.list.address_of_(), "mnt_list"):
+    try:
+        mounts = list_for_each_entry("struct mount", ns.list.address_of_(), "mnt_list")
+    except AttributeError:
+        mounts = rbtree_inorder_for_each_entry(
+            "struct mount", ns.mounts.address_of_(), "mnt_node"
+        )
```
osandov, line 319: Just a couple of nits documented in `CONTRIBUTING.rst`:

1. Add a comment documenting the version divergence.
2. Put the case for the latest kernel first.

```suggestion
    # Since Linux kernel commit 2eea9ce4310d ("mounts: keep list of mounts in
    # an rbtree") (in v6.8), the mounts in a namespace are in a red-black tree.
    # Before that, they're in a list.
    try:
        mounts = rbtree_inorder_for_each_entry(
            "struct mount", ns.mounts.address_of_(), "mnt_node"
        )
    except AttributeError:
        mounts = list_for_each_entry("struct mount", ns.list.address_of_(), "mnt_list")
```

osandov (changes requested):
Thanks! Just a couple of nits.

drgn/helpers/linux/fs.py
morbidrsa: Unfortunately that made the CI unhappy when I tried it.

osandov: Ah, I see, `struct mnt_namespace` previously had a member named `mounts` which was the number of mounts, so the `AttributeError` doesn't trigger in the `try`. Okay, let's just add the comment.

osandov (approved):
I have a related fix for path_lookup(), so I'll add the comment on my end. Thanks!

Issue #324, opened 2023-07-05
Add APIs for writing to memory and variables


osandov:

brenns10:
Discussion from LPC 2023 related to writing to memory, and breakpoints:
https://lpc.events/event/17/contributions/1471/
https://www.youtube.com/watch?v=6jRi44yp3YQ

PR #384, opened 2024-02-25
Add write support to Program

Tim---:
Hi,

I'm trying to add support for writing memory related to #324.

For now, the changes for the Python API are an optional `write_fn` parameter to `Program.add_memory_segment`, and the method `Program.write`.
Before I try to tackle the missing pieces (writable objects, and optional write support in the Program constructors), I wanted to make sure I'm on the right track.

Note that on the C-side, instead of adding `*_write` functions as counterparts to the `*_read` functions, I tried to maintain as much as possible a unique `*_rw(bool is_write, void *buf, size_t count)` kind of functions. 
This has the advantage of reusing the internal logic of some functions for both cases (eg: `drgn_program_read_rw`).
However, I'm now seeing that maybe the buf parameter should be `const void *` for the write functions, or I need to make some dirty casts :/.

Also, the naming choices are maybe not the best (`memory_interface` is meh, and if we keep the `*_rw` functions, maybe `*_transfer` would be prettier).

Anyway, feel free to review the changes and tell me if I made some mistakes, I'll need to rebase it anyway to add the Signed-Off-By that I forgot in the commits :).

### DIFF
--- a/_drgn.pyi
+++ b/_drgn.pyi
@@ -402,12 +402,33 @@ class Program:
         :raises FaultError: if the address is invalid; see :meth:`read()`
         """
         ...
+    def write(
+        self, address: IntegerLike, data: bytes, physical: bool = False
+    ) -> bytes:
+        """
+        Writes *data* starting at *address* in the program memory. The
+        address may be virtual (the default) or physical if the program
+        supports it.
+
+        >>> prog.write(0xffffffffbe012b40, b'swapper/0\x00\x00\x00\x00\x00\x00\x00')
+
+        :param address: The starting address.
+        :param data: The bytes to write.
+        :param physical: Whether *address* is a physical memory address. If
+            ``False``, then it is a virtual memory address. Physical memory can
+            usually only be read when the program is an operating system
+            kernel.
+        :raises FaultError: if the address range is invalid or the type of
+            address (physical or virtual) is not supported by the program
+        """
+        ...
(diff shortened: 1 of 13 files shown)


Tim---:
I tried to go a bit further and started implementing `__setattr__` and `__setitem__` methods for the `Object` class.
For now I just try to implement it for "value" objects, but I seem to have reached a rather annoying problem:

If I assign `point.x = 123`, I can read `point.x` with the new value. But if I so the same with a nested structure like `segment.a.x = 123`, the value is not changed because `segment.a` seems to return a copy of the point object.

Do you have an idea of how this could be handled for writing objects ?

osandov:
Hi, sorry for the ridiculously late response. I recently added a couple of experimental helpers for writing to kernel memory/objects: https://drgn.readthedocs.io/en/latest/helpers.html#drgn.helpers.experimental.kmodify.write_memory. That renewed my interest in having this more generic interface, since it would be possible to plug the `write_memory()` helper into a memory writer. Let me know if this is still something that you're interested in and I can give this PR a look.

> For now I just try to implement it for "value" objects, but I seem to have reached a rather annoying problem

I have conflicting feelings about allowing writes to value objects in general. In the Python API, Objects are currently immutable. Allowing writes could break some expectations there. But, it is potentially useful. Did you have any particular use cases in mind?

Issue #377, opened 2023-12-06
More convenient API for reading an array


pcc:
Sometimes I need to read all elements of a dynamically sized array. This can be slow, especially with a high latency remote debugging connection to the target. So I would really like an API for reading all elements of the array at once, so that I can read the array with a small number of round trips to the target. I came up with this workaround (for my current use case, which is reading the KASAN ring buffer):
```
cast('struct kasan_stack_ring_entry(*) [%d]' % prog['stack_ring'].size, prog['stack_ring'].entries)[0].read_()
```
But it would be nice if there were a better way to do this. One possibility might be a `read_array_()` method for objects of pointer type that reads a specified number of elements via the pointer and returns the array.

PR #388, opened 2024-03-18
Add SymbolIndex and kallsyms helpers

brenns10:
This is the next step of the DWARFless debugging / CTF feature, #176. Unlike the symbol finder, there aren't _too many_ questions of API, instead we're just implementing the API. This PR adds:

* The `SymbolIndex` type, which holds an immutable list of symbols and indexes them for efficient name and address based lookup. Some interesting items here:
    * It uses a hash for name lookup, and an interesting binary search approach for address lookup, in order to support duplicated names and overlapping address ranges.
    * I chose `SymbolIndex` over `SymbolTable` since I didn't want to overload the common usage of that term applying to ELF symbol tables. That said, `SymbolIndex` is a bit strange so I'm open to suggestions.
* Two C helpers (`load_proc_kallsyms()` and `load_builtin_kallsyms()`) enumerate all kallsyms symbols and return a `SymbolIndex` object. The first uses `/proc/kallsyms` and the second uses information from `VMCOREINFO` to find the kallsyms tables.
    * The proc version lets you specify a filename, in case you happen to have a copy of `/proc/kallsyms` from a vmcore.
    * `/proc/kallsyms` includes module symbols, while the built-in vmlinux kallsyms of course does not. I've included an option to parse these symbols, however by default, we do not. This is because you can actually get the ELF symbol objects, which have more information, directly from the `struct module`.
* A Python helper (`load_module_kallsyms()`) for loading module symbols from the kernel image and creating a symbol index from that. This does require debuginfo for vmlinux.

I also of course have tests for the above.

Below are two issues you'll probably notice and want to discuss:

1. The `guess_long_names()` function works around torvalds/linux@73bbb94466fd ("kallsyms: support "big" kernel symbols") which was merged in 6.1. I didn't catch it immediately so I'm not sure whether there's any use in sending a commit to include a version number for kallsyms... Especially since I believe it's unlikely to be backported, but I could be wrong. It has certainly not been backported to stable kernels.
2. The final commit of the branch adds an optimization where we can mark symbols as having a lifetime. Symbols with static lifetimes are not freed by `drgn_symbol_destroy()`. This enables the `SymbolIndex` to provides symbols with zero copying, even all the way out to the Python layer! However, I'm not certain it's worth the complexity it adds. That's why I kept the change as a second commit; it's easy to drop if you don't like it. I haven't measured any performance impact and admittedly, I don't know if it would be noticeable.


brenns10:
Alright, sorry for the failure (it was an old commit changing how ELF symbols are stored in `struct mod_kallsyms`). I had been hoping for a clean bill of health on the first try :D
But it's ready now, excited to take the new symbol finder API out for a spin :D

osandov (commented):
I've only looked at the first two commits so far, so feel free to ignore these comments until I'm done or address them now, up to you. I'll continue reviewing tomorrow.

libdrgn/kallsyms.c
```diff
@@ -0,0 +1,625 @@
...
+	}
+
+	/*
+	 * Find the end of the last token, so we get the overall length of
+	 * token_table. Then copy the token_table into host memory.
+	 */
+	last_token = loc->kallsyms_token_table + kr->token_index[UINT8_MAX];
+	do {
+		err = drgn_program_read_u8(prog, last_token, false,
+					   (uint8_t *)&data);
```
osandov, line 167: Use `drgn_program_read_memory()` so you don't need to cast the pointer.

libdrgn/kallsyms.c
```diff
@@ -0,0 +1,625 @@
...
+		major *= 10;
+		major += osrelease[i] - '0';
+	}
+	for (i = i + 1; i < sizeof(prog->vmcoreinfo.osrelease) && osrelease[i] && osrelease[i] != '.'; i++) {
+		char c = osrelease[i];
+		if (c < '0' || c > '9')
+			break;
+		minor *= 10;
+		minor += osrelease[i] - '0';
+	}
```
osandov, line 109: The version number parsing in `orc_version_from_osrelease()` is a little more concise, please copy that. (I debated turning it into a helper function for 3e7244d25fd12a2d497f40fdde2e8e9157d4e1a1, but I decided that I'd prefer to keep it painful to do version checks to discourage it whenever possible.)

libdrgn/kallsyms.c
```diff
@@ -0,0 +1,625 @@
...
+ * near the beginning of the address space, and a large gap before normal kernel
+ * symbols.  The result of this is that we can create symbols with incredibly
+ * large sizes, and then drgn's symbolization will print addresses using that
+ * symbol and a very large offset, which is absolutely meaningless.
+ *
+ * To avoid this, we set a cap on the length of a symbol. Unfortunately, this is
+ * a heuristic. It's entirely possible to have very large data symbols. This
+ * value is chosen somewhat arbitrarily, but seems to produce decent results.
+ */
+#define MAX_SYMBOL_LENGTH 0x10000
```
osandov, line 60: FWIW, here are the top 10 symbol sizes in a vmtest kernel, a few of which are over this threshold:

```
$ eu-readelf -s build/vmtest/x86_64/kernel-6.11.0-rc2-vmtest30.1default/build/vmlinux | awk '{ print $3 "\t" $8 }' | sort -rn | head
360448  _printk_rb_static_infos
262144  early_dynamic_pgts
131072  __log_buf
98304   _printk_rb_static_descs
65536   stack_pools
65536   __brk_early_pgt_alloc
65536   __brk_dmi_alloc
65536   __apicid_to_node
49536   i8042_dmi_quirk_table
32768   ftrace_stacks
```

I have no idea how useful those symbols are.

libdrgn/kallsyms.c
```diff
@@ -0,0 +1,625 @@
...
+	if (!addresses)
+		return &drgn_enomem;
+
+	if (loc->kallsyms_addresses) {
+		/*
+		 * The kallsyms addresses are stored as plain addresses in an
+		 * array of unsigned long! Read the appropriate size array and
+		 * do any necessary byte swaps.
+		 */
+		if (!bits64) {
```
osandov, line 372:
```
if (!foo)
else
```

breaks my brain, please swap it to
```
if (bits64)
else
```

libdrgn/kallsyms.c
```diff
@@ -0,0 +1,625 @@
...
+			for (int i = 0; i < kr->num_syms; i++) {
+				if (bswap)
+					addresses[i] = bswap_32(addr32[i]);
+				else
+					addresses[i] = addr32[i];
+			}
+		} else {
+			err = drgn_program_read_memory(prog, addresses,
+						loc->kallsyms_addresses,
+						kr->num_syms * sizeof(uint32_t),
```
osandov, line 392: Is this supposed to be `sizeof(uint64_t)`?

libdrgn/kallsyms.c
```diff
@@ -0,0 +1,625 @@
...
+		 * interpret them.
+		 */
+		uint64_t relative_base;
+		if (bits64) {
+			err = drgn_program_read_u64(prog, loc->kallsyms_relative_base,
+						false, &relative_base);
+			if (err)
+				return err;
+			if (bswap)
+				relative_base = bswap_64(relative_base);
```
osandov, line 419: `drgn_program_read_u64()` already does the byte swap.
```suggestion
```

libdrgn/kallsyms.c
```diff
@@ -0,0 +1,625 @@
...
+		} else {
+			uint32_t rel32;
+			err = drgn_program_read_u32(prog, loc->kallsyms_relative_base,
+						    false, &rel32);
+			if (err)
+				return err;
+			if (bswap)
+				rel32 = bswap_32(rel32);
```
osandov, line 427: Ditto.
```suggestion
```

libdrgn/kallsyms.c
```diff
@@ -0,0 +1,625 @@
...
+	while (len) {
+		char *token_ptr = &kr->token_table[kr->token_index[*data]];
+		while (*token_ptr) {
+			if (skipped_first) {
+				if (!string_builder_appendc(sb, *token_ptr)) {
+					string_builder_null_terminate(sb);
+					return offset;
+				}
+			} else {
+				if (kind_ret)
```
osandov, line 250: It doesn't seem like you ever pass NULL, right? If so, let's just remove this check.

libdrgn/kallsyms.c
```diff
@@ -0,0 +1,625 @@
...
+					*kind_ret = *token_ptr;
```
osandov, line 251: Can we return without ever setting this, leaving the caller with uninitialized garbage?

osandov (changes requested):
Okay, I finally made it through the whole series. Good stuff! I can't claim that I fully understood the kallsyms format this time around, but I at least tried to look for coding errors.

The guts of the implementation look good, but the `drgn.helpers.linux.kallsyms` interface is a bit overwhelming. There are lots of similar-looking helpers that I wouldn't know how to make use as an end user. I wonder if we can pare it down a bit. What are the most important use cases that you're targeting?

brenns10:
Thanks for the review! I'll take a look through it in detail and try to get an update ASAP, rebased on main as well. It looks like you made a lot of good catches, so thanks for that.

> The guts of the implementation look good, but the drgn.helpers.linux.kallsyms interface is a bit overwhelming. There are lots of similar-looking helpers that I wouldn't know how to make use as an end user. I wonder if we can pare it down a bit. What are the most important use cases that you're targeting?

Yeah, I have even confused myself... I've occasionally imported from there and not known which function I want to use. So it does need to be simplified.

The primary use cases I can think of are:

1. The full non-DWARF approach where you use kallsyms, CTF || BTF, and ORC or frame pointers for unwinding. In this case, you need to do a delicate dance: first load vmlinux kallsyms, then the types (you need the symbol table to find the BTF, though CTF doesn't require the symbol table first), then the module kallsyms (since module kallsyms requires vmlinux types). In the final CTF branch, there's a simple `drgn.helpers.linux.ctf:load_ctf()` function that just does everything for you in the right order, so most of the time nobody would need to think about it.
2. The partial DWARF approach, where you have vmlinux DWARF loaded, but you didn't extract all the modules. This actually happens all the time for us internally, because we typically extract only the debuginfo we need. In that case, it's very nice to be able to add a symbol finder for the kernel modules, so that symbolization works as expected without loading the module debuginfo.

That would mean at a minimum, we'd really just need `load_vmlinux_kallsyms()` and `load_module_kallsyms()`. I don't think the `load_builtin_kallsyms()` and `load_proc_kallsyms()` functions are really necessary to be part of the API (though keeping them as private/internal functions would be nice, because I could see myself wanting to use them hackily).

The `module_kallsyms()` function is exceptionally poorly named. I included this as a public function because I could imagine a world where a user may have some kernel modules' DWARF debuginfo loaded, but they want to load the other modules' symbols. They could use this function to get that subset of symbols. Or, alternatively, I could see it being useful for a user to simply ask: what symbols are part of this kernel module? But to sum up, I don't have a single great use case for this function. I wouldn't be disappointed to scrap it from the API, or at a minimum, it should be renamed to something clearer (e.g. `get_module_symbols()`).

libdrgn/symbol.c
```diff
@@ -174,3 +178,291 @@ void drgn_symbol_result_builder_array(struct drgn_symbol_result_builder *builder
...
+		}
+
+		entry.key = current_str;
+		entry.value.start = current;
+		entry.value.end = next;
+		if (drgn_symbol_name_table_insert(&ret->htab, &entry, NULL) < 0) {
+			free(ret->name_sort);
+			free(ret->max_addrs);
+			drgn_symbol_name_table_deinit(&ret->htab);
+			ret->name_sort = NULL;
```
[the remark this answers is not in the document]
brenns10: It does not! I think I just occasionally get paranoid and set freed pointers to null (but... not all the time, as evidenced here). I'll remove it.

libdrgn/python/symbol_index.c
```diff
@@ -0,0 +1,141 @@
...
+	static char *kwnames[] = {"symbols", NULL};
+	PyObject *list_obj;
+
+	if (!PyArg_ParseTupleAndKeywords(args, kwds, "O", kwnames, &list_obj))
+		return NULL;
+
+	_cleanup_pydecref_ PyObject *seq_obj =
+		PySequence_Fast(list_obj, "expected sequence of Symbols");
+	if (!seq_obj)
+		return NULL;
```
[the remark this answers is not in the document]
brenns10: Yes, I agree! Will replace this with an iterator.

libdrgn/python/symbol_index.c
```diff
@@ -0,0 +1,141 @@
...
+	size_t len = PySequence_Fast_GET_SIZE(seq_obj);
+	if (len == 0)
+		return PyErr_Format(PyExc_ValueError,
+				    "symbol finder must contain at least one symbol");
```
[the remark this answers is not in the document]
brenns10: There is a line in the initialization function which unconditionally indexes into the array at element 0, and I believe that's why I had enforced that requirement. But it's trivial to guard against the case of an empty array, and I see your point about empty symbol indices being useful (or at least, leading to simpler code for users). So I fixed that up, removed the requirement, and added tests to ensure that initializing and querying an empty `SymboIndex` is safe and correct.

libdrgn/python/symbol_index.c
```diff
@@ -0,0 +1,141 @@
...
+		Symbol *sym = (Symbol *)item;
+		if (!drgn_symbol_index_builder_add(&builder, sym->sym)) {
+			drgn_symbol_index_builder_deinit(&builder);
+			return PyErr_NoMemory();
+		}
+	}
+
+	struct drgn_error *err =
+		drgn_symbol_index_init_from_builder(&index_obj->index, &builder);
+	// On error, the builder and index are already deinitialized
```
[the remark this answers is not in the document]
brenns10: Good catch. The answer was no. I know I could simply not use the cleanup attribute for this case, but honestly I like them so much that I think the easier answer is to ensure that these data structures can be deinitialized twice safely. So I added a memset to `drgn_symbol_index_deinit()` to make it safe, and documented the error conditions a bit better.

libdrgn/symbol.c
```diff
@@ -174,3 +178,291 @@ void drgn_symbol_result_builder_array(struct drgn_symbol_result_builder *builder
...
+	for (uint32_t i = 1; i < ret->num_syms; i++) {
+		uint64_t max_addr = ret->symbols[i].address + ret->symbols[i].size;
+		ret->max_addrs[i] = max(ret->max_addrs[i - 1], max_addr);
+	}
+
+	// Sort the "name_sort" array by name so we get runs of symbols with the
+	// same name
+	for (uint32_t i = 0; i < ret->num_syms; i++)
+		ret->name_sort[i] = i;
+	qsort_r(ret->name_sort, ret->num_syms, sizeof(ret->name_sort[0]),
```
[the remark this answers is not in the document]
brenns10: I liked the helper function you wrote on that issue a lot since it makes the code a bit easier to read. So I added it under the name `qsort_arg` and converted existing users. But if you'd prefer I just use a thread local directly without an added helper function, I can switch it :)

tests/linux_kernel/helpers/test_kallsyms.py
```diff
@@ -68,7 +72,7 @@ def test_static_data(self):
             Symbol("null", 0x0, 8, SymbolBinding.UNIQUE, SymbolKind.UNKNOWN),
             Symbol("local_data", 0x8, 8, SymbolBinding.UNKNOWN, SymbolKind.OBJECT),
             Symbol("global_bss", 0x10, 16, SymbolBinding.GLOBAL, SymbolKind.OBJECT),
-            Symbol("weak_symbol", 0x20, 32, SymbolBinding.WEAK, SymbolKind.UNKNOWN),
+            Symbol("weak_symbol", 0x20, 32, SymbolBinding.WEAK, SymbolKind.OBJECT),
```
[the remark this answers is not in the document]
brenns10: Yes, looks like it, thank you!

libdrgn/kallsyms.c
```diff
@@ -0,0 +1,625 @@
...
+/** Load kallsyms directly from the /proc/kallsyms file */
+struct drgn_error *drgn_load_proc_kallsyms(const char *filename, bool modules,
+					   struct drgn_symbol_index *ret)
+{
+	FILE *fp = fopen(filename, "r");
+	if (!fp)
+		return drgn_error_create_os("Error opening kallsyms", errno, filename);
+
+	struct drgn_error *err = NULL;
+	struct drgn_symbol sym = {0};
```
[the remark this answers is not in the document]
brenns10: I'm curious, I thought that empty initializers are a C23 feature / GNU extension. Is that ok for drgn's C11 requirement?

libdrgn/kallsyms.c
```diff
@@ -0,0 +1,625 @@
...
+			 * we expect it to be \0 */
+			 err = drgn_error_format(DRGN_ERROR_SYNTAX,
+						 "Invalid address \"%s\" in kallsyms line %zu",
+						 addr_str, line_number);
+			 break;
+		}
+
+		// Append the previous symbol now that we have the next symbol's
+		// address to compute the size, only if we didn't get a symbol
+		// size before.
```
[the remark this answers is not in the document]
brenns10: That's a really good question! I can't parse what I meant. I think I may have copied this from a different context and failed to recognize the comment doesn't make sense. Deleted it.

libdrgn/kallsyms.c
```diff
@@ -0,0 +1,625 @@
...
+struct drgn_error *drgn_load_proc_kallsyms(const char *filename, bool modules,
+					   struct drgn_symbol_index *ret)
+{
+	FILE *fp = fopen(filename, "r");
+	if (!fp)
+		return drgn_error_create_os("Error opening kallsyms", errno, filename);
+
+	struct drgn_error *err = NULL;
+	struct drgn_symbol sym = {0};
+	struct drgn_symbol_index_builder builder;
```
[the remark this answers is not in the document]
brenns10: Yes it does. Thanks to the changes I made earlier, it's now safe to deinitialize this even after constructing the symbol index, so I added `_cleanup_(drgn_symbol_index_builder_deinit)` to the declaration.

libdrgn/kallsyms.c
brenns10: The entire purpose of `MAX_SYMBOL_LENGTH` is to ensure we don't allow `__per_cpu_end` to have a size which extends across the entire userspace memory region (and possibly similar issues in other architectures). So I don't mind increasing this.

That said, I did notice that on my kernel (6.9 based) there are some kallsyms symbols that are now coming out of order:
```
ffffffffc028d4e0 t __pfx_fuse_invalidate_atime  [fuse]
ffffffffc02900f0 t __pfx_fuse_do_setattr        [fuse]
ffffffffc028f8a0 t fuse_reverse_inval_entry     [fuse]
```
This was unintentionally catching the overflow from the subtraction here. I've now explicitly caught that overflow. This only happens for module symbols -- vmlinux symbols are in order. I don't really have any solution for this other than not using `/proc/kallsyms` for module kallsyms: getting it from `struct module` is better.

libdrgn/kallsyms.c
brenns10: Yes, what a catch. I'm not sure how I didn't notice the broken addresses for these.

I'm able to test:

- `!BASE_RELATIVE && !ABSOLUTE_PERCPU` - this means kallsyms addresses are just an array of unsigned long. This is _not_ the default, but it's possible to build a normal x86_64 kernel with this. This was affected by this issue and will be fixed.
- `BASE_RELATIVE && ABSOLUTE_PERCPU` - this means kallsyms addresses are 32-bit with a base address, and they have special handling for negative values for percpu region. This is the default for all x86_64 kernels.
- `BASE_RELATIVE && !ABSOLUTE_PERCPU` - same as the previous, but no handling of negatives for the percpu region. Most x86_64 kernels fail to build because they can't encode percpu addresses. But this is the default for aarch64 kernels and so it gets lots of testing too.

There is no `!BASE_RELATIVE && ABSOLUTE_PERCPU` - or maybe there is but it doesn't result in a functional difference, so there's nothing to test.

libdrgn/kallsyms.c
brenns10: Yes, I think that's worth guarding against. I think the easiest place to do it is actually in `drgn_load_builtin_kallsyms()`, where we can simply check whether the string builder length is zero and raise an error.

There's a lot of stuff in the kallsyms tables that, if fuzzed, could cause these parsers to behave weirdly or maybe crash. I'm not certain how paranoid we want to be here. At this point, if the correct set of symbols are provided, with real values, and we've managed to read the token and name table successfully (without faulting or mallocing a buffer that's way too large), then it seems likely that the data will match the kallsyms format. So in practical terms, these sorts of things shouldn't happen. What are your opinions here? I can add paranoia to this, but it does feel like it will take rather unclear code and make it even worse...

brenns10:
Ok after a few botched pushes I think I've addressed everything here. I've also rebased on `main` to drop my `drgn.helpers.linux.module` changes which conflicted with #411. I did leave open the comments with more substantial replies in case you want to continue those discussions, though I think I addressed those too. Thanks!

brenns10:
Noticed that the branch had become stale due to some merge conflicts.

This first push here has just resolved the conflicts. Got a lint error (git rebase does not run pre-commit :/). I still need to incorporate your new binary search macro, so I'll resolve both shortly.

brenns10:
I'm loving the binary search helpers, they simplified things a lot here:
```diff
diff --git a/libdrgn/symbol.c b/libdrgn/symbol.c
index 46e21a6b..92bcfdb9 100644
--- a/libdrgn/symbol.c
+++ b/libdrgn/symbol.c
@@ -6,6 +6,7 @@
 #include <stdlib.h>
 #include <string.h>
 
+#include "binary_search.h"
 #include "drgn_internal.h"
 #include "string_builder.h"
 #include "symbol.h"
@@ -300,38 +301,21 @@ drgn_symbol_index_deinit(struct drgn_symbol_index *index)
 static void address_search_range(struct drgn_symbol_index *index, uint64_t address,
 				 uint32_t *start_ret, uint32_t *end_ret)
 {
-	// bsearch() only handles exact matches, and we need to find what is
-	// essentially the correct insertion point, even if there's not an exact
-	// match. So bsearch() can't be used here. It's helpful to think of this
-	// code in terms of the Python library functions bisect_right().
-	uint32_t lo = 0, hi = index->num_syms;
-	uint32_t mid;
-
 	// First, identify the maximum symbol index which could possibly contain
 	// this address. Think of this as:
 	//   end_ret = bisect_right([s.address for s in symbols], address)
-	while (lo < hi) {
-		mid = lo + (hi - lo) / 2;
-		if (address < index->symbols[mid].address)
-			hi = mid;
-		else
-			lo = mid + 1;
-	}
-	*end_ret = lo;
+	#define less_than_start(a, b) (*(a) < (b)->address)
+	*end_ret = binary_search_gt(index->symbols, index->num_syms, &address,
+				    less_than_start);
+	#undef less_than_start
 
 	// Second, identify first symbol index which could possibly contain this
 	// address. We need to use "max_addrs" for this task:
 	//    bisect_right(max_addrs, address)
-	hi  = lo;
-	lo = 0;
-	while (lo < hi) {
-		mid = lo + (hi - lo) / 2;
-		if (address < index->max_addrs[mid])
-			hi = mid;
-		else
-			lo = mid + 1;
-	}
-	*start_ret = lo;
+	#define less_than_end(a, b) (*(a) < *(b))
+	*start_ret = binary_search_gt(index->max_addrs, index->num_syms, &address,
+				    less_than_end);
+	#undef less_than_end
 }
 
 struct drgn_error *
```

libdrgn/symbol.c
```diff
@@ -174,3 +178,286 @@ void drgn_symbol_result_builder_array(struct drgn_symbol_result_builder *builder
...
+drgn_symbol_index_init_from_builder(struct drgn_symbol_index *index,
+				    struct drgn_symbol_index_builder *builder)
+{
+	size_t names_len = builder->names.len;
+	char *names = string_builder_steal(&builder->names);
+	names = realloc(names, names_len);
+	if (!names) {
+		symbol_vector_deinit(&builder->symbols);
+		return &drgn_enomem;
+	}
```
osandov, line 426: This leaks the old `names` pointer in the failure case. We can tolerate `realloc` failing, so maybe:
```suggestion
	char *tmp_names = realloc(names, names_len);
	if (tmp_names)
		names = tmp_names;
```

osandov (commented):
Signing off for the day after looking at commits 1-3. Will continue tomorrow, thanks!

libdrgn/symbol.c
```diff
@@ -174,3 +178,286 @@ void drgn_symbol_result_builder_array(struct drgn_symbol_result_builder *builder
...
+	free(index->symbols);
+	free(index->max_addrs);
+	drgn_symbol_name_table_deinit(&index->htab);
+	free(index->strings);
+	free(index->name_sort);
+	// Simplify error handling by ensuring deinit is safe to call twice
+	memset(index, 0, sizeof(*index));
+}
+
+static void address_search_range(struct drgn_symbol_index *index, uint64_t address,
```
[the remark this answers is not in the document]
brenns10: Wow, brain fart, I thought I had included that into the branch, but apparently not. Updated now with that folded into the proper commit.

libdrgn/kallsyms.c
```diff
@@ -0,0 +1,611 @@
...
+ * @param offset Starting index within names array for this symbol
+ * @param sb Buffer to write output symbol to
+ * @param[out] kind_ret Where to write the symbol kind data
+ * @returns The offset of the next symbol in the names array
+ */
+static unsigned int
+kallsyms_expand_symbol(struct kallsyms_reader *kr, unsigned int offset,
+		       struct string_builder *sb, char *kind_ret)
+{
+	uint8_t *data = &kr->names[offset];
```
osandov, line 210: It's *technically* possible that the data that we read when we read the whole table is different from the data that we read when we skipped through the lengths. (Yes, practically that won't happen, but imagine a buggy memory reader or something like that.) So, we need to make sure that `offset` is in bounds here and when we read the long length.

Likewise, check that `offset + len` is within bounds, too.

libdrgn/kallsyms.c
```diff
@@ -0,0 +1,611 @@
...
+	if ((len & 0x80) && kr->long_names) {
+		data++;
+		offset++;
+		len = (0x7F & len) | (*data << 7);
+	}
+
+	offset += len + 1;
+	data += 1;
+	while (len) {
+		char *token_ptr = &kr->token_table[kr->token_index[*data]];
```
osandov, line 223: We need to make sure that `kr->token_index[*data] < len(kr->token_table)` (where `len(kr->token_table)` is the index of the last null-terminator). It'd probably be best to validate the entire `token_index` once in `kallsyms_copy_tables()` so that we don't need to check here.

libdrgn/kallsyms.c
```diff
@@ -0,0 +1,611 @@
...
+		while (*token_ptr) {
+			if (skipped_first) {
+				if (!string_builder_appendc(sb, *token_ptr)) {
+					string_builder_null_terminate(sb);
```
osandov, line 227: Are we handling ENOMEM by truncating the string? If the `appendc` failed, then this `null_terminate` will likely fail, and we'll return a non-null-terminated string and blow up later. This needs to actually return an error.

libdrgn/kallsyms.c
```diff
@@ -0,0 +1,611 @@
...
+				skipped_first = true;
+			}
+			token_ptr++;
+		}
+
+		data++;
+		len--;
+	}
+
+	string_builder_null_terminate(sb);
```
osandov, line 241: This can fail with ENOMEM.

drgn/helpers/linux/kallsyms.py
```diff
@@ -73,3 +84,155 @@ def load_vmlinux_kallsyms(prog: Program) -> SymbolIndex:
...
+        has_typetab = True
+    except LookupError:
+        has_typetab = False
+
+    # The symtab field is a pointer, but it points at an array of Elf_Sym
+    # objects. Indexing it requires drgn to do pointer arithmetic and issue a
+    # lot of very small /proc/kcore reads, which can be a real performance
+    # issue. So convert it into an object representing a correctly-sized array,
+    # and then read that object all at once. This does one /proc/kcore read,
+    # which is a major improvement!
```
osandov, line 198: Note to self: this is another use-case for #377.

drgn/helpers/linux/kallsyms.py
```diff
@@ -73,3 +84,155 @@ def load_vmlinux_kallsyms(prog: Program) -> SymbolIndex:
...
+    return Symbol(  # type: ignore
+        name,
+        obj.st_value.value_(),
+        obj.st_size.value_(),
+        binding,
+        kind,
+    )
+
+
+def module_kallsyms(module: Object) -> List[Symbol]:
```
osandov, line 162: You mentioned that you wouldn't mind removing this from the public API for now, so let's do that unless you've changed your mind.

osandov (changes requested):
Okay, that's the rest of it, looking good overall. I gave the kallsyms code a much closer read than I had before, so there are some new comments there that I missed in the last round. Have a good weekend!

libdrgn/python/program.c
```diff
@@ -504,6 +504,13 @@ py_symbol_find_fn(const char *name, uint64_t addr,
 		  enum drgn_find_symbol_flags flags, void *arg,
 		  struct drgn_symbol_result_builder *builder)
 {
+	// Fast path for SymbolIndex: don't bother converting to and from Python
+	// types, as this is a C finder.
+	if (PyObject_TypeCheck(PyTuple_GET_ITEM(arg, 1), &SymbolIndex_type)) {
```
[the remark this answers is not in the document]
brenns10: Fundamentally we don't actually care about subclasses of `SymbolIndex_type`, because it actually can't be subclassed right now:

```python
>>> from drgn import SymbolIndex
>>> class MySymbolIndex(SymbolIndex):
...     def __call__(self, *args, **kwargs):
...         print("I'm a problem for you!")
...         super().__call__(self, *args, **kwargs)
  File "<stdin>", line 1, in <module>
TypeError: type '_drgn.SymbolIndex' is not an acceptable base type
```

And even if it could be inherited from, that would actually be a problem, since we would ignore any overridden `__call__()` method with this fast path.

Would it be any safer to do `Py_TYPE(PyTuple_GET_ITEM(arg, 1)) == &SymbolIndex_type`? This would avoid the logic for checking if it's a subclass altogether.

libdrgn/util.h
```diff
@@ -211,4 +211,15 @@ static inline uint64_t uint_max(int n)
...
 
+/**
+ * Similar to qsort_r (passes @a arg to @a compar) but **not** reentrant
+ *
+ * The qsort_r() function's main feature is that it is reentrant, but also adds
+ * the convenience of including an argument to the callback function.
+ * Unfortunately it is a glibc extension. This provides a similar API but it is
+ * only thread-safe, not reentrant. See qsort_r(3) for details.
+ */
+void qsort_arg(void *base, size_t nmemb, size_t size,
```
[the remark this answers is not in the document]
brenns10: Hm, I'll look into it and at least throw it onto the pile of lower priority cleanups I have planned for later. I probably should learn the autotools magic necessary for those sorts of checks anyway.

libdrgn/python/helpers.c
```diff
@@ -291,3 +292,56 @@ PyObject *drgnpy_linux_helper_pgtable_l5_enabled(PyObject *self, PyObject *arg)
...
+PyObject *
+drgnpy_linux_helper_load_builtin_kallsyms(PyObject *self, PyObject *args,
+					  PyObject *kwds)
+{
+	static char *kwnames[] = {"prog", "names", "token_table", "token_index", "num_syms",
+	                          "offsets", "relative_base", "addresses", "_stext", NULL};
+	struct kallsyms_locations kl;
+	PyObject *prog_obj;
+
+	if (!PyArg_ParseTupleAndKeywords(args, kwds, "OKKKKKKKK", kwnames,
```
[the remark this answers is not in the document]
brenns10: I created a `u64_converter` similar to the `index_converter` to help contain the messiness. It's still a bit more verbose but at least it's easy to read and there's no juggling a bunch of PyObjects.

libdrgn/python/program.c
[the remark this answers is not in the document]
osandov: Yeah, I think we can get away with that.

libdrgn/util.h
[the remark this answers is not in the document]
osandov: I'm looking into migrating from our setuptools+autotools mess to Meson, so don't sweat learning the autotools stuff for now unless you really want to.

libdrgn/kallsyms.c
```diff
@@ -0,0 +1,611 @@
...
+	names_idx = 0;
+	kr->long_names = guess_long_names(prog);
+	for (size_t i = 0; i < kr->num_syms; i++) {
+		err = drgn_program_read_u8(prog,
+					   loc->kallsyms_names + names_idx,
+					   false, &len_u8);
+		if (err)
+			return err;
+		len = len_u8;
+		if ((len & 0x80) && kr->long_names) {
```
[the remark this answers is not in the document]
brenns10: Considering that the max we could represent with 2 bytes of ULB128 is actually even lower (2^14 == 16k) and the bad track record for these sorts of limits, I definitely get your concern. Though I think there's an equally valid concern that, since kallsyms is a bespoke format, it may not make sense to support anything beyond what the existing kallsyms generator & consumer support. From the existing code, we have no real guarantee that they would use ULEB128 with three bytes, versus making some more substantial change to the format to accommodate larger symbol names.

I might go ahead and add a comment, but also raise an error if the MSB of the second byte is set (this would indicate a third byte is expected). That way we can be made aware of format breakages, but we don't make any assumptions about the future behavior of kallsyms.

Let me know if you disagree with the approach though.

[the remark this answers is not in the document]
osandov: I didn't look at the kernel kallsyms code long enough to realize that they are also hard-coding at most 2 bytes despite describing it as ULEB128, so I'm fine with the comment either with or without the error.

libdrgn/kallsyms.c
brenns10: Yeah, I think this was an airheaded attempt to avoid having to include an error in this function's signature 🙃

brenns10:
I double checked to ensure the s390x failure was transient, so we've got a clean test run :)

I believe I have addressed all your comments, and I avoided rebasing on the `main` so the diff between versions stays clean. Thanks for the review again!

osandov:
Yeah Packit has been flaky today. I'll give these another look today!

libdrgn/kallsyms.c
```diff
@@ -0,0 +1,659 @@
...
+
+	// Bounds check the bytes we will read.
+	*offset += len;
+	if (*offset >= kr->names_len)
+		return drgn_error_format(DRGN_ERROR_OTHER,
+					 "kallsyms: invalid offset %" PRIu32 ", max offset is %zu",
+					 *offset, kr->names_len);
+	// Leave offset pointing at the next token, which we will bounds check
+	// on the next call.
+	*offset += 1;
```
osandov, line 260: I should've mentioned this earlier, but `struct binary_buffer` can save you from doing a lot of this tricky bounds checking manually. Something like this passes tests, but please look it over and update the `kallsyms_expand_symbol` comment:

```diff
diff --git a/libdrgn/kallsyms.c b/libdrgn/kallsyms.c
index 5057dffc..1e939959 100644
--- a/libdrgn/kallsyms.c
+++ b/libdrgn/kallsyms.c
@@ -4,6 +4,7 @@
 #include <ctype.h>
 #include <stddef.h>
 
+#include "binary_buffer.h"
 #include "kallsyms.h"
 #include "program.h"
 #include "drgn_internal.h"
@@ -221,6 +222,14 @@ kallsyms_copy_tables(struct drgn_program *prog, struct kallsyms_reader *kr,
 	return NULL;
 }
 
+static struct drgn_error *kallsyms_binary_buffer_error(struct binary_buffer *bb,
+						       const char *pos,
+						       const char *message)
+{
+	return drgn_error_format(DRGN_ERROR_OTHER,
+				 "couldn't parse kallsyms: %s", message);
+}
+
 /**
  * Extract the symbol name and type
  * @param kr Registry containing kallsyms data
@@ -230,36 +239,22 @@ kallsyms_copy_tables(struct drgn_program *prog, struct kallsyms_reader *kr,
  * @returns The offset of the next symbol in the names array
  */
 static struct drgn_error *
-kallsyms_expand_symbol(struct kallsyms_reader *kr, uint32_t *offset,
+kallsyms_expand_symbol(struct kallsyms_reader *kr,
+		       struct binary_buffer *names_bb,
 		       struct string_builder *sb, char *kind_ret)
 {
-	if (*offset >= kr->names_len)
-		return drgn_error_format(DRGN_ERROR_OTHER,
-					 "kallsyms: invalid offset %" PRIu32 ", max offset is %zu",
-					 *offset, kr->names_len);
-	uint8_t *data = &kr->names[*offset];
-	unsigned int len = *data;
+	struct drgn_error *err;
+
+	uint64_t len;
+	err = binary_buffer_next_uleb128(names_bb, &len);
+	if (err)
+		return err;
+	const uint8_t *data = (uint8_t *)names_bb->pos;
+	err = binary_buffer_skip(names_bb, len);
+	if (err)
+		return err;
+
 	bool skipped_first = false;
-
-	// Bounds checking: if we've overrun the buffer, we'll return an error
-	// below when we bounds check *offset + len + 1
-	if ((len & 0x80) && kr->long_names && *offset + 1 < kr->names_len) {
-		data++;
-		(*offset)++;
-		len = (0x7F & len) | (*data << 7);
-	}
-
-	// Bounds check the bytes we will read.
-	*offset += len;
-	if (*offset >= kr->names_len)
-		return drgn_error_format(DRGN_ERROR_OTHER,
-					 "kallsyms: invalid offset %" PRIu32 ", max offset is %zu",
-					 *offset, kr->names_len);
-	// Leave offset pointing at the next token, which we will bounds check
-	// on the next call.
-	*offset += 1;
-
-	data += 1;
 	while (len) {
 		char *token_ptr = &kr->token_table[kr->token_index[*data]];
 		while (*token_ptr) {
@@ -291,12 +286,14 @@ search_for_string(struct kallsyms_reader *kr, const char *name, ssize_t *ret)
 {
 	STRING_BUILDER(sb);
 	size_t len = strlen(name);
-	uint32_t names_idx = 0;
+	struct binary_buffer names_bb;
+	binary_buffer_init(&names_bb, kr->names, kr->names_len, false,
+			   kallsyms_binary_buffer_error);
 	for (ssize_t i = 0; i < kr->num_syms; i++) {
 		char kind;
 		sb.len = 0;
 		struct drgn_error *err =
-			kallsyms_expand_symbol(kr, &names_idx, &sb, &kind);
+			kallsyms_expand_symbol(kr, &names_bb, &sb, &kind);
 		if (err)
 			return err;
 		if (sb.len == len && strcmp(name, sb.str) == 0) {
@@ -534,13 +531,15 @@ drgn_load_builtin_kallsyms(struct drgn_program *prog,
 	drgn_symbol_index_builder_init(&builder);
 	STRING_BUILDER(sb);
 
-	uint32_t names_idx = 0;
+	struct binary_buffer names_bb;
+	binary_buffer_init(&names_bb, kr.names, kr.names_len, false,
+			   kallsyms_binary_buffer_error);
 	for (int i = 0; i < kr.num_syms; i++) {
 		struct drgn_symbol symbol;
 		char kind;
 		uint64_t size = 0;
 		sb.len = 0;
-		err = kallsyms_expand_symbol(&kr, &names_idx, &sb, &kind);
+		err = kallsyms_expand_symbol(&kr, &names_bb, &sb, &kind);
 		if (err)
 			return err;
 		if (sb.len == 0)
```

libdrgn/kallsyms.c
```diff
@@ -0,0 +1,659 @@
...
+				       kr->token_table_len, false);
+	if (err)
+		return err;
+
+	// Ensure that all members of token_index are in-bounds for indexing
+	// into token_table.
+	for (size_t i = 0; i <= UINT8_MAX; i++)
+		if (kr->token_index[i] >= kr->token_table_len)
+			return drgn_error_format(DRGN_ERROR_OTHER,
+						 "kallsyms: token_index out of bounds (token_index[%zu] = %u >= %zu",
```
osandov, line 171: Unclosed parenthesis in your error message :)

osandov (changes requested):
Thanks, just need a few final easy fixes.

libdrgn/kallsyms.c
brenns10: Wow, that looks like an incredibly convenient API!

brenns10: I reimplemented it myself from the binary buffer docstrings, and came to basically the same code, so I feel good about the correctness. Thanks for this, I'll probably end up redoing some of my BTF code with the binary_buffer for better safety.

libdrgn/python/util.c
```diff
@@ -78,6 +78,20 @@ int index_converter(PyObject *o, void *p)
 	}
 }
 
+int u64_converter(PyObject *o, void *p)
+{
+	uint64_t *arg = p;
+
+	if (o == Py_None)
+		return 1;
+
```
osandov, line 87: This leaves the argument uninitialized if it is `None`, which I don't think is what you intended. If you drop this`, `PyNumber_Index()` will take care of the error checking.
```suggestion
```

osandov:
(LGTM other than the `u64_converter()` comment)

libdrgn/python/util.c
brenns10: Ah, thank you for that catch. I switched up the cases when I was re-reading the function quickly.

osandov (approved):
Thanks for all of your work on this, it looks ready to go :tada:

brenns10:
Yay, that's great! Thanks for all the reviewing work on this!

PR #389, opened 2024-03-19
contrib/irq: fix for v6.5 and later kernels.

imran-kn:
v6.5 moved irq descriptors into a maple tree,
earlier they used to exist in a radix tree.
Change contrib/irq.py to accomodate this change.


osandov:
Ran it on 6.6 and it worked, thanks!

Issue #350, opened 2023-08-22
Allow overriding / providing vmcoreinfo data


brenns10:
I had a bit of a hack that let me specify my own vmcoreinfo note, in case the one provided is incorrect or missing.

* Incorrect vmcoreinfo is generated by Hyper-V when generating a kdump from a hypervisor snapshot. See: https://github.com/Azure/azure-linux-utils/issues/12
* Qemu/libvirt devices can be missing vmcoreinfo if the hypervisor or guest didn't enable support for the `-device vmcoreinfo` virtual device.

In either case, the vmcoreinfo note itself is generally still present somewhere inside the guest memory, and it's not actually that tough to find. I have a [tool](https://github.com/brenns10/kernel_stuff/blob/master/vmcoreinfo/dumpphys.c) that can search the pages of an ELF/kdump vmcore for the vmcoreinfo note, rather quickly based on the idea that it will only appear at a PAGE_SIZE boundary, and it will start with `OSRELEASE=`.

Supposing you've used such a tool, Drgn should have an advanced feature somewhere to allow you to specify your own vmcoreinfo note. Crash doesn't allow this exact feature -- but it lets you specify architecture/configuration details such as KASLR offsets which would normally be found in the note.

The hack I had allowed you to specify the vmcoreinfo note contents in an environment variable, which was a bad API. Spitballing a few APIs here (mostly thinking of Python-facing API):

1. Allow `set_core_dump()` (and friends) to accept an optional buffer which contains the vmcoreinfo. This would raise the question of whether `set_kernel()` would allow it, since that's functionally the same as `set_core_dump("/proc/kcore")`.
1. Add an optional parameter to the `Program` constructor.
1. Allow the user to specify the file path for a file containing vmcoreinfo, via an environment variable.

Of those, I prefer the optional parameter to the `Program` constructor. It's nice because it mirrors the `platform` parameter, and in cases where vmcoreinfo is broken, typically the other fields which drgn relies on to determine the platform are also broken, so Drgn would fail to read the vmcore without specifying platform as well.

@osandov if you have a preferred API, let me know. Feel free to assign me on this (I can't update assignees, labels, etc). I'll probably submit something within the next month or two when I find time for it.

Homura2333:
Hi, I encountered an issue while using `makedumpfile -R new_vmcore <vmcore` followed by running `sudo dragon -c new_vmcore`. I received the following error:
```
[...]

PR #396, opened 2024-05-24
Support setting custom vmcoreinfo

brenns10:
This branch has sat on the back burner for a while, but new discussion in #350 led me to update it and get it ready to merge. It adds the ability to give drgn some data to be interpreted as the vmcoreinfo when the Program is created. Later, when a core dump is attached, drgn will respect the pre-existing info rather than trying to get it from the core dump.

There's a bit of a delicate dance to be had with libkdumpfile, because it also wants to use the vmcoreinfo to find the page tables. For vmcores missing the vmcoreinfo, it seems that libkdumpfile doesn't know how to find their architecture, so we need to manually provide the architecture, then the dump FD, then the vmcoreinfo.

While I was at it, I threw in RISC-V support for libkdumpfile, which has been available since 0.5.4, but since we have some if-statements that enumerate all the libkdumpfile architectures, we never updated those and thus haven't been able to use it (I think... I don't actually have a RISC-V core dump handy).

---

Testing this is of course the important part. One obvious candidate where this code makes drgn work is hypervisor core dumps that are missing vmcoreinfo. QEMU will produce them if you don't have `-device vmcoreinfo`, which drgn already notes and produces a nice error message about.

Thankfully, drgn's own test framework runs QEMU without that option, so it's easy to generate a good test vmcore:

```
python -m vmtest.vm -k 6.9*
...
# Issue key sequence: Ctrl-A C to switch to the QEMU monitor
(qemu) dump-guest-memory file.elf
## OR
(qemu) dump-guest-memory -z file.kdump.flat
```

Running drgn against it will fail:

```
$ drgn -c file.kdump.flat -s build/vmtest/x86_64/kernel-6.9.1-vmtest29.1default/build/vmlinux
drgn 0.0.26 (using Python 3.9.18, elfutils 0.190, with libkdumpfile)
warning: the given file is in the makedumpfile flattened format; if open fails or is too slow, reassemble it with 'makedumpfile -R newfile <oldfile'
  File "/home/stepbren/.local/bin/drgn", line 33, in <module>
    sys.exit(load_entry_point('drgn', 'console_scripts', 'drgn')())
  File "/usr/lib64/python3.9/site-packages/drgn/cli.py", line 270, in _main
    prog.set_core_dump(args.core)
Exception: kdump_vmcoreinfo_raw: linux.vmcoreinfo.raw is not set

$ drgn -c foo.elf -s build/vmtest/x86_64/kernel-6.9.1-vmtest29.1default/build/vmlinux
drgn 0.0.26 (using Python 3.9.18, elfutils 0.190, with libkdumpfile)
error: unrecognized QEMU memory dump; for Linux guests, run QEMU with '-device vmcoreinfo', compile the kernel with CONFIG_FW_CFG_SYSFS and CONFIG_KEXEC, and load the qemu_fw_cfg kernel module before dumping the guest memory (requires Linux >= 4.17 and QEMU >= 2.11)
```

But with this branch, you can use my [dumpphys](https://github.com/brenns10/kernel_stuff/blob/master/vmcoreinfo/dumpphys.c) tool like below to get a text file containing the vmcoreinfo:
```
dumpphys -c file.kdump.flat -o file.vmcoreinfo -i
## or
dumpphys -c file.elf -o file.vmcoreinfo -i
```

And then, you can go ahead and run drgn, but add new arguments `-i file.vmcoreinfo` to specify the vmcoreinfo, and `-a X86_64` to specify the architecture (since libkdumpfile cannot detect it). At that point, drgn works like normal.

---

The current known issue with this branch is that this won't work with ELF vmcores that drgn opens natively. Drgn expects to have virtual address mappings provided for it, at least for the linear/direct map. When they're not there, drgn will fail with a recursive page translation error:

```
>>> prog["slab_caches"]
  File "/home/stepbren/repos/drgn/drgn/cli.py", line 141, in _displayhook
    text = value.format_(columns=shutil.get_terminal_size((0, 0)).columns)
_drgn.FaultError: recursive address translation; page table may be missing from core dump: 0xffffffff9662aff8
```

A workaround is to use libkdumpfile for those vmcores via `DRGN_USE_LIBKDUMPFILE_FOR_ELF`. I would like to address that issue next, but I'd like to do it properly, and that may be a bit more involved (do we add a new arch-specific callback? etc). I was hoping this could be merged since it's a bit more straightforward, and then we could focus on the best solution for these recursive address translation errors.

### DIFF
--- a/_drgn.pyi
+++ b/_drgn.pyi
@@ -67,7 +67,11 @@ class Program:
     :meth:`[] <.__getitem__>` operator.
     """
 
-    def __init__(self, platform: Optional[Platform] = None) -> None:
+    def __init__(
+        self,
+        platform: Optional[Platform] = None,
+        vmcoreinfo: Union[bytes, str, None] = None,
+    ) -> None:
         """
         Create a ``Program`` with no target program. It is usually more
         convenient to use one of the :ref:`api-program-constructors`.
@@ -75,6 +79,9 @@ class Program:
         :param platform: The platform of the program, or ``None`` if it should
             be determined automatically when a core dump or symbol file is
             added.
+        :param vmcoreinfo: Optionally provide the ``VMCOREINFO`` note data for
+            Linux kernel core dumps, which will override any detected data. When
+            not provided or ``None``, automatically detect the info.
         """
         ...
     flags: ProgramFlags
--- a/drgn/cli.py
+++ b/drgn/cli.py
@@ -210,6 +210,21 @@ def _main() -> None:
     )
 
     parser.add_argument(
+        "-a",
+        "--architecture",
+        metavar="ARCH",
+        choices=[a.name for a in drgn.Architecture]
+        + [a.name.lower() for a in drgn.Architecture],
+        help="set the program architecture, in case it can't be auto-detected",
+    )
+    parser.add_argument(
+        "-i",
+        "--vmcoreinfo",
+        type=str,
+        default=None,
+        help="path to vmcoreinfo file (overrides any already present in the file)",
+    )
+    parser.add_argument(
         "--log-level",
         choices=["debug", "info", "warning", "error", "critical", "none"],
         default="warning",
@@ -258,7 +273,16 @@ def _main() -> None:
     else:
         logger.setLevel(args.log_level.upper())
 
-    prog = drgn.Program()
+    platform = None
+    if args.architecture:
+        platform = drgn.Platform(drgn.Architecture[args.architecture.upper()])
+
+    vmcoreinfo = None
+    if args.vmcoreinfo:
+        with open(args.vmcoreinfo, "rb") as f:
+            vmcoreinfo = f.read()
+
+    prog = drgn.Program(platform=platform, vmcoreinfo=vmcoreinfo)
     try:
         if args.core is not None:
             prog.set_core_dump(args.core)
--- a/libdrgn/kdump.c
+++ b/libdrgn/kdump.c
@@ -35,11 +35,16 @@ static struct drgn_error *drgn_platform_from_kdump(kdump_ctx_t *ctx,
 		arch = &arch_info_arm;
 	else if (strcmp(str, KDUMP_ARCH_PPC64) == 0)
 		arch = &arch_info_ppc64;
-	/* libkdumpfile doesn't support RISC-V */
 	else if (strcmp(str, KDUMP_ARCH_S390X) == 0)
 		arch = &arch_info_s390x;
 	else if (strcmp(str, KDUMP_ARCH_S390) == 0)
 		arch = &arch_info_s390;
+#if KDUMPFILE_VERSION >= KDUMPFILE_MKVER(0, 5, 4)
+	else if (strcmp(str, KDUMP_ARCH_RISCV64) == 0)
+		arch = &arch_info_riscv64;
+	else if (strcmp(str, KDUMP_ARCH_RISCV32) == 0)
+		arch = &arch_info_riscv32;
+#endif
 	else
 		arch = &arch_info_unknown;
 
@@ -79,11 +84,16 @@ static struct drgn_error *drgn_platform_to_kdump(kdump_ctx_t *ctx,
 		arch_str = KDUMP_ARCH_ARM;
 	else if (platform->arch == &arch_info_ppc64)
 		arch_str = KDUMP_ARCH_PPC64;
-	/* libkdumpfile doesn't support RISC-V */
 	else if (platform->arch == &arch_info_s390x)
 		arch_str = KDUMP_ARCH_S390X;
 	else if (platform->arch == &arch_info_s390)
 		arch_str = KDUMP_ARCH_S390;
+#if KDUMPFILE_VERSION >= KDUMPFILE_MKVER(0, 5, 4)
+	else if (platform->arch == &arch_info_riscv64)
+		arch_str = KDUMP_ARCH_RISCV64;
+	else if (platform->arch == &arch_info_riscv32)
+		arch_str = KDUMP_ARCH_RISCV32;
+#endif
 
 	if (arch_str) {
 		ks = kdump_set_string_attr(ctx, KDUMP_ATTR_ARCH_NAME, arch_str);
--- a/libdrgn/kdump.c
+++ b/libdrgn/kdump.c
@@ -63,6 +63,40 @@ static struct drgn_error *drgn_platform_from_kdump(kdump_ctx_t *ctx,
 	return NULL;
 }
 
+static struct drgn_error *drgn_platform_to_kdump(kdump_ctx_t *ctx,
+						 const struct drgn_platform *platform)
+{
+	kdump_status ks;
+	char *arch_str = NULL;
+
+	if (platform->arch == &arch_info_x86_64)
+		arch_str = KDUMP_ARCH_X86_64;
+	else if (platform->arch == &arch_info_i386)
+		arch_str = KDUMP_ARCH_IA32;
+	else if (platform->arch == &arch_info_aarch64)
+		arch_str = KDUMP_ARCH_AARCH64;
+	else if (platform->arch == &arch_info_arm)
+		arch_str = KDUMP_ARCH_ARM;
+	else if (platform->arch == &arch_info_ppc64)
+		arch_str = KDUMP_ARCH_PPC64;
+	/* libkdumpfile doesn't support RISC-V */
+	else if (platform->arch == &arch_info_s390x)
+		arch_str = KDUMP_ARCH_S390X;
+	else if (platform->arch == &arch_info_s390)
+		arch_str = KDUMP_ARCH_S390;
+
+	if (arch_str) {
+		ks = kdump_set_string_attr(ctx, KDUMP_ATTR_ARCH_NAME, arch_str);
+		if (ks != KDUMP_OK) {
+			return drgn_error_format(DRGN_ERROR_OTHER,
+						 "kdump_set_string_attr(\"%s\"): %s",
+						 arch_str, kdump_get_err(ctx));
+		}
+	}
+
+	return NULL;
+}
+
 static struct drgn_error *drgn_read_kdump(void *buf, uint64_t address,
 					  size_t count, uint64_t offset,
 					  void *arg, bool physical)
@@ -91,6 +125,38 @@ struct drgn_error *drgn_program_set_kdump(struct drgn_program *prog)
 	if (!ctx) {
 		return drgn_error_create(DRGN_ERROR_OTHER,
 					 "kdump_new() failed");
+	}
+
+	/*
+	 * We need to be careful to set libkdumpfile attributes in the correct
+	 * order, in order to achive the desired result in the rare cases when
+	 * the program has the architecture and/or vmcoreinfo already set. In
+	 * these cases, frequently, the information is not available to
+	 * libkdumpfile, and so the user is providing it instead. We should set:
+	 *
+	 * - Architecture before file descriptor. This is because when
+	 *   libkdumpfile reads the vmcore, it immediately parses the notes in
+	 *   the header. If it doesn't know the current architecture, it will
+	 *   skip the architecture-specific notes, such as the PRSTATUS. This
+	 *   manifests to users as a missing "cpu.number" attribute, and thus a
+	 *   failure to get accurate stack traces for on-CPU tasks.
+	 * - Vmcoreinfo after architecture. This is because when libkdumpfile
+	 *   gets a new vmcoreinfo note set, it marks the address translation
+	 *   metadata as dirty and resets its address translation info using the
+	 *   new data. If no architecture is set, it skips this setup since it
+	 *   won't know how to do so. This manifests for users as FaultError
+	 *   when they access almost any memory address.
+	 *
+	 * For the common case, where we are just setting the FD and OS type,
+	 * and then letting libkdumpfile give us the vmcoreinfo and platform
+	 * info, the only ordering constraint is the obvious one: we need to set
+	 * the FD before getting data from libkdumpfile.
+	 */
+
+	if (prog->has_platform) {
+		err = drgn_platform_to_kdump(ctx, drgn_program_platform(prog));
+		if (err)
+			goto err;
 	}
 
 	ks = kdump_set_number_attr(ctx, KDUMP_ATTR_FILE_FD, prog->core_fd);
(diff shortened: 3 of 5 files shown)


iostapyshyn:
> The current known issue with this branch is that this won't work with ELF vmcores that drgn opens natively. Drgn expects to have virtual address mappings provided for it, at least for the linear/direct map. When they're not there, drgn will fail with a recursive page translation error:

The same issue exists with guest memory dumps from QEMU that don't include VM translations. Currently, drgn (without libkdumpfile) works out of the box with QEMU dumps generated with the paging flag (dump-guest-memory **-p**). When exporting in this way, QEMU performs page table walks and replicates the VM mapping via ELF segments. This appears to be slow and highly inefficient: In best case this takes about a minute and the resulting ELF has a little over 60000 segments, but in some cases the QEMU process hangs indefinitely (I suppose while the guest is executing a process with a huge address space).

The physical memory dumps (without paging) are *much faster* and work with drgn when `DRGN_USE_LIBKDUMPFILE_FOR_ELF=1`.

Without libkdumpfile, `read_memory_via_pgtable` starts from `swapper_pg_dir`, which is stored in VMCOREINFO as a virtual address. If the core dump doesn't have a translation for `swapper_pg_dir`, the page table walk will attempt to translate that address again via `read_memory_via_pgtable`, resulting in recursive address translation fault.

I propose applying [the kernel mapping offset](https://elixir.bootlin.com/linux/latest/source/arch/x86/include/asm/page_64.h#L37) to `swapper_pg_dir` in such cases, which I've implemented in [this commit](https://github.com/iostapyshyn/drgn/commit/2132454fd8494b0bac4c2ac7e1cdc01cc1aa9a40) for x86_64, but haven't checked other architectures yet. I'll gladly research a bit more to come up with a general solution though.

brenns10:
> QEMU dumps generated with the paging flag (dump-guest-memory -p)

Ooof, yeah, this option is heavily discouraged due to performance. Also, a malicious guest could cause a bit of a denial of service by finagling the page tables in just the right way so that the page mappings are way too large & complex for QEMU to handle.

But, drgn should _also_ work out-of-the-box with vmcores generated by `dump-guest-memory` _without -p_, so long as the host QEMU is started with `-device vmcoreinfo` and the guest Linux version is > 4.17 configured with `CONFIG_FW_CFG_SYSFS` and `CONFIG_KEXEC` (which are [pretty universally enabled](https://oracle.github.io/kconfigs/?config=FW_CFG_SYSFS&config=KEXEC&config=UTS_RELEASE)). See [this message](https://github.com/osandov/drgn/blob/main/libdrgn/program.c#L574-L580). If that's not the case, that's definitely a new bug to investigate. ((Note that `DRGN_USE_LIBKDUMPFILE_FOR_ELF=1` may still be required for such ELF vmcores.))

> I propose applying [the kernel mapping offset](https://elixir.bootlin.com/linux/latest/source/arch/x86/include/asm/page_64.h#L37) to swapper_pg_dir in such cases, which I've implemented in [this commit](https://github.com/iostapyshyn/drgn/commit/2132454fd8494b0bac4c2ac7e1cdc01cc1aa9a40) as a PoC for x86_64, but haven't checked other architectures yet. I'll gladly work on a general solution though.

Oops, I hope you didn't spend _too_ long on that since I had a similar (but less polished) workaround in my [vmcoreinfo_wip](https://github.com/brenns10/drgn/commit/0bea94451a875fe2f892d098002c74c0363b97c7) branch. Glad to know we came a similar solution :D

That said, I like your idea of moving the fixes into the page table iterator, that can allow the architecture code to be self-contained.

```diff
-			table_physical = false;
+			table_physical = true;
```

The above is from your changes to the x86_64 pgtable iterator, I think there's a risk that unconditionally setting `table_physical = true` here may result in incorrect behavior when reading userspace pages, since in that case the page table comes from the `pgd` field of `struct mm` which is always a virtual address. You could probably adjust this to only happen when `pgtable == swapper_pg_dir`, something like this:

```diff
diff --git a/libdrgn/arch_x86_64.c b/libdrgn/arch_x86_64.c
index a0a51c36..d7a2a4a4 100644
--- a/libdrgn/arch_x86_64.c
+++ b/libdrgn/arch_x86_64.c
@@ -616,7 +616,10 @@ linux_kernel_pgtable_iterator_next_x86_64(struct drgn_program *prog,
 	for (;; level--) {
 		uint64_t table;
 		bool table_physical;
-		if (level == levels) {
+		if (level == levels && it->it.pgtable == prog->vmcoreinfo.swapper_pg_dir) {
+			table = it->it.pgtable + prog->vmcoreinfo.phys_base - START_KERNEL_MAP;
+			table_physical = true;
+		} else if (level == levels) {
 			table = it->it.pgtable;
 			table_physical = false;
 		} else {
```

That way the changes are entirely contained within the arch-specific section of the page table code, and we can add similar code for other architectures as we see fit. I've pushed a new commit with this change and added a coauthor tag for you on it.

libdrgn/python/program.c
```diff
@@ -238,10 +239,13 @@ static void drgnpy_end_blocking(struct drgn_program *prog, void *arg, void *stat
 static Program *Program_new(PyTypeObject *subtype, PyObject *args,
 			    PyObject *kwds)
 {
-	static char *keywords[] = { "platform", NULL };
+	static char *keywords[] = { "platform", "vmcoreinfo", NULL };
 	PyObject *platform_obj = NULL;
-	if (!PyArg_ParseTupleAndKeywords(args, kwds, "|O:Program", keywords,
-					 &platform_obj))
+	const char *vmcoreinfo = NULL;
+	size_t vmcoreinfo_size;
```
osandov, line 245: This should technically be `Py_ssize_t`:

```suggestion
	Py_ssize_t vmcoreinfo_size;
```

_drgn.pyi
```diff
@@ -67,14 +67,21 @@ class Program:
     :meth:`[] <.__getitem__>` operator.
     """
 
-    def __init__(self, platform: Optional[Platform] = None) -> None:
+    def __init__(
+        self,
+        platform: Optional[Platform] = None,
+        vmcoreinfo: Union[bytes, str, None] = None,
```
osandov, line 73: Let's make this a keyword-only parameter.

drgn/cli.py
```diff
@@ -209,6 +209,21 @@ def _main() -> None:
         help="don't load any debugging symbols that were not explicitly added with -s",
     )
 
+    parser.add_argument(
```
osandov, line 212: As you mentioned in the commit message, we can put these in an argument group:

```python
    advanced_group = parser.add_argument_group("advanced")
    advanced_group.add_argument("--architecture", ...)
    advanced_group.add_argument("--vmcoreinfo", ...)
```

drgn/cli.py
```diff
@@ -209,6 +209,21 @@ def _main() -> None:
...
+        "-a",
```
osandov, line 213: I'd rather not use up short options for these niche options.
```suggestion
```

drgn/cli.py
```diff
@@ -209,6 +209,21 @@ def _main() -> None:
...
+        "--architecture",
+        metavar="ARCH",
+        choices=[a.name for a in drgn.Architecture]
+        + [a.name.lower() for a in drgn.Architecture],
+        help="set the program architecture, in case it can't be auto-detected",
+    )
+    parser.add_argument(
+        "-i",
```
osandov, line 221:
```suggestion
```

drgn/cli.py
```diff
@@ -209,6 +209,21 @@ def _main() -> None:
...
+        "--vmcoreinfo",
```
osandov, line 222: Let's add this for consistency with the other options taking paths:
```suggestion
        metavar="PATH",
```

drgn/cli.py
```diff
@@ -209,6 +209,21 @@ def _main() -> None:
...
+        type=str,
+        default=None,
```
osandov, line 224: I think ``None`` is the default ``default``.
```suggestion
```

drgn/cli.py
```diff
@@ -258,7 +273,16 @@ def _main() -> None:
     else:
         logger.setLevel(args.log_level.upper())
 
-    prog = drgn.Program()
+    platform = None
+    if args.architecture:
+        platform = drgn.Platform(drgn.Architecture[args.architecture.upper()])
+
+    vmcoreinfo = None
+    if args.vmcoreinfo:
```
osandov, line 281: Nit:
```suggestion
    if args.vmcoreinfo is not None:
```
(Not that an empty path is particularly meaningful, but I'd prefer that we get an error for that instead of silently ignoring it.)

libdrgn/kdump.c
```diff
@@ -138,6 +204,30 @@ struct drgn_error *drgn_program_set_kdump(struct drgn_program *prog)
...
+			free(vmcoreinfo);
+			err = &drgn_enomem;
+			goto err;
+		}
+		kdump_attr_t attr;
+		attr.type = KDUMP_BLOB;
+		attr.val.blob = blob;
+		ks = kdump_set_attr(ctx, "linux.vmcoreinfo.raw", &attr);
+		if (ks != KDUMP_OK) {
+			free(vmcoreinfo);
```
osandov, line 234: From my understanding of the libkdumpfile documentation, `blob` owns `vmcoreinfo` now, so this would cause `kdump_block_decref()` to double free, right?
```suggestion
```

libdrgn/drgn_program_parse_vmcoreinfo.inc.strswitch
```diff
@@ -92,6 +100,7 @@ struct drgn_error *drgn_program_parse_vmcoreinfo(struct drgn_program *prog,
 			if (err)
 				return err;
 			break;
+
```
osandov, line 103: Stray newline.
```suggestion
```

libdrgn/arch_x86_64.c
```diff
@@ -616,7 +618,15 @@ linux_kernel_pgtable_iterator_next_x86_64(struct drgn_program *prog,
 	for (;; level--) {
 		uint64_t table;
 		bool table_physical;
-		if (level == levels) {
+		if (level == levels && prog->vmcoreinfo.phys_base &&
+		    it->it.pgtable == prog->vmcoreinfo.swapper_pg_dir) {
+			// Avoid recursive address translation on swapper_pg_dir by
+			// directly resolving to a physical address. Don't do
+			// this if phys_base is 0, since that likely means it
+			// was not present in the vmcoreinfo.
```
osandov, line 626: Maybe add:

> It has been present since Linux kernel commit 401721ecd1dc ("kexec: export the value of phys_base instead of symbol address") (in v4.10).

libdrgn/arch_x86_64.c
```diff
@@ -18,6 +18,8 @@
 
 #include "arch_x86_64_defs.inc"
 
+#define START_KERNEL_MAP 0xffffffff80000000ULL
```
osandov, line 21: Use `UINT64_C` instead of `ULL`:
```suggestion
#define START_KERNEL_MAP UINT64_C(0xffffffff80000000)
```
Please also add a comment that this is `__START_KERNEL_map` in the kernel.

P.S. I'm amazed that this has never changed! I tracked it down to this commit in `history.git`: https://git.kernel.org/pub/scm/linux/kernel/git/history/history.git/commit/?id=0457d99a336be658cea1a5bdb689de5adb3b382d.

libdrgn/program.h
```diff
@@ -169,6 +169,8 @@ struct drgn_program {
 		uint64_t va_bits;
 		/** Whether 5-level paging was enabled on x86-64. */
 		bool pgtable_l5_enabled;
+		/** phys_base on x86_64 */
+		uint64_t phys_base;
```
osandov, line 173: Put this before `pgtable_l5_enabled` for better struct packing.

osandov (changes requested):
Thanks for this! I kind of hate that this is necessary, but that's not your fault, and I'm happy with how you handled it. A few minor comments throughout. The last commit also needs a commit message.

libdrgn/arch_x86_64.c
brenns10: Ah I really should have looked that up myself, I got lazy. I noticed that there was a `SYMBOL(phys_base)` in 4.9 and found that really funny: what a chicken & egg scenario! I'll add that comment.

libdrgn/arch_x86_64.c
brenns10: I also am amazed! It's one of those things I feel dirty hardcoding, but with a solid history like that, I guess I feel a little bit better :)

libdrgn/kdump.c
brenns10: Yeah, you're definitely correct, and it really obviously looks  wrong too. Good catch.

More interestingly however is that the libkdumpfile code seems to indicate that even in the error case, the reference to the blob is already stolen, and thus we cannot free it in our error path. See [this comment](https://github.com/ptesarik/libkdumpfile/blob/tip/src/kdumpfile/attr.c#L898-L903). However, there are clearly cases of errors inside of `kdump_set_attr()` where the function returns and the blob is not decref'd. In summary: on some failures, we can't even decref the blob. On other failures, we should. And there's no way to tell 😜 

So my vote here (I know it's gross) is to heavily comment this and get rid of both the `free(vmcoreinfo)` and the `kdump_blob_decref()`.

brenns10: Ok, I filed https://github.com/ptesarik/libkdumpfile/issues/81 for Petr, we'll see if he agrees too. Assuming he does, I'll remove the error handling and document the leak .

osandov: If `kdump_set_attr()` is supposed to take ownership, then yes, we should definitely remove `kdump_blob_decref()`, and I don't think we even need a comment; we just need to report any leaks as a libkdumpfile bug.

brenns10: Ok, dropped the comment and the `kdump_blob_decref()`, we can revisit it if Petr says the ownership semantic is supposed to be different.

libdrgn/program.c
```diff
@@ -264,6 +264,7 @@ drgn_program_set_core_dump_fd_internal(struct drgn_program *prog, int fd,
 	const char *vmcoreinfo_note = NULL;
 	size_t vmcoreinfo_size = 0;
 	bool have_nt_taskstruct = false, is_proc_kcore;
+	bool have_vmcoreinfo = prog->vmcoreinfo.raw;
```
osandov, line 267: I just noticed that `drgn_program_set_core_dump_fd_internal()` and `drgn_program_set_kdump()` both free `prog->vmcoreinfo.raw` in the error case, so the error handling needs to be updated to handle the case that the vmcoreinfo was already set.

osandov (changes requested):
A couple more things I noticed in this round.

libdrgn/program.c
brenns10: Looks like we're getting `have_vmcoreinfo` and `had_vmcoreinfo` 🙃

libdrgn/program.c
```diff
@@ -600,7 +603,9 @@ drgn_program_set_core_dump_fd_internal(struct drgn_program *prog, int fd,
 out_notes:
 	// Reset anything we parsed from ELF notes.
 	prog->aarch64_insn_pac_mask = 0;
-	free(prog->vmcoreinfo.raw);
+	// Free vmcoreinfo buffer if it was not provided by the caller
+	if (!had_vmcoreinfo)
+		free(prog->vmcoreinfo.raw);
 	memset(&prog->vmcoreinfo, 0, sizeof(prog->vmcoreinfo));
```
brenns10, line 609: Oooh wow my mistake, should have re-read it more carefully. We need to preserve `prog->vmcoreinfo` because it won't get parsed again.

osandov (approved):

PR #398, opened 2024-05-28
contrib/bpf_inspect.py: add interactive mode

Asphaltt:
It's convenient to run drgn shell with bpf helpers in bpf_inspect.py.

### DIFF
--- a/contrib/bpf_inspect.py
+++ b/contrib/bpf_inspect.py
@@ -4,6 +4,7 @@
 
 """List BPF programs or maps and their properties unavailable via kernel API."""
 
+import sys
 import argparse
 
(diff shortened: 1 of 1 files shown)


osandov (changes requested):
Thanks for this! Maybe this is a sign that some of these should be adapted into drgn helpers, but this is fine for now. One minor comment.

Asphaltt:
> Maybe this is a sign that some of these should be adapted into drgn helpers, but this is fine for now.

These functions are too simple for me to inspect details of bpf progs and bpf maps. Therefore, I'm planning to improve them.

osandov (approved):

PR #400, opened 2024-05-28
Fix CI failure on Python 3.12

brenns10:
Our setup.py imports setuptools which is no longer provided by Python 3.12. It must be installed manually via pip. We did not notice this because the "nodeenv" package listed setuptools in its install_requires, thus providing the dependency for us.

Four hours ago nodeenv 1.9.0 was released, and it happened to drop the setuptools dependency. This broke CI. Add an explicit setuptools requirement in our pip command to fix it.


brenns10:
The 3.12 failure can be seen in [this build](https://github.com/osandov/drgn/actions/runs/9277262509/job/25526067124?pr=396). The 3.6 failure for that branch is for a genuine issue with my change, but the 3.12 issue cropped up by chance at the same time. I root-caused it by comparing to this recent [build](https://github.com/osandov/drgn/actions/runs/9275276182/job/25519580192) which installed setuptools as a dependency of nodeenv, while my build no longer did.

osandov (approved):
Thanks for taking the time to track down the cause!

PR #401, opened 2024-05-29
contrib/bpf_inspect.py: hide args from interactive mode

Asphaltt:
When run `list_bpf_progs()` in drgn shell, it's meaningless to pass `args` to `list_bpf_progs()`. Therefore, let's hide `args` from drgn shell.

BTW, select appropriate global variables and functions to run drgn shell.

So, it's ready to add more bpf functions to inspect bpf progs, bpf maps, bpf links and so on.


osandov (approved):
Looks reasonable, thanks.

PR #405, opened 2024-06-18
Update trove classifiers

brenns10:
I noticed it still said "beta" which struck me as odd! Added in the Linux kernel classifier too. Sorry for a tiny PR but I figured it could be just in time for a 0.0.27.
https://pypi.org/classifiers/


osandov:
Maybe someday we'll even have drgn version 1.0.0!

PR #407, opened 2024-06-24
scripts/build_manylinux_in_docker.sh: clean build dir between builds

brenns10:
Currently the python 3.13 manylinux build of drgn fails with a segfault. The reason is that the build system doesn't know not to share a build directory between cp313-cp313, and cp313-cp313t. Whichever is built second will reuse object files from the first build, which are incompatible. The current result of this is a segmentation fault:

```
+ /opt/python/cp313-cp313t/bin/drgn --version 
/io/scripts/build_manylinux_in_docker.sh: line 101: 43331 Segmentation fault
```

Adding insult to injury, the segmentation fault disappears when you try to build only one version in order to save time while debugging.

There may be a better fix for this particular issue, but deleting the build directory doesn't cause any problems and will ensure this doesn't pop up again in other cases.


brenns10:
Closing this in favor of #410

PR #410, opened 2024-06-25
setup.py: use a separate build directory for free-threading python

brenns10:
Ok, so yesterday's solution #407 to the 3.13 wheel build failure correctly got at the root cause, but it didn't really address the issue outside of `scripts/build_manylinux_in_docker.sh`.

This solution actually will update the `build_temp` directory to contain a "t" suffix on free-threading builds, which means that we avoid issues of sharing build directories both in the wheel build environment, and in normal development.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -11,6 +11,7 @@ import re
 import shlex
 import subprocess
 import sys
+import sysconfig
 
 from setuptools import Command, find_packages, setup
 from setuptools.command.build_ext import build_ext as _build_ext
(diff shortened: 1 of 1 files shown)


brenns10:
To expand a bit on this, distutils sets the build_temp based on sys.implementation.cache_tag:
https://github.com/pypa/setuptools/blob/222ebf921a4a1cca69d6a5193121f3090580e502/setuptools/_distutils/command/build.py#L82

As far as I can tell, I think this should be fixed in distutils too, though drgn's incremental builds makes it particularly vulnerable to the issue. This approach here should resolve the build and development situation in the meantime.

brenns10:
I just filed a related PR, pypa/distutils#260, which would address this in distutils too.

osandov:
I'm happy with this approach, thank you! I agree that it'd be nice if distutils handled this itself. There's a check in the same spirit for pydebug, so it seems like a reasonable request. I guess if distutils does add that, we'd be doing it twice, but that's purely cosmetic.

brenns10:
Yeah, I was thinking about the `313tt` thing, but honestly, if that happens we can fix it up, and it is just cosmetic.

PR #411, opened 2024-07-03
Add linux kernel module helpers

brenns10:
There are a few common use cases that I find myself re-writing: iterating over the list of modules, looking up a module by name, and getting memory addresses related to modules (or vice versa). Add helpers for each.

---

I know there is the _drgn_ module API which is upcoming, and I don't necessarily want to confuse or overlap with that. So maybe it's possible this will duplicate that API in some form, but I find myself needing to access kernel modules or look them up by name/address rather frequently. So here's some code and tests (tested on 6.9 and 4.19, hoping to let CI take a look at the rest :wink: ).

### DIFF
--- a/drgn/helpers/linux/module.py
+++ b/drgn/helpers/linux/module.py
@@ -1,0 +1,223 @@
+# Copyright (c) 2024 Oracle and/or its affiliates
+# SPDX-License-Identifier: LGPL-2.1-or-later
+"""
+Modules
+-------
+
+The ``drgn.helpers.linux.module`` module contains helpers for working with
+loaded kernel modules.
+"""
+import operator
+from typing import Iterable, List, Tuple, Union
+
+from drgn import NULL, IntegerLike, Object, Program
+from drgn.helpers.common.prog import takes_program_or_default
+from drgn.helpers.linux.list import list_for_each_entry
+from drgn.helpers.linux.rbtree import rb_find
+
+__all__ = (
+    "address_to_module",
+    "find_module",
+    "for_each_module",
+    "module_address_regions",
+    "module_percpu_region",
+)
+
+
+@takes_program_or_default
+def for_each_module(prog: Program) -> Iterable[Object]:
+    """
+    Returns all loaded kernel modules
+
+    :returns: Iterable of ``struct module *`` objects
+    """
+    return list_for_each_entry("struct module", prog["modules"].address_of_(), "list")
+
+
+@takes_program_or_default
+def find_module(prog: Program, name: Union[str, bytes]) -> Object:
+    """
+    Lookup a kernel module by name, or return NULL if not found
+
+    :param name: name to search for
+    :returns: the ``struct module *`` by that name, or NULL
+    """
+    if isinstance(name, str):
+        name = name.encode()
+    for module in for_each_module(prog):
+        if module.name.string_() == name:
+            return module
+    return NULL(prog, "struct module *")
+
+
+def module_percpu_region(mod: Object) -> Tuple[int, int]:
+    """
+    Lookup the percpu memory region of a module.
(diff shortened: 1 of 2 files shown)


brenns10:
So, I'm not sure how the module API issue 332 got linked to this pull request... I did not mention it by number? I mentioned the title though. Is that enough to accidentally link issues and pull requests? If so... booo.

drgn/helpers/linux/module.py
```diff
@@ -0,0 +1,228 @@
...
+    """
+    Returns all loaded kernel modules
+
+    :returns: Iterable of ``struct module *`` objects
+    """
+    return list_for_each_entry("struct module", prog["modules"].address_of_(), "list")
+
+
+@takes_program_or_default
+def lookup_module(prog: Program, name: Union[str, bytes]) -> Object:
```
osandov, line 37: The equivalent kernel function is named [`find_module()`](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/include/linux/module.h?h=v6.10-rc7#n657), so let's use that name.

drgn/helpers/linux/module.py
```diff
@@ -0,0 +1,228 @@
...
+        cmp,
+    )
+    if mem:
+        return mem.mtn.mod
+    else:
+        return NULL(prog, "struct module *")
+
+
+@takes_program_or_default
+def address_to_module(prog: Program, addr: IntegerLike) -> Object:
```
osandov, line 193: The equivalent kernel function is [`__module_address()`](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/include/linux/module.h?h=v6.10-rc7#n614), but that name doesn't make much sense to me, so this is fine.

drgn/helpers/linux/module.py
```diff
@@ -0,0 +1,228 @@
...
+    # in use depends on the seqcount, which gets incremented for each
+    # modification. This is a really neat approach that allows reads in parallel
+    # with a writer. In our use case, it's probably not worth verifying the
+    # seqcount after the fact. What we do need is the index (0 or 1). This may
+    # be a seqcount_latch_t, or before 24bf401cebfd6 ("rbtree_latch: Use
+    # seqcount_latch_t"), a regular seqcount_t.
+    try:
+        idx = int(mod_tree.root.seq.seqcount.sequence & 1)
+    except AttributeError:
+        idx = int(mod_tree.root.seq.sequence & 1)
```
osandov, line 153: I have a slight preference for:
```suggestion
        idx = mod_tree.root.seq.seqcount.sequence.value_() & 1
    except AttributeError:
        idx = mod_tree.root.seq.sequence.value_() & 1
```

osandov (changes requested):
Thanks, I agree that this still makes sense even with the module API. Some minor comments. (I unlinked the issue, maybe it was because your branch had the same name as the branch I had previously linked to it? That's weird.)

osandov:
One more thing, the test case is failing for me with !SMP:

```
$ python3 setup.py test -k '*tiny'
...
FAIL: test_module_percpu_region (tests.linux_kernel.helpers.test_module.TestModules.test_module_percpu_region)
  File "/home/osandov/repos/drgn3/tests/linux_kernel/helpers/test_module.py", line 32, in test_module_percpu_region
    self.assertTrue(start <= pcpu_addr <= start + size)
AssertionError: False is not true
```

brenns10:
Huh, I guess I was getting too clever about that test. On !SMP the start should be 0, i.e. NULL, so I will probably only conditionally assert if `start != 0`. Since it's still a nice test of functionality for SMP kernels.

osandov (approved):

PR #408, opened 2024-06-25
contrib/bpf_inspect.py: show used maps and subprogs for bpf prog

Asphaltt:
Show used maps and subprogs for bpf prog.

```bash
$ sudo drgn ./contrib/bpf_inspect.py p --show-details
     2: BPF_PROG_TYPE_TRACING            hid_tail_call
	used map:      2: BPF_MAP_TYPE_PROG_ARRAY          hid_jmp_table
  2295: BPF_PROG_TYPE_EXT                entry                            tail_call_reachable
	used map:   4208: BPF_MAP_TYPE_ARRAY               tailcall.bss
	used map:   4207: BPF_MAP_TYPE_PROG_ARRAY          jmp_table
	func[ 0]:      0: BPF_PROG_TYPE_EXT                entry                            tail_call_reachable
	func[ 1]:      0: BPF_PROG_TYPE_EXT                subprog_tail
  2400: BPF_PROG_TYPE_EXT                entry                            tail_call_reachable
	used map:   4248: BPF_MAP_TYPE_ARRAY               tailcall.bss
	used map:   4247: BPF_MAP_TYPE_PROG_ARRAY          jmp_table
	func[ 0]:      0: BPF_PROG_TYPE_EXT                entry                            tail_call_reachable
	func[ 1]:      0: BPF_PROG_TYPE_EXT                subprog_tail
  4235: BPF_PROG_TYPE_SCHED_CLS          entry1
	used map:   5092: BPF_MAP_TYPE_ARRAY               .rodata
	func[ 0]:      0: BPF_PROG_TYPE_SCHED_CLS          entry1
	func[ 1]:      0: BPF_PROG_TYPE_SCHED_CLS          subprog1
	func[ 2]:      0: BPF_PROG_TYPE_SCHED_CLS          subprog2
```

Signed-off-by: Leon Hwang <hffilwlqm@gmail.com>

PR #412, opened 2024-07-09
contrib: bpf_inspect: adjust syntax for python < 3.12

brenns10:
Hi, this is a totally optional PR, but I noticed that `contrib/bpf_inspect.py` uses nested f-string quotes that are only available in Python 3.12 or later. It would be nice to be able to run this on older Python versions. If @Asphaltt doesn't mind it's a simple tweak to make these work on the older versions.

### DIFF
--- a/contrib/bpf_inspect.py
+++ b/contrib/bpf_inspect.py
@@ -200,10 +200,10 @@ def list_bpf_progs(show_details=False):
                 print(f"\tlinked: {BpfProg(linked_prog)}")
 
         for map_ in bpf_prog.get_used_maps():
-            print(f"\t{"used map:":9} {map_}")
+            print(f"\t{'used map:':9} {map_}")
 
         for index, subprog in bpf_prog.get_subprogs():
-            print(f"\t{f"func[{index:>2}]:":9} {subprog}")
+            print(f"\t{f'func[{index:>2}]:':9} {subprog}")
 
 
 def __list_bpf_progs(args):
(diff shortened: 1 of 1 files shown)


Asphaltt:
May you fix it after merging #408 ? Because it uses nested f-string quotes too.

brenns10:
That makes perfect sense! I'm happy to wait, I just wanted to post the PR since it was a patch I needed to include in my RPM builds for drgn on python 3.6-based Oracle Linux versions.

osandov:
Ok, #408 is merged. Even though we don't want to enforce style guidelines on `contrib`, I wonder if we should at least check the syntax with `vermin`.

brenns10:
>  Even though we don't want to enforce style guidelines on contrib, I wonder if we should at least check the syntax with vermin.

Maybe, yeah. But I don't necessarily think that the contrib directory needs to have the same Python compatibility guarantees (3.6+) as the rest of drgn -- part of the goal is to lower the barrier to entry. Expecting compatibility with old (EOL...) python versions is exactly one of those barriers that we would probably like to avoid here?

I wasn't trying to require that this file or any other, e.g., be 3.6-compatible, but rather helpfully come around after the fact to make easy changes to broaden compatibility, since it wasn't intrusive.

As a distributor I'm happy to either downstream patch, or flat-out remove files from the contrib directory if they're not compatible with the Python I'm shipping on. It's kind of my job to make sure these things work :)

osandov:
Yeah, I was hoping to avoid any requirements beyond the bare minimum for `contrib`, so I'm fine keeping it that way and accepting fixes after the fact.

PR #413, opened 2024-07-12
tests/linux_kernel: Skip TestStackDepot unless test kmod is present

sourabhjains:
TestStackDeport needs the drgn test kernel module, so if it is not present, skip the test.

Error reported:

```
ERROR: setUpClass (tests.linux_kernel.helpers.test_stackdepot.TestStackDepot)
  File "/root/drgn/tests/linux_kernel/helpers/test_stackdepot.py", line 15, in setUpClass
    if not cls.prog["drgn_test_have_stackdepot"]:
KeyError: 'drgn_test_have_stackdepot'

Ran 980 tests in 4.056s
```


osandov:
Good catch, thank you.

sourabhjains:
Thanks for accepting the patch.

Issue #415, opened 2024-07-17
Test failure on 6.1.98-vmtest30.0tiny


brenns10:
I'm getting this on the `main` branch for `6.1.98-vmtest30.0tiny`:
```
======================================================================
ERROR: test_identify_slab_cache (tests.linux_kernel.helpers.test_common.TestIdentifyAddress) (size='small')
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/stepbren/repos/drgn/tests/linux_kernel/helpers/test_common.py", line 45, in test_identify_slab_cache
    self.assertIsNone(identify_address(obj))
  File "/home/stepbren/repos/drgn/drgn/helpers/common/prog.py", line 113, in wrapper
    return f(args[0].prog_, *args, **kwds)
  File "/home/stepbren/repos/drgn/drgn/helpers/common/memory.py", line 198, in identify_address
    return _identify_kernel_address(prog, addr, cache)
  File "/home/stepbren/repos/drgn/drgn/helpers/common/memory.py", line 132, in _identify_kernel_address
    slab_info = _get_slab_cache_helper(slab.slab_cache).object_info(
AttributeError: 'struct slab' has no member 'slab_cache'

======================================================================
ERROR: test_print_annotated_memory (tests.linux_kernel.helpers.test_common.TestPrintAnnotatedMemory)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/stepbren/repos/drgn/tests/linux_kernel/helpers/test_common.py", line 90, in test_print_annotated_memory
    print_annotated_memory(
  File "/home/stepbren/repos/drgn/drgn/helpers/common/prog.py", line 111, in wrapper
    return f(*args, **kwds)
  File "/home/stepbren/repos/drgn/drgn/helpers/common/memory.py", line 255, in print_annotated_memory
    identified = identify_address(prog, value, cache=cache)
  File "/home/stepbren/repos/drgn/drgn/helpers/common/prog.py", line 111, in wrapper
    return f(*args, **kwds)
  File "/home/stepbren/repos/drgn/drgn/helpers/common/memory.py", line 198, in identify_address
    return _identify_kernel_address(prog, addr, cache)
  File "/home/stepbren/repos/drgn/drgn/helpers/common/memory.py", line 132, in _identify_kernel_address
    slab_info = _get_slab_cache_helper(slab.slab_cache).object_info(
AttributeError: 'struct slab' has no member 'slab_cache'

======================================================================
[...]

PR #416, opened 2024-07-18
Fix test failures on SLOB / tiny

brenns10:
I also added a `-f` option to the `setup.py` because I wanted to run the tests locally and noticed I had no easy way to say "just run the tiny tests". I didn't bother trying to add any logic to make `-f` and `-F` conflict... it's a developer tool after all.

### DIFF
--- a/drgn/helpers/linux/mm.py
+++ b/drgn/helpers/linux/mm.py
@@ -1377,3 +1377,15 @@ def totalram_pages(prog: Program) -> int:
         return prog["_totalram_pages"].counter.value_()
     except KeyError:
         return prog["totalram_pages"].value_()
+
+
+@takes_program_or_default
+def in_direct_map(prog: Program, addr: IntegerLike) -> bool:
+    """
+    Return True if an address is within the kernel's direct memory mapping
+    :param addr: address to check
+    """
+    addr = operator.index(addr)
+    start_addr = pfn_to_virt(prog["min_low_pfn"]).value_()
+    end_addr = (pfn_to_virt(prog["max_low_pfn"]) + prog["PAGE_SIZE"]).value_()
+    return start_addr <= addr < end_addr
--- a/drgn/helpers/common/memory.py
+++ b/drgn/helpers/common/memory.py
@@ -14,26 +14,12 @@ import typing
 from typing import Any, Dict, Optional
 
 import drgn
-from drgn import (
-    FaultError,
-    IntegerLike,
-    Object,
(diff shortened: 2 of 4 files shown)


brenns10:
Oh, boy, also, Fixes #415, which I did not include in a commit message, but that's probably ok.

brenns10:
A manual test run on my fork with the "test-all-kernel-flavors" option set: (but not all flavors)
https://github.com/brenns10/drgn/actions/runs/9983302801

brenns10:
This implementation doesn't run all kernel flavors on push to `main` (like we do with all Python versions). It only does it for the tests after vmtest build, and when a PR has the tag set.

brenns10:
Oh fun, no space left on device! Looks like we'll need to avoid pre-fetching _all_ kernels and we'll need a way to delete them after testing, for CI only:
https://github.com/brenns10/drgn/actions/runs/9983302801

osandov:
Would you mind splitting the setup.py and CI changes into a separate PR? I'll take a closer look tomorrow, thanks again!

brenns10:
Yeah that seems wise. I'll split them up in the morning!

drgn/helpers/linux/mm.py
```diff
@@ -1377,3 +1377,20 @@ def totalram_pages(prog: Program) -> int:
...
+    """
+    Return True if an address is within the kernel's direct memory mapping
+    :param addr: address to check
+    """
+    addr = operator.index(addr)
+    try:
+        start_addr = pfn_to_virt(prog["min_low_pfn"]).value_()
+        end_addr = (pfn_to_virt(prog["max_low_pfn"]) + prog["PAGE_SIZE"]).value_()
+        return start_addr <= addr < end_addr
+    except NotImplementedError:
```
osandov, line 1393: Don't bother catching this exception. The helper should fail hard if virtual address translation isn't supported like the other mm helpers.

osandov (changes requested):
Looks good other than one comment, thanks!

drgn/helpers/linux/mm.py
brenns10: Would you like this helper's callers to fail on this as well, or should I push that exception handler up to each of the three callers?

osandov: Let's keep it the way it currently is: `_identify_kernel_address()` should catch it, but the SLAB helpers should fail.

osandov (approved):

PR #417, opened 2024-07-18
Add option to run CI on all kernel flavors, plus other improvements

brenns10:
This branch adds three improvements related to testing different kernel flavors:

1. Allows locally running tests against all kernel versions of a given flavor by `python setup.py test -K -f flavor`. This is just a convenience that I wanted.
2. Adds a "test-all-kernel-flavors" option similar to the "test-all-python-versions" option. Its main difference is that we don't test all flavors on push to the `main` branch (whereas we do test all Python versions for every push to `main`). I think that makes sense, because pushes to main can be rather frequent, and testing each flavor triples the test burden (which is of course multiplicative with all the Python versions). The "test-all-kernel-flavors" option is active after each vmtest build, or when the label is applied to a PR, or for a manual run on the Github Actions UI.
3. Finally, to avoid filling up the disk space with downloaded kernels, I was forced to add some Github Actions specific logic to the testing, in order to cleanup downloaded kernels. We delete kernels after we test them, and we also set a size limit on the queue used to communicate downloaded kernels back to the main test thread. This ensures that at a certain point, the download thread pauses until a message is removed from the queue, and thus we never have more than a finite limit of kernels on-disk. This is important because of course the download speed is frequently faster than running the tests themselves.

I tested part 3 locally and verified that when `GITHUB_ACTIONS=true`, my `build/vmtest` directory never grew beyond ~2.5 GiB for a standard `-K` test, and it was emptied of kernels by the end of the run. I didn't want to do the `-K -F` locally, but I'd assume the result is about the same. For a local `-K` test with `GITHUB_ACTIONS=false`, the `build/vmtest` directory reached 7.5 GiB quickly and never decreased.

Test runs in Github Actions:

1. A test run which includes the changes of #416: https://github.com/brenns10/drgn/actions/runs/9998854777
2. A test run which does not include the changes of #416 and thus is expected to fail: https://github.com/brenns10/drgn/actions/runs/9998964897

### DIFF
--- a/vmtest/download.py
+++ b/vmtest/download.py
@@ -284,9 +284,11 @@ def _download_thread(
 
 @contextmanager
 def download_in_thread(
-    download_dir: Path, downloads: Iterable[Download]
+    download_dir: Path, downloads: Iterable[Download], max_pending_kernels: int = 0
 ) -> Generator[Iterator[Downloaded], None, None]:
-    q: "queue.Queue[Union[Downloaded, Exception]]" = queue.Queue()
+    q: "queue.Queue[Union[Downloaded, Exception]]" = queue.Queue(
+        maxsize=max_pending_kernels
+    )
 
     def aux() -> Iterator[Downloaded]:
         while True:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,10 +11,20 @@ on:
         type: boolean
         default: false
         required: true
+      test_all_kernel_flavors:
+        description: "Run tests on all kernel flavors"
+        type: boolean
+        default: false
+        required: true
   workflow_call:
     inputs:
       test_all_python_versions:
         description: "Run tests on all Python versions"
+        type: boolean
+        default: false
+        required: true
+      test_all_kernel_flavors:
+        description: "Run tests on all kernel flavors"
         type: boolean
         default: false
         required: true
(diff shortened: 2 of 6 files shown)


vmtest/download.py
```diff
@@ -286,7 +288,12 @@ def _download_thread(
 def download_in_thread(
     download_dir: Path, downloads: Iterable[Download]
 ) -> Generator[Iterator[Downloaded], None, None]:
-    q: "queue.Queue[Union[Downloaded, Exception]]" = queue.Queue()
+    # Downloading too many files before they can be used for testing runs the
+    # risk of filling up the limited disk space is Github Actions. Set a limit
+    # of no more than 5 files which can be downloaded ahead of time. This is a
+    # magic number with no testing, but should work.
+    maxsize = 5 if IN_GITHUB_ACTIONS else 0
```
osandov, line 295: This would be cleaner as an argument passed to `download_in_thread()`. Other than that, this is a really clever solution.

osandov (changes requested):
This is awesome, thank you! One minor comment below.

vmtest/download.py
brenns10: Thanks! Would you like `vmtest/__main__.py` to pass this parameter based on `GITHUB_ACTIONS` environment variable? There are two callers, but of course `setup.py` is the only one used in CI. But in case we switch over to `vmtest/__main__.py` we may want the logic in both places.

osandov: Good catch, yeah, just copy the logic in both places. I have a dream of getting off setuptools entirely, and `vmtest/__main__.py` will replace this if that ever happens.

osandov (approved):

osandov:
I was hoping that adding the label to this PR would do a test run, but maybe since it's not in the main branch yet, it didn't. Oh well, this looks good!

PR #418, opened 2024-07-19
drgn.internal.sudohelper: pass privileged command as string instead of path

osandov:
At Meta, our Python executables are packaged in a way that doesn't leave the original .py or .pyc files easily accessible or importable. This makes the sudohelper fail with:

```
  can't open file '/proc/self/fd/3/drgn/internal/sudohelper.pyc': [Errno 2] No such file or directory
```

We can work around this by passing the helper command as a string literal. It's ugly, but it works with our Meta package and with traditional installations.

@brenns10 would you mind testing this out on your systems?

### DIFF
--- a/drgn/internal/sudohelper.py
+++ b/drgn/internal/sudohelper.py
@@ -3,7 +3,6 @@
 
 """Helper for opening a file as root and transmitting it via unix socket"""
 import array
-import os
 from pathlib import Path
 import pickle
 import socket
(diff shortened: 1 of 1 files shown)


brenns10 (approved):
Tested this in development on my laptop, and also built an RPM and tried it out with that. Worked fine for me. It's a funny workaround but also in a way simpler :)

PR #423, opened 2024-08-07
Cgroup scripts

Werkov:
This is based on discussion around [cgroup.stat](https://lore.kernel.org/all/20240711025153.2356213-1-longman@redhat.com/) and as an exercise with drgn, I'm sharing it in the case anyone found that useful.
There is also a little change outside of `contrib/`.

BTW How is that with drgn and RCU protected data? Is it simply neglected and one should be ready for transient issues when dereferencing gone data (that's actually not specific to RCU) when debugging a running system?

### DIFF
--- a/drgn/helpers/linux/cgroup.py
+++ b/drgn/helpers/linux/cgroup.py
@@ -159,13 +159,12 @@ def _css_for_each_impl(
         pos = next_fn(pos, css)
         if not pos:
             break
-        if pos.flags & pos.prog_["CSS_ONLINE"]:
-            yield pos
+        yield pos
 
 
 def css_for_each_child(css: Object) -> Iterator[Object]:
     """
-    Iterate through children of the given css.
+    Iterate through children (offline included) of the given css.
 
     :param css: ``struct cgroup_subsys_state *``
     :return: Iterator of ``struct cgroup_subsys_state *`` objects.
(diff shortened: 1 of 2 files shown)


drgn/helpers/linux/cgroup.py
```diff
@@ -152,14 +152,14 @@ def css_next_descendant_pre(pos: Object, root: Object) -> Object:
...
-    next_fn: Callable[[Object, Object], Object], css: Object
+    next_fn: Callable[[Object, Object], Object], css: Object, only_online: bool
 ) -> Iterator[Object]:
     pos = NULL(css.prog_, "struct cgroup_subsys_state *")
     while True:
         pos = next_fn(pos, css)
         if not pos:
             break
-        if pos.flags & pos.prog_["CSS_ONLINE"]:
+        if not only_online or pos.flags & pos.prog_["CSS_ONLINE"]:
```
osandov, line 162: If I'm reading it correctly, the kernel code doesn't limit `css_for_each_child()` or `css_for_each_descendant_pre()` to only online css, is that right? If so, then I'd prefer to just remove the `CSS_ONLINE` check entirely, since it's better for helpers to be as similar as possible to the kernel functions they are based on.

osandov (commented):
This is great, thank you. I have one question about the helper change.

drgn/helpers/linux/cgroup.py
Werkov: Moved the onlineness test from helper (to mimic kernel behavior) to selected users.

Werkov:
Force pushed a change addressing the review.

osandov:
Thanks! I fixed up a stray revert message left in the first commit message.

PR #426, opened 2024-08-14
helpers.linux.fs: add dentry_path_first_mount()

brenns10:
This is a bit of a drive-by helper idea, inspired by something we have in drgn-tools. We frequently want a way to just convert a dentry to a path even if we don't know the vfsmount associated with the path we want. This helper is a good way to do it. Let me know your thoughts!

### DIFF
--- a/drgn/helpers/linux/fs.py
+++ b/drgn/helpers/linux/fs.py
@@ -167,16 +167,54 @@ def d_path(vfsmnt: Object, dentry: Object) -> bytes:
     ...
 
 
+@overload
+def d_path(dentry: Object) -> bytes:
+    """
+    Return the full path of a dentry.
+
+    Since a mount is not provided, this arbitrarily selects a mount to determine
+    the path.
+
+    :param dentry: ``struct dentry *``
+    """
+    ...
+
+
 def d_path(  # type: ignore  # Need positional-only arguments.
-    path_or_vfsmnt: Object, dentry: Optional[Object] = None
+    arg1: Object, arg2: Optional[Object] = None
 ) -> bytes:
-    if dentry is None:
-        vfsmnt = path_or_vfsmnt.mnt
-        dentry = path_or_vfsmnt.dentry.read_()
+    if arg2 is None:
+        try:
+            mnt = container_of(arg1.mnt, "struct mount", "mnt")
+            dentry = arg1.dentry.read_()
+        except AttributeError:
+            # Select an arbitrary mount from this dentry's super block. We
+            # choose the first non-internal mount. Internal mounts exist for
+            # kernel filesystems (e.g. debugfs) and they are mounted at "/".
+            # Paths from these mounts aren't usable in userspace and they're
+            # confusing. If there's no other option, we will use the first
+            # internal mount we encountered.
+            #
+            # The MNT_INTERNAL flag is defined as a macro in the kernel source.
+            # Introduced in 2.6.34 and has not been modified since.
+            MNT_INTERNAL = 0x4000
+            internal_mnt = None
+            dentry = arg1
+            for mnt in list_for_each_entry(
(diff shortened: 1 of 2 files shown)


drgn/helpers/linux/fs.py
```diff
@@ -219,6 +219,32 @@ def dentry_path(dentry: Object) -> bytes:
...
+    or most "expected" mount point, as opposed to bind-mounts or mounts from
+    non-init filesystem namespaces. In short, this is very likely to get you a
+    useful filesystem path for diagnostics, without needing to have a ``struct
+    vfsmount *``.
+
+    :param dentry: ``struct dentry *``
+    """
+    # Defined as a macro in the kernel source. Introduced in 2.6.34 and has not
+    # been modified since.
+    MNT_INTERNAL = 0x4000
```
osandov, line 238: Can you remind me what `MNT_INTERNAL` means?

osandov (commented):
I love the idea, but I'm not a fan of the name. One option would be to overload `d_path` even more so that it can take a `dentry` with no mount:

```python3
def d_path(  # type: ignore  # Need positional-only arguments.
    arg1: Object, arg2: Optional[Object] = None
) -> bytes:
    if arg2 is None:
        try:
            mnt = container_of(arg1.mnt, "struct mount", "mnt")
        except AttributeError:
            dentry = arg1
            mnt = ... your code to find the first mount
        else:
            dentry = arg1.dentry.read_()
    else:
        mnt = container_of(arg1, "struct mount", "mnt")
        dentry = arg2.read_()
```

(And in that case, maybe underspecify which mount we choose in case we want a different heuristic later.)

What do you think?

drgn/helpers/linux/fs.py
brenns10: That's actually a really good question. I know that before I added this check, there were cases where I got rather unexpected results. That said, reading through the code, it should be set on superblocks which have `SB_KERNMOUNT` set, which are created from the `kern_mount()` function (rather than by userspace).

I'll see if I can go through and find the particular vmcore that led me to add this, and try to understand the context better. As it is, I've treated it a bit like magic which is less than ideal :sweat_smile:

brenns10:
> What do you think?

That's even better! I am not 100% attached to the specific heuristic either. It's really just a way to get _any_ path for a dentry that goes all the way down to the root filesystem. I'll update this, probably tomorrow though. Thanks!

drgn/helpers/linux/fs.py
brenns10: Ok, so I decided to go the exploratory route rather because reading the code was confusing me a bit. I made a bit more progress this way. I wrote a drgn script to find any super block with a `MNT_INTERNAL` and then printed all the mounts for that super.

It seems like a lot of kernel filesystems (e.g. selinuxfs, mqueue, tracefs, debugfs) have their first mount on the list set `MNT_INTERNAL`, and these mount points appear to be on their own root. I guess the reason is because these filesystems would exist even if userspace never mounted them, so the kernel makes its own "mount point" first in an internal root filesystem?

And I think this justifies skipping over these mounts in general. Since they're internal to the kernel, they're not seen by userspace, and so they would be a bit confusing to see when debugging. And because they are added to the list first, they will always be the first element of the list for these in-kernel filesystems.

A practical example of an incorrect value avoided by this check might be `/sys/kernel/tracing/trace_pipe`. Without the check, the function would return `b'/trace_pipe'` rather than the more useful `b'/sys/kernel/tracing/trace_pipe'` (or wherever tracefs happens to be mounted).

The flip side is that, if you managed to get your hands on a dentry which _only_ has an internal mount, then this function currently raises an error. So arguably the first internal mount should be saved, and returned as a fallback for the case where no mount is found at all.

https://gist.github.com/brenns10/1d6262133b64615c349722e143dfb3f3

PR #432, opened 2024-09-21
Add name to drgn.Thread

ryantimwilson:
This PR addresses https://github.com/osandov/drgn/issues/143 by adding `name` to drgn's Thread abstraction.

This behaves differently for 3 use cases:
1. Linux kernel - Use task_struct.comm 
2. Live process - Read /proc/<tid>/comm
3. Core dump - Use fname but only valid for main thread. Returns None for other threads

Notes:
1. We do not cache thread names for Linux kernel and live process since these can change during the lifetime of the process. We do cache for core dump since this will not change
2. As a result of not caching, drgn_thread_name will malloc a new string and expect Python to free the memory when the Python object goes out of scope.
3. We assume all thread names are 16 characters including null character. Given this has been a kernel setting for awhile, I don't imagine it changing.

### DIFF
--- a/_drgn.pyi
+++ b/_drgn.pyi
@@ -1110,6 +1110,19 @@ class Thread:
 
     tid: Final[int]
     """Thread ID (as defined by :manpage:`gettid(2)`)."""
+    name: Optional[str]
+    """
+    Thread name, or ``None`` if unknown.
+
+    See `PR_SET_NAME
+    <https://man7.org/linux/man-pages/man2/PR_SET_NAME.2const.html>`_ and
+    `/proc/pid/comm
+    <https://man7.org/linux/man-pages/man5/proc_pid_comm.5.html>`_.
+
+    .. note::
+        Linux userspace core dumps only save the name of the main thread, so
+        :attr:`name` will be ``None`` for other threads.
+    """
     object: Final[Object]
     """
     If the program is the Linux kernel, the ``struct task_struct *`` object for
--- a/libdrgn/drgn.h
+++ b/libdrgn/drgn.h
@@ -3429,6 +3429,15 @@ struct drgn_error *drgn_thread_object(struct drgn_thread *thread,
 struct drgn_error *drgn_thread_stack_trace(struct drgn_thread *thread,
 					   struct drgn_stack_trace **ret);
 
+/**
+ * Get name for the thread represented by @p thread.
+ *
+ * @param[out] ret Returned thread name, or @c NULL if not found. On success, it
+ * should be freed with free(). On error, it is not modified.
+ * @return @c NULL on success, non-@c NULL on error.
+ */
+struct drgn_error *drgn_thread_name(struct drgn_thread *thread, char **ret);
+
 /** @} */
 
 #endif /* DRGN_H */
--- a/libdrgn/program.c
+++ b/libdrgn/program.c
@@ -868,6 +868,35 @@ static struct drgn_error *get_prpsinfo_pid(struct drgn_program *prog,
 	return NULL;
 }
 
+static struct drgn_error *get_prpsinfo_fname(struct drgn_program *prog,
+					   const char *data, size_t size,
+					   const char **ret)
+{
+	bool is_64_bit;
+	struct drgn_error *err = drgn_program_is_64_bit(prog, &is_64_bit);
+	if (err)
+		return err;
+	size_t offset = is_64_bit ? 40 : 28;
+	// pr_fname is defined as 16 byte buffer in elf_prpsinfo
+	// https://github.com/torvalds/linux/blob/075dbe9f6e3c21596c5245826a4ee1f1c1676eb8/include/linux/elfcore.h#L73
+#define PR_FNAME_LEN 16
+	if (size < offset + PR_FNAME_LEN) {
+		return drgn_error_create(DRGN_ERROR_OTHER,
+					 "NT_PRPSINFO is truncated");
+	}
+	// No need to make a copy: the data returned by elf_getdata_rawchunk()
+	// is valid for the lifetime of the Elf handle, and prog->core is valid for
+	// the lifetime of prog.
+	const char *tmp = data + offset;
+	size_t len = strnlen(tmp, PR_FNAME_LEN);
+	if (len == PR_FNAME_LEN)
+#undef PR_FNAME_LEN
+		return drgn_error_create(DRGN_ERROR_OTHER,
+					 "pr_fname is not null terminated");
+	*ret = tmp;
+	return NULL;
+}
+
 struct drgn_error *drgn_thread_dup_internal(const struct drgn_thread *thread,
 					    struct drgn_thread *ret)
 {
@@ -949,6 +978,7 @@ drgn_program_cache_core_dump_notes(struct drgn_program *prog)
 	uint32_t first_prstatus_tid;
 	bool found_prpsinfo = false;
 	uint32_t prpsinfo_pid;
+	const char *prpsinfo_fname = NULL;
 
 	if (prog->core_dump_notes_cached)
 		return NULL;
@@ -1011,6 +1041,12 @@ drgn_program_cache_core_dump_notes(struct drgn_program *prog)
 						       &prpsinfo_pid);
 				if (err)
 					goto err;
+				err = get_prpsinfo_fname(prog,
+						       (char *)data->d_buf + desc_offset,
+						       nhdr.n_descsz,
+						       &prpsinfo_fname);
+				if (err)
+					goto err;
 				found_prpsinfo = true;
 			} else if (nhdr.n_type == NT_PRSTATUS) {
 				uint32_t tid;
@@ -1042,6 +1078,7 @@ out:
 						       &prpsinfo_pid);
 			/* If the PID isn't found, then this is NULL. */
 			prog->main_thread = it.entry;
+			prog->core_dump_fname_cached = prpsinfo_fname;
 		}
 		if (found_prstatus) {
 			/*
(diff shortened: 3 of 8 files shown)


libdrgn/program.c
```diff
@@ -868,6 +870,28 @@ static struct drgn_error *get_prpsinfo_pid(struct drgn_program *prog,
...
+		return err;
+
+	size_t offset = is_64_bit ? 40 : 28;
+	char pr_fname[TASK_COMM_LEN];
+	if (size < offset + sizeof(pr_fname)) {
+		return drgn_error_create(DRGN_ERROR_OTHER,
+					 "NT_PRPSINFO is truncated");
+	}
+	memcpy(pr_fname, data + offset, sizeof(pr_fname));
+	*ret = strndup(pr_fname, sizeof(pr_fname));
```
osandov, line 889: We actually don't need to make a copy here: the data returned by `elf_getdata_rawchunk()` is valid for the lifetime of the `Elf` handle, and `prog->core` is valid for the lifetime of `prog`. So all we need to do here is check that the size is sufficient and that the name is null-terminated, then we can just use the pointer directly.

Btw, let's hard-code 16 for that instead of using `TASK_COMM_LEN` to indicate that the length is defined by the note format. The kernel source code does the same: https://github.com/torvalds/linux/blob/master/include/linux/elfcore.h#L68-L73

libdrgn/program.c
```diff
@@ -1489,6 +1522,86 @@ drgn_thread_object(struct drgn_thread *thread, const struct drgn_object **ret)
...
+drgn_thread_name_linux_kernel(struct drgn_thread *thread, char **ret)
+{
+	struct drgn_error *err;
+	DRGN_OBJECT(comm, drgn_object_program(&thread->object));
+	err = drgn_object_member_dereference(&comm, &thread->object, "comm");
+	if (!err)
+		err = drgn_object_read_c_string(&comm, ret);
+	if (err)
+		return err;
+	return NULL;
```
osandov, line 1535: Trivial nitpick:
```suggestion
	return err;
```

libdrgn/program.c
```diff
@@ -1489,6 +1522,86 @@ drgn_thread_object(struct drgn_thread *thread, const struct drgn_object **ret)
...
+	char buf[TASK_COMM_LEN];
+	size_t bytes_read = 0;
+	while (bytes_read < sizeof(buf)) {
+		size_t bytes_read_iter = fread(buf + bytes_read, 1, sizeof(buf) - bytes_read, file);
+		if (bytes_read_iter == 0) {
+			break;
+		}
+		bytes_read += bytes_read_iter;
+	}
+	fclose(file);
```
osandov, line 1561: `FILE` buffering does win us anything here, and it has more confusing error handling than using a bare fd. We also already have a retry loop helper for fds. How about something like this:
```suggestion
	_cleanup_close_ int fd = open(path, O_RDONLY);
	if (fd < 0)
		return drgn_error_create_os("open", errno, path);

	char buf[TASK_COMM_LEN];
	ssize_t bytes_read = read_all(fd, buf, sizeof(buf));
	if (bytes_read < 0)
		return drgn_error_create_os("read", errno, path);
```

libdrgn/program.c
```diff
@@ -1489,6 +1522,86 @@ drgn_thread_object(struct drgn_thread *thread, const struct drgn_object **ret)
...
+
+	if (bytes_read > 0) {
+		// Remove trailing new line
+		if (buf[bytes_read - 1] == '\n') {
+			buf[bytes_read - 1] = '\0';
+		}
+		*ret = strndup(buf, sizeof(buf));
+		if (!*ret)
+			return &drgn_enomem;
+	}
```
osandov, line 1571: We still need to set `*ret` to something if `bytes_read == 0`. Either `NULL` or an empty string would be fine. And, if we're being defensive against getting garbage without a newline in it, we probably want to limit based on `bytes_read` rather than the size of the buffer. I _think_ this covers both cases, but please double-check my logic.
```suggestion
	if (bytes_read > 0 && buf[bytes_read - 1] == '\n')
		bytes_read--;
	*ret = strndup(buf, bytes_read);
	if (!*ret)
		return &drgn_enomem;
```

libdrgn/program.c
```diff
@@ -1489,6 +1522,86 @@ drgn_thread_object(struct drgn_thread *thread, const struct drgn_object **ret)
...
+drgn_thread_name_userspace_core(struct drgn_thread *thread, char **ret)
+{
+	struct drgn_error *err = drgn_program_cache_core_dump_notes(thread->prog);
+	if (err)
+		return err;
+	// Core dumps only contain the main thread name so check if this is the main thread.
+	// Otherwise, set ret to NULL which will return None in Python.
+	bool is_main_thread = thread->prog->main_thread && thread->prog->main_thread->tid == thread->tid;
+	if (is_main_thread && thread->prog->core_dump_fname_cached) {
+		*ret = strndup(thread->prog->core_dump_fname_cached, TASK_COMM_LEN);
```
osandov, line 1585: If we check that the string is null-terminated when we cache it like I suggested, then this can be `strdup()` instead of `strndup()`.

libdrgn/program.h
```diff
@@ -140,6 +140,7 @@ struct drgn_program {
 	uint64_t aarch64_insn_pac_mask;
 	bool core_dump_notes_cached;
 	bool prefer_orc_unwinder;
+	char* core_dump_fname_cached;
```
osandov, line 143: C style nitpick:
```suggestion
	char *core_dump_fname_cached;
```

libdrgn/python/thread.c
```diff
@@ -56,6 +56,21 @@ static DrgnObject *Thread_get_object(Thread *self)
 	return_ptr(ret);
 }
 
+static PyObject *Thread_get_name(Thread *self)
+{
+	_cleanup_free_ char *ret = NULL;
```
osandov, line 61: If we want to use `_cleanup_free_` for this (which I think we do), then we need `drgn_thread_name()` to guarantee that it only modifies `*ret` on success instead of saying that its contents are undefined on error. For context, before drgn made use of `_cleanup_` attributes, I usually made no such guarantees, but now I've tended to provide that guarantee (e.g., https://github.com/osandov/drgn/commit/e91fcbe22a2c5fcaf8b8765323023919499deb92).

That will just mean using a temporary in a few places instead of assigning directly to `*ret`. E.g., instead of

```c
*ret = strdup(...);
if (!*ret)
	return &drgn_enomem;
```

do this:

```c
char *tmp = strdup(...);
if (!tmp)
	return &drgn_enomem;
*ret = tmp;
```

libdrgn/python/thread.c
```diff
@@ -56,6 +56,21 @@ static DrgnObject *Thread_get_object(Thread *self)
...
+	struct drgn_error *err = drgn_thread_name(&self->thread, &ret);
+	if (err)
+		return set_drgn_error(err);
+	if (!ret) {
+		return Py_None;
+	}
```
osandov, line 67: This needs to be `Py_RETURN_NONE` so that `None` is properly reference-counted on older Python versions where it is not immortal:
```suggestion
	if (!ret)
		Py_RETURN_NONE;
```

libdrgn/python/thread.c
```diff
@@ -56,6 +56,21 @@ static DrgnObject *Thread_get_object(Thread *self)
...
+	_cleanup_pydecref_ PyObject *buf = PyBytes_FromString(ret);
+	if (!buf)
+		return NULL;
+	return_ptr(buf);
```
osandov, line 71: I'm on the fence about this, so let me know if you disagree, but I think making this a `str` instead of `bytes` is a nicer interface. We just need to be carefuly about decoding errors. The filesystem decoder should be good enough. Also, no need for the intermediate variable here.
```suggestion
	return PyUnicode_DecodeFSDefault(name);
```

tests/linux_kernel/test_threads.py
```diff
@@ -50,3 +50,8 @@ def test_crashed_thread(self):
             "crashed thread is only defined for core dumps",
             self.prog.crashed_thread,
         )
+
+    def test_thread_name(self):
+        pid = os.getpid()
+        thread = self.prog.thread(pid)
+        self.assertEqual(thread.object.comm.string_(), thread.name)
```
osandov, line 57: Testing against `thread.object.comm.string_()` is almost a tautology since that's how this works under the hood. Let's test against reading `/proc/{os.getpid()}/comm` instead.

osandov (changes requested):
This looks great overall, thanks! I have several mostly trivial comments.

libdrgn/python/thread.c
ryantimwilson: Agree, I kinda like str better too, I just wanted to avoid encoding issues. But if there's a convenient function for decoding based on default encoder, lets use it!

ryantimwilson:
@osandov thanks for the comments, learned a lot! I've addressed them, let me know if it looks okay

libdrgn/program.c
```diff
@@ -868,6 +868,35 @@ static struct drgn_error *get_prpsinfo_pid(struct drgn_program *prog,
...
+					   const char *data, size_t size,
+					   const char **ret)
+{
+	bool is_64_bit;
+	struct drgn_error *err = drgn_program_is_64_bit(prog, &is_64_bit);
+	if (err)
+		return err;
+	size_t offset = is_64_bit ? 40 : 28;
+	// pr_fname is defined as 16 byte buffer in elf_prpsinfo
+	// https://github.com/torvalds/linux/blob/master/include/linux/elfcore.h#L73
```
osandov, line 881: Let's use a permalink for this.

```suggestion
	// https://github.com/torvalds/linux/blob/075dbe9f6e3c21596c5245826a4ee1f1c1676eb8/include/linux/elfcore.h#L73
```

libdrgn/program.c
```diff
@@ -1489,6 +1526,77 @@ drgn_thread_object(struct drgn_thread *thread, const struct drgn_object **ret)
...
+		+ max_decimal_length(uint32_t)
+		+ 1];
+	snprintf(path, sizeof(path), FORMAT, thread->tid);
+#undef FORMAT
+	_cleanup_close_ int fd = open(path, O_RDONLY);
+	if (fd < 0)
+		return drgn_error_create_os("open", errno, path);
+	// Defined as 16 byte buffer in sched.h
+	// https://github.com/torvalds/linux/blob/master/include/linux/sched.h#L306
+	char buf[16];
```
osandov, line 1555: I missed this last time, but `/proc/$pid/comm` can actually be larger than 16 bytes (currently only for kernel threads, although there were discussions about allowing it more generally IIRC): https://github.com/torvalds/linux/blob/075dbe9f6e3c21596c5245826a4ee1f1c1676eb8/fs/proc/array.c#L101. Since this isn't tied to a binary file format, we might as well future-proof this and use at least 64.

libdrgn/drgn.h
```diff
@@ -3429,6 +3429,15 @@ struct drgn_error *drgn_thread_object(struct drgn_thread *thread,
 struct drgn_error *drgn_thread_stack_trace(struct drgn_thread *thread,
 					   struct drgn_stack_trace **ret);
 
+/**
+ * Get name for the thread represented by @p thread.
+ *
+ * @param[out] ret Returned thread name. On success, it should be freed with
+ * free(). On error, it is NULL.
```
osandov, line 3436: Slight clarification:
```suggestion
 * @param[out] ret Returned thread name, or @c NULL if not found. On success, it
 * should be freed with free(). On error, it is not modified.
```

_drgn.pyi
```diff
@@ -1110,6 +1110,17 @@ class Thread:
...
     tid: Final[int]
     """Thread ID (as defined by :manpage:`gettid(2)`)."""
+    name: Optional[str]
+    """
+    Thread name, or ``None`` if unknown.
+    See `PR_SET_NAME
+    <https://man7.org/linux/man-pages/man2/prctl.2.html#:~:text=PR_SET_NAME>`_
+    and `/proc/[pid]/comm
+    <https://man7.org/linux/man-pages/man5/proc.5.html#:~:text=%2Fproc%2F%5Bpid%5D%2Fcomm>`_.
+    .. note::
```
osandov, line 1120: These links have moved since the original PR. Also, Sphinx needs some blank lines or it gets mad.

```suggestion
    Thread name, or ``None`` if unknown.

    See `PR_SET_NAME
    <https://man7.org/linux/man-pages/man2/PR_SET_NAME.2const.html>`_ and
    `/proc/pid/comm
    <https://man7.org/linux/man-pages/man5/proc_pid_comm.5.html>`_.

    .. note::

```

osandov (commented):
This is pretty much ready to go after some easy fixes, thanks!

libdrgn/program.c
ryantimwilson: Oh wow, TIL!

osandov (approved):
LGTM, thanks!

PR #438, opened 2024-10-10
README.rst: the Arch Package was moved to [extra]

christian-heusel:
This now simplifies the needed setup and therefore the documentation
here aswell.

Signed-off-by: Christian Heusel <christian@heusel.eu>


osandov:
Thank you for officially packaging drgn!

PR #441, opened 2024-10-10
cli: catch FaultError when displaying objects

### LINKED ISSUE #440: Confusing FaultErrors in CLI
The example in #439 of objects with bad addresses, that cause a `FaultError` when displayed, is interesting.

Right now, if you print such an object (like the percpu variable `cpuhp_state`), you get the following error, which admittedly is small and does not crash the REPL like in #439.

```
>>> prog["cpuhp_state"]
  File "/home/stepbren/repos/drgn/drgn/cli.py", line 141, in _displayhook
    text = value.format_(columns=shutil.get_terminal_size((0, 0)).columns)
_drgn.FaultError: could not read memory from kdump: Cannot get page I/O address: PDPT table not present: p4d[0] = 0x0: 0x202a0
>>>
```

It's not too confusing in this case, because there's no real code being executed _except_ for the formatting of the variable. So it doesn't take too long to realize that this must be because it's a percpu variable, and drgn encountered the FaultError while it was trying to format it.

But I've had situations where I wrote some more complex logic, or called a function, which returned an object that can't be formatted. In those cases, I initially assume that the FaultError is in _my_ code and frequently start debugging, until I realize that the error was just with the formatting: the logic was correct, but the object returned couldn't be formatted.

So I wonder if it wouldn't make sense to change the CLI behavior so that it catches these FaultErrors while formatting, and instead just prints the repr()?

osandov:
Modify `_displayhook` to catch `FaultError`, log a warning, and print the `repr()` instead. Fixes #440.

### DIFF
--- a/drgn/cli.py
+++ b/drgn/cli.py
@@ -138,7 +138,11 @@ def _displayhook(value: Any) -> None:
         return
     setattr(builtins, "_", None)
     if isinstance(value, drgn.Object):
-        text = value.format_(columns=shutil.get_terminal_size((0, 0)).columns)
+        try:
+            text = value.format_(columns=shutil.get_terminal_size((0, 0)).columns)
+        except drgn.FaultError as e:
+            logger.warning("can't print value: %s", e)
+            text = repr(value)
     elif isinstance(value, (drgn.StackFrame, drgn.StackTrace, drgn.Type)):
         text = str(value)
     else:


osandov:
@brenns10 let me know what you think of this fix.

brenns10 (approved):
That's great! It's also nice that the message is even suppressed by `drgn -q`. Thanks for the quick fix.

Issue #443, opened 2024-10-11
Add sysfs helpers


osandov:
## Background

When debugging anything related to a device driver, the first step is often finding the `struct device` that represents the device in the kernel. One way to find a device is by its sysfs path. However, drgn doesn't currently have an easy way to do that. You can _kind of_ do it with something like:

```
>>> path = path_lookup("/sys/devices/pci0000:00/0000:00:06.0/0000:04:00.0/nvme/nvme0/nvme0n1")
>>> kn = cast("struct kernfs_node *", path.dentry.d_inode.i_private)
>>> kobject = cast("struct kobject *", kn.priv)
>>> device = container_of(kobject, "struct device", "kobj")
```

But this has a few issues:

* Clearly it's not convenient at all.
* `path_lookup()` walks the dentry cache, which may not be populated for a particular path depending on memory pressure.
* `path_lookup()` doesn't resolve symlinks (yet, see #216), so you have to use the canonical path instead of a more convenient symlink like `/sys/block/nvme0n1`.

Instead, we should improve our kernfs helpers and provide sysfs-specific helpers on top of those. This can be broken down into multiple smaller features.

## kernfs

sysfs is actually an instance of [kernfs](https://en.wikipedia.org/wiki/Kernfs_(Linux)), which is generic infrastructure for pseudo file systems. A kernfs filesystem comprises a tree of [`struct kernfs_node`](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/include/linux/kernfs.h?id=1d227fcc72223cbdd34d0ce13541cbaab5e0d72f#n181), each of which represents a file or directory.

We already have a few basic helpers for inspecting kernfs: https://drgn.readthedocs.io/en/latest/helpers.html#kernfs. I would like two extensions to those:

- [ ] Add a `kernfs_children()` helper that returns an iterator over the children of a directory in kernfs. Children are represented by [`struct kernfs_node::dir::children`](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/include/linux/kernfs.h?id=1d227fcc72223cbdd34d0ce13541cbaab5e0d72f#n156).
[...]

PR #449, opened 2024-10-29
helpers.linux.kernfs: add kernfs_children() and follow_symlinks support

kylee0215:
Add kernfs_children() to iterate over the children of a directory in kernfs.
Add support of follow_symlinks.


drgn/helpers/linux/kernfs.py
```diff
@@ -61,28 +63,44 @@ def kernfs_path(kn: Object) -> bytes:
...
-def kernfs_walk(parent: Object, path: Path) -> Object:
+def kernfs_walk(parent: Object, path: Path, follow_symlinks: bool = False) -> Object:
     """
     Find the kernfs node with the given path from the given parent kernfs node.
 
     :param parent: ``struct kernfs_node *``
     :param path: Path name.
+    :param follow_symlinks: If follow_symlinks is ``False``, and the
+        last component of a path is a symlink, the function will
+        return ``struct kernfs_node *`` of the symbolic link.
```
brenns10, line 74: Why only the last component?

I guess I don't know well enough whether there are cases where symlinks would be in the middle of a real kernfs path. So maybe it won't make a difference. It just seems reasonable to support it within the path too.

brenns10 (commented):
I think this all looks good, just one question regarding why to only follow symlinks at the end of the path.

drgn/helpers/linux/kernfs.py
osandov: That probably came from my description in #443. My intention was that internal symlinks are always followed, and the last one is controlled by the parameter, just like `O_NOFOLLOW` and the `os` module. I haven't read the code to see if there was a misunderstanding.

osandov: Ok, back from vacation and read the code now. Yes, this needs to be changed to always follow symlinks that are not the last component.

osandov (changes requested):
Thanks for working on this! A couple of comments.

osandov:
Superseded by #512 and #549.

PR #450, opened 2024-11-05
contrib/bpf_inspect.py: fix BpfProg repr

zegao96:
When run drgn on kernel v5.4, there is no `tail_call_reachable`
in `prog-> aux`. Check before retrieving the struct in order to
avoid panic


osandov:
Thanks for the fix!

PR #451, opened 2024-12-02
libdrgn: kdump: simplify getting the PRSTATUS attributes

ptesarik:
Since the attribute hierarchy cannot change while iterating over the PRSTATUS attributes, it is not necessary to take an attribute reference.

Although the attribute blob itself should not change either, it is a good idea to keep its data pinned, because a raw pointer to it is stored in the drgn_thread_set hash table. If some code tries to modify the PRSTATUS attribute data, the attempt will fail with KDUMP_ERR_BUSY rather than leave a dangling pointer in the hash table and possibly cause a UAF bug later.

### DIFF
--- a/libdrgn/kdump.c
+++ b/libdrgn/kdump.c
@@ -299,39 +299,31 @@ struct drgn_error *drgn_program_cache_kdump_threads(struct drgn_program *prog)
 	}
 
 	/*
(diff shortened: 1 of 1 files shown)


ptesarik:
Argh, scratch it. I thought I knew what I wanted with the reference counts, but it turns out I don't and I would cause regressions.

PR #452, opened 2024-12-03
libdrgn: prepare for libkdumpfile-0.5.5

ptesarik:
Adapt to libkdumpfile attribute lifetime fixes.


osandov:
Thanks! I was only able to test 0.5.4, but I'm taking your word for it on the upcoming 0.5.5.

PR #455, opened 2024-12-16
drgn.internal.repl: also catch AttributeError in enhanced REPL attempt

osandov:
I didn't see this in practice, it's mostly out of paranoia. In particular, if `_pyrepl.readline._setup()` goes away, this will catch it.

@brenns10 any objections?


brenns10 (approved):
Good catch! I used to write these with just `except ImportError` and I learned my lesson regarding `ModuleNotFoundError`, but `AttributeError` is another good one to add in this situation.

PR #456, opened 2024-12-17
Add drgn_error_catch() - take 2

brenns10:
I had opened a PR a while back for a new static inline, `drgn_error_catch()`, that makes it easier to catch all errors of a specific code, and destroy them. I tried to convert a bunch of users in that PR which made it go stale super quick. I don't really think that it's necessary to convert a bunch all in one go. So this PR just adds the helper, and does an example conversion in `libdrgn/linux_kernel_helpers.c`. That ensures that we can see how it's used and catch any errors, but it also won't conflict with the module API branch (I double-checked), and shouldn't go stale so quickly.

I think this would be a nice thing to have in general. It's also related to some of the preparatory commits I have for the CTF branch, and I figured it would be nice to keep closer to upstream by adding this now.

### DIFF
--- a/libdrgn/error.h
+++ b/libdrgn/error.h
@@ -155,6 +155,24 @@ static inline void drgn_recursion_guard_cleanup(int **guard)
 	__attribute__((__cleanup__(drgn_recursion_guard_cleanup), __unused__))	\
 	int *PP_UNIQUE(recursion_count_ptr) = &unique_recursion_count
 
+/**
+ * Catch a certain kind of @ref drgn_error and free it
+ *
+ * If @a errp points to a non-@c NULL error whose code matches @a code, then the
+ * free the error (if necessary), replace the pointer value with @c NULL, and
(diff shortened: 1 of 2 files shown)


osandov:
Looks good to me conceptually, although there's a build error (which apparently makes the CI action get stuck. Oops.)

brenns10:
How embarrassing, I swear I built and ran this, but I must not have had the `linux_kernel_helpers.c` changes applied yet. I've fixed the build error and now actually tested loading a vmcore and /proc/kcore with this, as well as running vmtests against 6.12.

osandov (commented):
Oh, I just noticed that the new function is in `drgn.h`. Do you need it to be public? If not, let's move it to be internal in `error.h`. Other than that, LGTM.

brenns10:
Nope, I don't need a public API (we don't currently even release `drgn-devel`). I'll move it!

PR #458, opened 2024-12-19
Update Ubuntu instructions

michel-slm:
Mention that `drgn` is shipped with Ubuntu except jammy, but it's an older version


osandov:
Thanks!

PR #459, opened 2024-12-19
Make StackFrame.name fall back to symbol/PC and add StackFrame.function_name

osandov:
Details are in the commit message, @brenns10 does this suit your needs?


brenns10 (approved):
This looks great! Thanks for doing this change :D

PR #460, opened 2024-12-20
Support for built-in ORC information

brenns10:
Thanks to your work today with 3ce0feea ("tests: don't clobber file in use by libelf") and a1869f95 ("Make StackFrame.name fall back to symbol/PC and add StackFrame.function_name"), this branch is fully unblocked!

Now that the module API has landed, we can usually rely on having a `struct drgn_module` available for a kernel module, even if the module debuginfo is not loaded (either because you have only loaded debuginfo for the kernel, or because you are using CTF). My previous ORC support required essentially re-implementing a small portion of the module API, specifically a tree that mapped address ranges to ORC data. Now, I can drop all that complexity and share this branch which I think is ready for consideration.

This branch's main goal is to enable using built-in ORC information for stack unwinding, for the cases where a module does not have debuginfo. I think the commit messages here describe everything that's important. I think the testing is okay -- I wish I could test the vmlinux ORC, but I think I would need a type finder to allow the module API to even get initialized.

### DIFF
--- a/_drgn.pyi
+++ b/_drgn.pyi
@@ -838,12 +838,11 @@ class Program:
         self, module_obj: Object, *, create: Literal[False] = False
     ) -> RelocatableModule:
         """
-        Find a Linux kernel loadable module from a ``struct module`` object.
+        Find a Linux kernel loadable module from a ``struct module *`` object.
 
         Note that kernel modules are represented as relocatable modules.
 
-        :param module_obj: ``struct module`` or ``struct module *`` object for
-            the kernel module.
+        :param module_obj: ``struct module *`` object for the kernel module.
         :return: Relocatable module with a name and address matching
             *module_obj*.
         :raises LookupError: if no matching module has been created
@@ -855,14 +854,13 @@ class Program:
         self, module_obj: Object, *, create: Literal[True]
     ) -> Tuple[RelocatableModule, bool]:
         """
-        Find or create a Linux kernel loadable module from a ``struct module``
+        Find or create a Linux kernel loadable module from a ``struct module *``
         object.
 
         If a new module is created, its :attr:`~Module.address_range` and
         :attr:`~RelocatableModule.section_addresses` are set from *module_obj*.
 
-        :param module_obj: ``struct module`` or ``struct module *`` object for
-            the kernel module.
+        :param module_obj: `struct module *`` object for the kernel module.
         :return: Module and ``True`` if it was newly created or ``False`` if it
             was found.
         """
@@ -1665,6 +1663,14 @@ class Module:
     state/core dump when possible. Otherwise, when a file is assigned to the
     module, it is set to the file's build ID if it is not already set. It can
     also be set manually.
+    """
+    object: Object
+    """
+    The object associated with this module.
+
+    For Linux kernel loadable modules, this is the ``struct module *``
+    associated with the kernel module. For other kinds, this is currently an
+    absent object. The object may be set manually.
     """
     loaded_file_status: ModuleFileStatus
     """Status of the module's :ref:`loaded file <module-loaded-file>`."""
--- a/libdrgn/drgn.h
+++ b/libdrgn/drgn.h
@@ -1314,7 +1314,7 @@ drgn_module_find_or_create_relocatable(struct drgn_program *prog,
 				       struct drgn_module **ret, bool *new_ret);
 
 /**
- * Find a created Linux kernel loadable module from a ``struct module`` object.
+ * Find a created Linux kernel loadable module from a ``struct module *`` object.
  *
  * @param[out] new_ret @c true if the module was newly created, @c false if it
  * was found.
@@ -1324,7 +1324,7 @@ drgn_module_find_linux_kernel_loadable(const struct drgn_object *module_obj,
 				       struct drgn_module **ret);
 
 /**
- * Find a Linux kernel loadable module from a ``struct module`` object, creating
+ * Find a Linux kernel loadable module from a ``struct module *`` object, creating
  * it if it doesn't already exist.
  *
  * @param[out] new_ret @c true if the module was newly created, @c false if it
@@ -1569,6 +1569,26 @@ drgn_module_wanted_supplementary_debug_file(struct drgn_module *module,
 					    const char **supplementary_path_ret,
 					    const void **checksum_ret,
 					    size_t *checksum_len_ret);
+
+/**
+ * Return the object associated with this module.
+ *
+ * For some modules, there may be an object related to it. For example, drgn
+ * automatically identifies the Linux kernel `struct module *` associated with
+ * loadable modules, and associates it with them. Users may set or replace an
+ * associated object with @ref drgn_set_module_object().
+ *
+ * @param[out] ret Initialized object where the module object is placed
+ */
+struct drgn_error *
+drgn_module_object(const struct drgn_module *module, struct drgn_object *ret);
+
+/**
+ * Set the object associated with this module.
+ * @param[in] obj A new (or replacement) object for the module
+ */
+struct drgn_error *
+drgn_module_set_object(struct drgn_module *module, const struct drgn_object *obj);
 
 /** Debugging information finder callback table. */
 struct drgn_debug_info_finder_ops {
--- a/libdrgn/orc_info.c
+++ b/libdrgn/orc_info.c
@@ -13,6 +13,7 @@
 #include "debug_info.h" // IWYU pragma: associated
 #include "elf_file.h"
 #include "error.h"
+#include "log.h"
 #include "orc.h"
 #include "platform.h"
 #include "program.h"
@@ -34,7 +35,7 @@ static inline uint64_t drgn_raw_orc_pc(struct drgn_module *module,
 {
 	int32_t offset;
 	memcpy(&offset, &module->orc.pc_offsets[i], sizeof(offset));
-	if (drgn_elf_file_bswap(module->debug_file))
+	if (module->orc.bswap)
 		offset = bswap_32(offset);
 	return module->orc.pc_base + UINT64_C(4) * i + offset;
 }
@@ -44,7 +45,7 @@ drgn_raw_orc_entry_is_terminator(struct drgn_module *module, unsigned int i)
 {
 	uint16_t flags;
 	memcpy(&flags, &module->orc.entries[i].flags, sizeof(flags));
-	if (drgn_elf_file_bswap(module->debug_file))
+	if (module->orc.bswap)
 		flags = bswap_16(flags);
 	if (module->orc.version >= 3) {
 		// orc->type == ORC_TYPE_UNDEFINED
@@ -63,7 +64,7 @@ drgn_raw_orc_entry_is_preferred(struct drgn_module *module, unsigned int i)
 {
 	uint16_t flags;
 	memcpy(&flags, &module->orc.entries[i].flags, sizeof(flags));
-	if (drgn_elf_file_bswap(module->debug_file))
+	if (module->orc.bswap)
 		flags = bswap_16(flags);
 	// ORC_REG_SP_INDIRECT is used for the stack switching pattern used in
 	// the Linux kernel's call_on_stack()/call_on_irqstack() macros. See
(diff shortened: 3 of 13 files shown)


brenns10:
So, the error here was that the 4.9 kernel does not support ORC, and I was passing through the lookup error for "num_orcs" in `struct mod_arch_specific`, rather than swallowing it (since that shouldn't interrupt a stack trace).

The interesting thing is that after correcting that, the test passes, because on 4.9, unwind can be done via frame pointers rather than ORC. So really, my stack tracing test isn't (necessarily) testing the ability to unwind with ORC, it's testing the ability to unwind with _anything_ other than DWARF. I don't really have a knob to test this in the Python API. But unfortunately, this isn't really something easy to test in a C unit test either.

brenns10:
I made a couple changes:

1. Now, we log whenever we load built-in ORC.
2. The tests now detect the log message to verify that ORC is getting used for unwinding.
3. The tests also skip kernel 4.9 by detecting whether the kernel has the `__start_orc_unwind` symbol.

brenns10:
Interestingly, my test above that tested log messages failed on Python 3.6. I guess there is a difference in how the logging got initialized. I've added a context manager to explicitly set drgn's log level to DEBUG for the duration of the test, which I believe resolves the issue.

Finally, I've gone ahead and added one more test, which actually does test vmlinux ORC. It creates a Program with no debuginfo, and then copies a `struct pt_regs` from the Program which _has_ debuginfo, and asks to unwind that. It's quite funky, but it does work, and I tested it on 4.9 as well as 6.12.

With that, I do feel happy about the testing now!

_drgn.pyi
```diff
@@ -1499,6 +1499,14 @@ class Module:
     module, it is set to the file's build ID if it is not already set. It can
     also be set manually.
     """
+    object: Object
+    """
+    The object this module was created from.
+
+    For Linux kernel loadable modules, this is a reference to the ``struct
+    module`` this was created from. For other kinds, this is currently an absent
```
osandov, line 1507: A pointer (i.e. `struct module *`) feels slightly more natural to me (following the precedent of `drgn.Thread.object`). I'd be fine restricting `drgn.Program.linux_kernel_loadable_module()` to only accept `struct module *` and not `struct module` to make that simpler.

libdrgn/python/module.c
```diff
@@ -285,6 +285,24 @@ static int Module_set_build_id(Module *self, PyObject *value, void *arg)
 	return 0;
 }
 
+static DrgnObject *Module_object(Module *self, void *arg)
+{
+	struct drgn_error *err;
+
+	struct drgn_program *prog = drgn_module_program(self->module);
+	Program *prog_obj = container_of(prog, Program, prog);
```
osandov, line 293: There are enough places that do this that we probably need a `Program *Module_prog(Module *module)` akin to `DrgnObject_prog()` and `DrgnType_prog()`.

libdrgn/drgn.h
```diff
@@ -1570,6 +1570,17 @@ drgn_module_wanted_supplementary_debug_file(struct drgn_module *module,
...
+/**
+ * Return the object associated with this module.
+ *
+ * For Linux kernel loadable modules, this is a reference to the `struct module`
+ * they represent. Other kinds of modules may have other objects.
+ *
+ * @param[out] ret Initialized object where the module object is placed
+ */
+struct drgn_error *
+drgn_module_object(struct drgn_module *module, struct drgn_object *ret);
```
osandov, line 1582: Can you please also add a `drgn_module_set_object()` (and the corresponding Python binding setter)? The module API in general lets the user say it knows better than drgn, so we should have a way to override this (for any kind of module).

osandov (commented):
I perused the ORC changes and they looked sane overall, but I'll have to give them a closer look after the holidays. I looked over the `Module.object` change more closely, so some comments there.

The Python 3.6 logging issues are probably due to our very sketchy integration with Python logging: https://github.com/osandov/drgn/blob/a1869f9549c1af6f5393406895bfc97ef48a5ac5/libdrgn/python/program.c#L144-L155.

brenns10:
Ahh, thank you! I knew about the loglevel monkey-patching, but I hadn't read the full comment, so I missed the bit about syncing the log level when the Program is created. That makes sense, and I can easily fix it.

And yeah, I should have made more clear on this that I wasn't hoping for quick action before the holidays :) All the quick updates have been because I've been nerd-sniped by the tests, which are really fun to make work.

brenns10:
At your leisure: How do you feel about the API choice of `Module.object` having an absent object, instead of `None`, when there is no object present? On the one hand, it seems nice to not have an `Optional[Object]` type annotation. But on the other, part of me feels like it's overloading the concept of an absent object.

osandov:
Stick with the absent object, I prefer that over `None` in this case.

_drgn.pyi
brenns10: Yeah, I was thinking about that as well. I get that value pointers vs reference objects are always confusing to users, so I agree it would probably be better to stick with the convention that's already established.

I'd actually love to try to keep `linux_kernel_loadable_module()` accepting `struct module`. If we can't create a pointer to the object, then we can just leave the `Module.object` field absent as before. If you're ok with that.

The reason being that I really enjoyed the possibility of somebody constructing a `struct module` value Object, and then passing it in. I don't actually know how feasible that would be, but it's kinda fun :)

libdrgn/drgn.h
brenns10: It strikes me that this opens up the possibility for a user to assign a completely irrelevant object to the module, thus making it much more likely for an error to occur when we're using it. I don't have a problem with that, of course: good error handling is important regardless.

But in my experience, sometimes errors that come from too deep within the bowels of drgn are no longer helpful, because they don't come with a stack trace like Python errors. For example here: if a user ran `prog.thread(1).stack_trace()`, and got an error related to the ORC unwinder, it may not be obvious where it came from.

I know we can catch errors that we expect to be likely, and replace them with more informative ones. But I wonder if it wouldn't be worthwhile to add some sort of error-chaining functionality for drgn? That way, we could implement it so that any error returned by `drgn_read_builtin_orc()` might get wrapped with the context "encountered error while loading built-in ORC for module foo", without eliminating the specifics of the error that is wrapped.

That's just an oddball idea which doesn't have to be part of this PR, in any case. And I could make the case against it as well: maybe we should rely on logging to give us the context for the error! Just an idea to ponder over the holidays.

_drgn.pyi
brenns10: I've gone ahead and changed this to a `struct module *`, and for simplicity, I did restrict it so that `linux_kernel_loadable_module()` only accepts a pointer. We can always add back some flexibility if we find it useful.

libdrgn/drgn.h
brenns10: I've gone ahead and made `Module.object` assignable. I do still think error chaining could be useful to make errors inside the internals a bit easier to understand. But that sounds like a separate issue.

brenns10:
There was a strange issue with Github CI (504 error fetching kernels). Assuming they pass this time, this is ready for review again at your convenience. Hope you had a great holiday!

libdrgn/orc_info.c
```diff
@@ -367,18 +526,40 @@ struct drgn_error *drgn_module_parse_orc(struct drgn_module *module)
...
-	// clear them if anything goes wrong.
+	// Depending on the source of pc_offsets and entries, we may or may not
+	// need to free them. We should always clear the pointers out of the
+	// module to avoid an invalid or double free in
+	// drgn_module_orc_info_deinit().
 	_cleanup_(drgn_module_clear_orc) struct drgn_module *clear = module;
+	_cleanup_free_ void *cleanup_pc_offsets = NULL;
+	_cleanup_free_ void *cleanup_entries = NULL;
+
+	if (module->debug_file) {
```
osandov, line 540: If `debug_file` has (for some cursed reason) a different platform than the program, then `can_use_debug_file` in `drgn_module_parse_orc()` will be false, and it will call `drgn_module_parse_orc()` expecting to read the built-in ORC info. But then this will try to parse it from the debug file.

I think it'd be better to pass a `bool use_builtin` or something like that to `drgn_module_parse_orc()` to avoid that confusion.

libdrgn/orc_info.c
```diff
@@ -355,6 +356,164 @@ static struct drgn_error *drgn_read_orc_sections(struct drgn_module *module)
...
+		return err;
+
+	// We'll still use the section addresses for everything else, because
+	// the orc_header is only present there, and it should be a bit faster
+	// to read data which we already parsed, rather than going back to read
+	// it from program memory.
+	uint64_t orc_unwind;
+	uint64_t orc_unwind_ip;
+	uint64_t orc_header = 0;
+	drgn_module_get_section_address(module, ".orc_unwind", &orc_unwind);
```
osandov, line 502: Missing `err =` here and below.

libdrgn/orc_info.c
```diff
@@ -355,6 +356,164 @@ static struct drgn_error *drgn_read_orc_sections(struct drgn_module *module)
...
+#undef get_symbol
+
+	if ((unwind_ip_end - unwind_ip_start) % sizeof(int32_t))
+		return drgn_error_create(DRGN_ERROR_OTHER, "invalid built-in orc_unwind_ip range");
+	uint64_t num_entries = (unwind_ip_end - unwind_ip_start) / sizeof(int32_t);
+	if (num_entries > UINT_MAX)
+		return drgn_error_create(DRGN_ERROR_OTHER,
+					 "built-in orc_unwind_ip range is too large");
+
+	if ((unwind_end - unwind_start) % sizeof(struct drgn_orc_entry)	\
```
osandov, line 444: No need for  this backslash.
```suggestion
	if ((unwind_end - unwind_start) % sizeof(struct drgn_orc_entry)
```

libdrgn/orc_info.c
```diff
@@ -367,18 +526,40 @@ struct drgn_error *drgn_module_parse_orc(struct drgn_module *module)
...
+	if (module->prog->platform.arch->arch != DRGN_ARCH_X86_64)
 		return NULL;
 
-	// pc_offsets and entries point to the Elf_Data buffers until we're
-	// done. We don't want those freed by drgn_module_orc_info_deinit(), so
-	// clear them if anything goes wrong.
+	// Depending on the source of pc_offsets and entries, we may or may not
+	// need to free them. We should always clear the pointers out of the
+	// module to avoid an invalid or double free in
+	// drgn_module_orc_info_deinit().
```
osandov, line 535: Could you please retain a mention of how the buffers temporarily come from `Elf_Data` in the not built-in case?

libdrgn/orc_info.c
```diff
@@ -355,6 +356,164 @@ static struct drgn_error *drgn_read_orc_sections(struct drgn_module *module)
...
+		module->orc.version = orc_version_from_header(header_data);
+		if (module->orc.version < 0)
+			return drgn_error_create(DRGN_ERROR_OTHER,
+							"unrecognized .orc_header");
+	} else {
+		module->orc.version = orc_version_from_osrelease(module->prog);
+	}
+
+	_cleanup_free_ int32_t *pc_offsets = malloc_array(num_entries,
+							  sizeof(pc_offsets[0]));
```
osandov, line 384: Needs to check if the allocation succeeded.

libdrgn/orc_info.c
```diff
@@ -355,6 +356,164 @@ static struct drgn_error *drgn_read_orc_sections(struct drgn_module *module)
...
+	err = drgn_program_read_memory(module->prog, pc_offsets, unwind_ip,
+				       num_entries * sizeof(pc_offsets[0]), false);
+	if (err)
+		return err;
+
+	_cleanup_free_ struct drgn_orc_entry *entries =
+		malloc_array(num_entries, sizeof(entries[0]));
```
osandov, line 391: Likewise.

libdrgn/orc_info.c
```diff
@@ -318,7 +315,9 @@ static struct drgn_error *drgn_read_orc_sections(struct drgn_module *module)
 		err = read_elf_section(orc_header_scn, &orc_header);
 		if (err)
 			return err;
-		module->orc.version = orc_version_from_header(orc_header);
+		module->orc.version = -1;
+		if (orc_header->d_size == 20)
```
osandov, line 320: By the end of this PR, 20 is hard-coded in a bunch of places. Let's define a constant for it.

tests/linux_kernel/test_stack_trace.py
```diff
@@ -59,6 +62,60 @@ def test_by_pid_dwarf(self):
...
+            # Create a program with the core kernel debuginfo loaded,
+            # but without module debuginfo. Load a symbol finder using
+            # kallsyms so that the module's stack traces can still have
+            # usable frame names.
+            prog = Program()
+            prog.set_kernel()
+            try:
+                prog.load_default_debug_info()
+            except MissingDebugInfoError:
+                pass
```
osandov, line 103: Do you need to load the debug info for all of the other modules? If not, how about
```suggestion
            prog.load_debug_info(main=True)
```

osandov (changes requested):
I'm happy with the overall structure, just some minor comments. Thanks!

libdrgn/orc_info.c
brenns10: how embarassing! lol

tests/linux_kernel/test_stack_trace.py
brenns10: Drgn needs to at least _enumerate_ the modules, so that there is a `drgn_module` created. Otherwise, we have nowhere to put the ORC and the test fails. But we don't need to go searching for the debuginfo. I have achieved this in other places by placing a dummy call of `list(prog.loaded_modules())`.

brenns10: I went ahead and used `load_debug_info(main=True)` and added a call to `list(prog.loaded_modules())`. Though I'm not sure that is 100% satisfactory so I'll leave this open for further discussion.

libdrgn/orc_info.c
```diff
@@ -21,6 +21,8 @@
 
 DEFINE_VECTOR(uint64_range_vector, struct uint64_range);
 
+#define ORC_HASH_SIZE 20
```
osandov, line 24: If I may bikeshed, how about `ORC_HEADER_SIZE`? (Since it's an implementation detail that the header is a hash. Clearly I didn't follow that logic in my own naming of `orc_hash_6_4` and co.)

osandov (changes requested):
A few more minor nitpicks, plus one bug in "orc_info: apply debug file bias to pc_base at load" that I missed last time. Thanks!

libdrgn/orc_info.c
```diff
@@ -554,6 +554,7 @@ struct drgn_error *drgn_module_parse_orc(struct drgn_module *module,
 		cleanup_entries = module->orc.entries;
 	} else if (module->debug_file) {
 		err = drgn_read_orc_sections(module);
+		module->orc.pc_base += module->debug_file_bias;
```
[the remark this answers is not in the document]
brenns10: Great catch, thank you. I verified that it did break `remove_fdes_from_orc()` and have updated it so that the `debug_file_bias` is subtracted out within that function. I suppose I could have updated `drgn_orc_raw_pc()` but I don't think that it belongs there.

I could have postponed adjusting `orc.pc_base` until after the call to `remove_fdes_from_orc()`. But then `remove_fdes_from_orc()` would be broken for builtin-ORC. Because, related to our DM conversation, some debug_files may not have ORC, yet they still have `.debug_frame`. In that situation also, the bias needs to be subtracted away from the ORC PC as well.

libdrgn/orc_info.c
```diff
@@ -355,6 +356,168 @@ static struct drgn_error *drgn_read_orc_sections(struct drgn_module *module)
...
+	uint64_t num_entries = (unwind_ip_end - unwind_ip_start) / sizeof(int32_t);
+	if (num_entries > UINT_MAX)
+		return drgn_error_create(DRGN_ERROR_OTHER,
+					 "built-in orc_unwind_ip range is too large");
+
+	if ((unwind_end - unwind_start) % sizeof(struct drgn_orc_entry)
+	    || (unwind_end - unwind_start) / sizeof(struct drgn_orc_entry) != num_entries)
+		return drgn_error_create(DRGN_ERROR_OTHER, "invalid built-in orc_unwind range");
+
+	module->orc.version = -1;
```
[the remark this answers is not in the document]
brenns10: It is from an older version of the implementation and I forgot to remove it. Thanks for the catch.

brenns10:
Ok, I've gone ahead and addressed this batch of feedback. This time it didn't require a rebase to pull in recent changes from main. So it should be easier to see the changes from the previous revision.

Thank you for the catch on `remove_fdes_from_orc()`. In addition to resolving those finds, I did add a case to allow loading built-in ORC when there's a debug_file that does not have the ORC sections. In practice, after the `remove_fdes_from_orc()` step, it seems like most modules end up with just one ORC entry, so I'm not entirely certain whether it would actually be worth the cost.

brenns10:
The failure above was in 4.9, which doesn't have ORC. It was a really good catch as well. I should not return an error from `drgn_read_vmlinux_orc()` if the "non-optional" symbols are not found - this is an expected case where we don't load anything and just return NULL. The failure only started now because with my recent update, drgn checks for built-in ORC after it fails to find it in the debug file. However, the failure would have happened with any non-ORC kernel, if you tried to do a stack trace without having a debug file loaded too.

libdrgn/orc_info.c
```diff
@@ -152,11 +152,15 @@ remove_fdes_from_orc(struct drgn_module *module, unsigned int *indices,
 	unsigned int num_entries = *num_entriesp;
 	unsigned int new_num_entries = 0;
 
-	uint64_t start_pc = drgn_raw_orc_pc(module, 0);
+	// ORC can be built-in or from the debug file. Because of that, we
+	// always store the biased/actual address at orc.pc_base. Since we are
+	// comparing to the unbiased addresses in the debug_frame FDEs, we need
+	// to subtract the bias from the ORC PC.
+	uint64_t start_pc = drgn_raw_orc_pc(module, 0) - module->debug_file_bias;
```
osandov, line 161: It's somewhat confusing that `drgn_module_orc_info::preferred` contains unbiased addresses and `drgn_orc_pc()`/`drgn_raw_orc_pc()` is biased. `drgn_module_should_prefer_orc_cfi()` expects unbiased addresses, so I think it's correct as you've written it, just slightly awkward. Do you think we should change it to all be biased? I'm on the fence.

I'm going to add a test case that I should have added in 98995b617d4479daf4846a18b3ce97cbaa554628...

osandov (commented):
Thanks, this looks good other than one thing I wanted your opinion on.

osandov (approved):
Ok, I'm fine with leaving `preferred` as is. I'm planning on merging this as soon as I land the test case for `preferred` (in my `orc-preferred-test` branch), but I'm chasing down some weird AArch64 failures with that test.

osandov:
The failures I saw were sporadic on the tiny flavor. The test doesn't even try to cross the IRQ boundary on aarch64, it just sometimes fails to unwind out of crash_kexec at all. This was without your changes, so obviously unrelated.

osandov:
I definitely saw it on aarch64:6.10*tiny, among other versions I don't remember off the top of my head. I wouldn't mind another set of eyes, but no pressure!

osandov:
The new test passes on x86-64 with your changes, so I'm not going to block this on an obscure AArch64 issue any longer, I'll go ahead and merge this.

brenns10:
Ah, thanks! I've been playing around with the aarch64 builds but encountered an unrelated issue with QEMU on my desktop, will keep looking once I get that resolved though.

osandov:
Oh, I really should've mentioned that you need
```patch
 vmtest/config.py | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)

diff --git a/vmtest/config.py b/vmtest/config.py
index 8f3e42e3..d77f97a8 100644
--- a/vmtest/config.py
+++ b/vmtest/config.py
@@ -285,7 +285,7 @@ ARCHITECTURES = {
                 """,
             },
             kernel_org_compiler_name="aarch64-linux",
-            qemu_options=("-M", "virt", "-cpu", "cortex-a57"),
+            qemu_options=("-M", "virt", "-cpu", "cortex-a76"),
             qemu_console="ttyAMA0",
         ),
         Architecture(
```
or the tiny flavor hangs on boot. Hopefully that's not what you've been debugging :sweat: 

That requires a fairly recent QEMU. I haven't decided whether to commit that patch or make it conditional on the flavor so that it's still usable with older QEMU for the other flavors.

brenns10:
Oh yep that's the one, thanks! I didn't spend much time on it though so no worries.

Anyway, yeah I can repro the failure now, will take a look.

I have to imagine that there aren't too many users of all the vmtest infra (I only recently started using it) that you'd risk breaking.

PR #476, opened 2025-03-07
Fix exception handling in `for_each_child_dentry`

septatrix:
Accessing `dentry.d_children` fails before the lookup of `d_sib` on `struct dentry`.

```txt
  File "negdentdelete.py", line 119, in main
    for batch in yield_negative_dentries(prog, directory, args.chunk_size):
                 ~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "negdentdelete.py", line 52, in yield_negative_dentries
    for child in for_each_child_dentry(parent):
                 ~~~~~~~~~~~~~~~~~~~~~^^^^^^^^
  File "negdentdelete.py", line 37, in for_each_child_dentry
    dentry.d_children.address_of_(),
    ^^^^^^^^^^^^^^^^^
AttributeError: 'struct dentry' has no member 'd_children'. Did you mean: 'd_child'?
```

### DIFF
--- a/contrib/negdentdelete.py
+++ b/contrib/negdentdelete.py
@@ -37,7 +37,7 @@ def for_each_child_dentry(dentry: Object) -> Iterator[Object]:
             dentry.d_children.address_of_(),
             "d_sib",
         )
-    except LookupError:
+    except AttributeError:
         return list_for_each_entry(
             "struct dentry", dentry.d_subdirs.address_of_(), "d_child"
         )


osandov:
LGTM, you just need to amend your commit to add a Signed-off-by as per https://github.com/osandov/drgn/blob/main/CONTRIBUTING.rst#signing-off and force push.

osandov:
As long as it matches the commit author and email, it should be fine (it looks like your commit author is "Nils K" but the sign-off says "Septatrix").

contrib/negdentdelete.py
```diff
@@ -37,7 +37,7 @@ def for_each_child_dentry(dentry: Object) -> Iterator[Object]:
             dentry.d_children.address_of_(),
             "d_sib",
         )
-    except LookupError:
+    except AttributeError:
```
brenns10, line 40: What's worse is that I've clearly written this helper before and I got it right everywhere else:
https://github.com/oracle-samples/drgn-tools/blob/main/drgn_tools/dentry.py#L42

septatrix:
The mismatch was because I authored it in the web UI and added the sign-off locally. I reset the author so both should now refer to septatrix

brenns10 (approved):
Yeah thanks for fixing this up :)

Maybe `for_each_child_dentry()` should get promoted to a helper in `drgn.helpers.linux.fs` with this fix?

osandov:
Thanks!

PR #478, opened 2025-03-11
Add a "-e" option to execute code provided as a command-line argument

brenns10:
For example:
```
$ python -m drgn -qp0 -e 'print_annotated_memory(prog, id(prog), 256)'
ADDRESS           VALUE
000055cf0b4c4bd0: 0000000000000006
000055cf0b4c4bd8: 00007fa1ae3fbd80 [object symbol: Program_type+0x0]
000055cf0b4c4be0: 000055cf0b4b3780
000055cf0b4c4be8: 0000000000000000
000055cf0b4c4bf0: 000055cf0b4ba560
000055cf0b4c4bf8: 0000000000000000
000055cf0b4c4c00: 0007643700000003
000055cf0b4c4c08: 0000000000000000
000055cf0b4c4c10: 000055cf0b4c4d78
000055cf0b4c4c18: 0000000000000001
000055cf0b4c4c20: 000055cf0b4c4be0
000055cf0b4c4c28: 00007fa1ae3f6400 [object symbol: drgn_language_c+0x0]
000055cf0b4c4c30: 0000000000000000
000055cf0b4c4c38: 0000000000000000
000055cf0b4c4c40: 0000000000000000
000055cf0b4c4c48: 0000000000000001
000055cf0b4c4c50: 000055cf0b4c4be0
000055cf0b4c4c58: 00007fa1ae3f6320 [object symbol: drgn_language_cpp+0x0]
000055cf0b4c4c60: 0000000000000000
000055cf0b4c4c68: 0000000000000000
000055cf0b4c4c70: 0000000000000000
000055cf0b4c4c78: 0000000000000000
000055cf0b4c4c80: 0000000000000000
000055cf0b4c4c88: 0000000000000000
000055cf0b4c4c90: 0000000000000000
000055cf0b4c4c98: 0000000000000000
000055cf0b4c4ca0: 0000000000000000
000055cf0b4c4ca8: 0000000000000000
000055cf0b4c4cb0: 0000000000000000
000055cf0b4c4cb8: 0000000000000000
000055cf0b4c4cc0: 0000000000000000
000055cf0b4c4cc8: 0000000000000000
```

The sys.path has the current working directory added, just like the default CLI. This is unnecessary if you run drgn by `python -m drgn` which I normally do, but for most distro installs, drgn is a script and so the current working directory must be added manually.

### DIFF
--- a/drgn/cli.py
+++ b/drgn/cli.py
@@ -387,6 +387,12 @@ def _main() -> None:
         help="don't print any logs or download progress",
     )
     parser.add_argument(
+        "-e",
+        dest="exec",
+        metavar="CODE",
+        help="an expression or statement to evaluate, instead of running in interactive mode",
+    )
+    parser.add_argument(
         "script",
         metavar="ARG",
         type=str,
@@ -397,7 +403,7 @@ def _main() -> None:
 
     args = parser.parse_args()
 
-    if args.script:
+    if args.script and not args.exec:
         # A common mistake users make is running drgn $core_dump, which tries
         # to run $core_dump as a Python script. Rather than failing later with
         # some inscrutable syntax or encoding error, try to catch this early
@@ -413,7 +419,7 @@ def _main() -> None:
             )
         elif script_type == "elf":
             sys.exit(f"error: {args.script[0]} is a binary, not a drgn script")
-    else:
+    elif not args.exec:
         print(version, file=sys.stderr, flush=True)
 
     if args.log_level == "none":
@@ -496,7 +502,11 @@ def _main() -> None:
             if new:
                 module.try_file(extra_symbol_path)
 
-    if args.script:
+    if args.exec:
+        sys.path.insert(0, "")
+        sys.argv = ["-e"] + args.script
+        exec(args.exec, default_globals(prog))
+    elif args.script:
         sys.argv = args.script
         script = args.script[0]
         if pkgutil.get_importer(script) is None:
--- a/drgn/cli.py
+++ b/drgn/cli.py
@@ -21,9 +21,26 @@ from drgn.internal.repl import interact, readline
 from drgn.internal.rlcompleter import Completer
 from drgn.internal.sudohelper import open_via_sudo
 
-__all__ = ("run_interactive", "version_header")
+__all__ = ("default_globals", "run_interactive", "version_header")
 
 logger = logging.getLogger("drgn")
+
+# The list of attributes from the drgn module which are imported and inserted
+# into the global namespace for interactive debugging.
+_DRGN_GLOBALS = [
+    "FaultError",
+    "NULL",
+    "Object",
+    "alignof",
+    "cast",
+    "container_of",
+    "execscript",
+    "implicit_convert",
+    "offsetof",
+    "reinterpret",
+    "sizeof",
+    "stack_trace",
+]
 
 
 class _LogFormatter(logging.Formatter):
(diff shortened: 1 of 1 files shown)


drgn/cli.py
```diff
@@ -353,7 +403,9 @@ def _main() -> None:
 
     args = parser.parse_args()
 
-    if args.script:
+    if args.script and args.exec:
+        sys.exit("error: -e cannot be specified with script ARG")
```
osandov, line 407: I generally prefer to push checks like this into argparse instead, which we can do by putting these in a mutually exclusive group:

```diff
diff --git a/drgn/cli.py b/drgn/cli.py
index bebaaed4..63c542c2 100644
--- a/drgn/cli.py
+++ b/drgn/cli.py
@@ -386,19 +386,22 @@ def _main() -> None:
         const="none",
         help="don't print any logs or download progress",
     )
-    parser.add_argument(
+
+    script_group = parser.add_mutually_exclusive_group()
+    script_group.add_argument(
         "-e",
         dest="exec",
         metavar="CODE",
         help="an expression or statement to evaluate, instead of running in interactive mode",
     )
-    parser.add_argument(
+    script_group.add_argument(
         "script",
-        metavar="ARG",
+        metavar="SCRIPT [ARG...]",
         type=str,
         nargs=argparse.REMAINDER,
         help="script to execute instead of running in interactive mode",
     )
+
     parser.add_argument("--version", action="version", version=version)
 
     args = parser.parse_args()
```

(Renaming `ARG` -> `SCRIPT [ARG...]` makes the error message a little clearer.)

drgn/cli.py
```diff
@@ -386,6 +386,12 @@ def _main() -> None:
         const="none",
         help="don't print any logs or download progress",
     )
+    parser.add_argument(
+        "-e",
+        dest="exec",
```
osandov, line 391: Do we want a long option, too? I guess `python -c` doesn't have one, so I'm undecided.
```suggestion
        "-e",
        "--exec",
```

osandov (changes requested):
This is awesome and a long time coming, thanks! Just a couple of minor comments.

drgn/cli.py
brenns10: I totally agree with moving logic into argparse. But as I think about it, I realize there's one weakness with both our approaches: the remainder of the CLI args are no longer available to the code in `-e`. For instance, this actually works with python:

```
$ seq 10 | xargs python -c 'import sys; [print(f"python: {v}") for v in sys.argv]'
python: -c
python: 1
python: 2
python: 3
python: 4
python: 5
python: 6
python: 7
python: 8
python: 9
python: 10
```

I'm not sure what the practical use would be, but I think that symmetry with the Python `-c` option is a worthy goal. Instead of treating this as an error, we could just assume the user knows what they're doing, and do this instead:

```diff
diff --git a/drgn/cli.py b/drgn/cli.py
index bebaaed4..52e699f2 100644
--- a/drgn/cli.py
+++ b/drgn/cli.py
@@ -403,9 +403,7 @@ def _main() -> None:
 
     args = parser.parse_args()
 
-    if args.script and args.exec:
-        sys.exit("error: -e cannot be specified with script ARG")
-    elif args.script:
+    if args.script and not args.exec:
         # A common mistake users make is running drgn $core_dump, which tries
         # to run $core_dump as a Python script. Rather than failing later with
         # some inscrutable syntax or encoding error, try to catch this early
@@ -504,16 +502,17 @@ def _main() -> None:
             if new:
                 module.try_file(extra_symbol_path)
 
-    if args.script:
+    if args.exec:
+        sys.path.insert(0, "")
+        sys.argv = ["-e"] + args.script
+        exec(args.exec, default_globals(prog))
+    elif args.script:
         sys.argv = args.script
         script = args.script[0]
         if pkgutil.get_importer(script) is None:
             sys.path.insert(0, os.path.dirname(os.path.abspath(script)))
         drgn.set_default_prog(prog)
         runpy.run_path(script, init_globals={"prog": prog}, run_name="__main__")
-    elif args.exec:
-        sys.path.insert(0, "")
-        exec(args.exec, default_globals(prog))
     else:
         run_interactive(prog)
```
Thoughts?

drgn/cli.py
brenns10: I had thought about that for a bit before realizing that I didn't know exactly what `-e` even stands for...

- `--expression`, `--expr`
- `--evaluate`, `--eval`
- `--execute`, `--exec`

Three different words to choose from, each with a commonly used four-letter shortening. Technically `exec` is correct (the others imply only expressions, rather than statements, can be executed). But I'm not sure that distinction is obvious enough that `--execute` or `-e` stands out as correct. Especially given that you normally "execute" a script just like you'd "execute" a statement, so the most correct option doesn't even semantically highlight that the argument should be literal code, not a script.

So in my opinion, leaving it without a long option is best.

drgn/cli.py
osandov: Excellent point, yes, let's do that.

septatrix:
Why use `-e` and not the same flag as either Python or GDB?

brenns10:
> Why use `-e` and not the same flag as either Python or GDB?

Python uses `-c`, which is taken by drgn to specify the core dump file. So we can't use that, unfortunately.

GDB uses `-ex` or `--eval-command=`, but this isn't exactly equivalent: the command is executed, and then GDB still enters interactive mode. In order to achieve the same behavior here, you need to add `--batch`. The `-ex` arg is not in the getopt() / getopt_long() style that argparse offers, and that Drgn has always used, so I'm also not sure whether it's feasible.

So I think this does the best we can given the circumstances.

osandov (approved):

osandov:
> Python uses -c, which is taken by drgn to specify the core dump file.

And I deeply regret this :sweat: Merged, thanks!

PR #481, opened 2025-03-14
Use shell builtin instead of which

septatrix:
`command -v` is mandated by POSIX whereas `which` might not always be preset, e.g. in VMs


osandov:
Sounds good to me, any reason not to do it for the other calls to `which` in this same document?

septatrix:
> Sounds good to me, any reason not to do it for the other calls to `which` in this same document?

I use Fedora and stopped reading further :D (So no, I should also do it for all other occurrences)

Issue #483, opened 2025-03-17
Fix handling of userspace core dumps with truncated `pr_fname`


osandov:
@brenns10 reported the following error from a core dump of a process with a name longer than 16 characters in #480:

```
$ python -m drgn -q -c core.5063
drgn 0.0.30+88.g40f809c7.dirty (using Python 3.9.21, elfutils 0.191, with libkdumpfile)
For help, type help(drgn).
>>> import drgn
>>> from drgn import FaultError, NULL, Object, alignof, cast, container_of, execscript, implicit_convert, offsetof, reinterpret, sizeof, stack_trace
>>> from drgn.helpers.common import *
>>> next(prog.threads()).stack_trace()
Traceback (most recent call last):
  File "<console>", line 1, in <module>
Exception: pr_fname is not null terminated
```

This is because `pr_fname` is limited to 16 bytes, and the name is truncated if it is longer than that. I mistakenly assumed that if it was truncated, it would be truncated with a null byte, but apparently this isn't the case.

Unfortunately, `get_prpsinfo_fname()` relies on using the name directly from the core dump data, so fixing this will require allocating a copy of `pr_fname` that we guarantee is null-terminated (similar to earlier revisions of #432 before I gave bad advice :) ).

brenns10:
Moving my comment over:

> Yep, I've got that more or less working here. I'm looking at how to have a unit test case for it, but I think all the test cores are resources rather than being dynamically generated, and it looks like none of the base filenames were longer than 16 bytes.
> ```diff
> diff --git a/libdrgn/program.c b/libdrgn/program.c
> index 396cc013..298e9190 100644
> --- a/libdrgn/program.c
> +++ b/libdrgn/program.c
> @@ -136,6 +136,7 @@ void drgn_program_init(struct drgn_program *prog,
>  
>  void drgn_program_deinit(struct drgn_program *prog)
>  {
> +	free(prog->core_dump_fname_cached);
>  	drgn_thread_set_deinit(&prog->thread_set);
>  	/*
>  	 * For userspace core dumps, main_thread and crashed_thread are in
> @@ -924,7 +925,7 @@ static struct drgn_error *get_prpsinfo_pid(struct drgn_program *prog,
>  
>  static struct drgn_error *get_prpsinfo_fname(struct drgn_program *prog,
>  					   const char *data, size_t size,
> -					   const char **ret)
> +					   char **ret)
>  {
>  	bool is_64_bit;
[...]

PR #484, opened 2025-03-18
Handle truncated pr_fname

brenns10:
It appears that some core dumps may have a "pr_fname" which isn't nul-terminated, which leads to the following exception, even if you're not interested in the thread name:

```
>>> next(prog.threads()).stack_trace()
Exception: pr_fname is not null terminated
```

Handle this by creating a copy and nul-terminating it.

Closes: #483
Fixes: e6cc9b07 ("Add name to drgn.Thread")


brenns10:
It wouldn't be a PR from me if I didn't force-push it 3 times in rapid succession 😭 sorry lol

osandov:
Thanks for the quick fix!

PR #485, opened 2025-03-25
Fix CLI failure on Python < 3.10 and add smoke test

brenns10:
Looks like the CLI broke on Python < 3.10 with 4a6a9f33.

```
  File "/usr/lib64/python3.9/runpy.py", line 197, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib64/python3.9/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/home/stepbren/repos/drgn/drgn/__main__.py", line 14, in <module>
    from drgn.cli import _main
  File "/home/stepbren/repos/drgn/drgn/cli.py", line 193, in <module>
    class _TrySymbolsByBaseAction(argparse.Action):
  File "/home/stepbren/repos/drgn/drgn/cli.py", line 214, in _TrySymbolsByBaseAction
    **_bool_options(False),
TypeError: 'staticmethod' object is not callable
```

The fix is straightforward (though if you prefer a different way that's fine). I went ahead and added a smoke test that just runs the CLI in the simplest possible way. I verified it failed on 3.9. Then I added the fix.

The CLI is fun because we don't have any test coverage, which is fine because it's easy to manually test, but we don't get the variety of Python versions when we manually test.

### DIFF
--- a/drgn/cli.py
+++ b/drgn/cli.py
@@ -190,24 +190,24 @@ def _displayhook(value: Any) -> None:
     setattr(builtins, "_", value)
 
 
+def _bool_options(value: bool) -> Dict[str, Tuple[str, bool]]:
+    return {
+        option: ("try_" + option.replace("-", "_"), value)
+        for option in (
+            "module-name",
+            "build-id",
+            "debug-link",
+            "procfs",
+            "embedded-vdso",
+            "reuse",
+            "supplementary",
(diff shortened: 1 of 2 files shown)


osandov:
That's embarrassing, thanks!

PR #486, opened 2025-03-25
cli: fix default program for -e and handle !isatty(stdin)

osandov:
One fix for `-e`, plus a fix for issues with piping into drgn that a couple of people have reported recently (and a small cleanup).


osandov:
I'm going to rebase this on #485 and add some tests.

PR #487, opened 2025-03-28
page_pool page leak contrib scripts

dtatulea:
Here are the scripts used in my Netdevconf 0x19 talk.

### DIFF
--- a/contrib/pp_leak/scan_tcp_socks.py
+++ b/contrib/pp_leak/scan_tcp_socks.py
@@ -1,0 +1,90 @@
+#!/usr/bin/env drgn
+
+# Copyright (c) 2025 NVIDIA Corporation & Affiliates
+# SPDX-License-Identifier: LGPL-2.1-or-later
+
+help="""
+Detect leaked page_pool pages by scanning TCP sockets for SKBs from the
+receive queue that are using such leaked pages.
+
+It is a good idea to filter out by interface to reduce the run time
+of the script.
+"""
+
+
+import sys
+import argparse
+
+from drgn import (
+    Object,
+    FaultError,
+)
+from drgn.helpers.linux import (
+    hlist_nulls_empty,
+    sk_nulls_for_each,
+)
+from drgn.helpers.linux.mm import virt_to_page
+from drgn.helpers.linux.net import (
+    netdev_get_by_name,
+    skb_shinfo,
+    is_pp_page
+)
+
+
+def get_opts():
+    parser = argparse.ArgumentParser(description=help)
+    parser.add_argument(
+        "-i", "--interface", default=None, type=str, help="Filter by interface name.")
+
+    args = parser.parse_args()
+    return args
+
+
+opts = get_opts()
+ifindex = -1
+if opts.interface:
+    netdev = netdev_get_by_name(opts.interface)
+    if netdev.value_() == 0:
+        print(f"Netdev interface '{opts.interface}' not found.")
+        sys.exit(1)
+
+    ifindex = netdev.ifindex
+
+tcp_hashinfo = prog.object("tcp_hashinfo")
+
+for i in range(tcp_hashinfo.ehash_mask + 1):
+    head = tcp_hashinfo.ehash[i].chain
+    if hlist_nulls_empty(head):
+        continue
+
+    for sk in sk_nulls_for_each(head):
+
+        # Filter by interface:
+        if ifindex > 0 and sk.sk_rx_dst_ifindex.value_() != ifindex:
+            continue
+
+        first_skb = sk.sk_receive_queue.next
+        skb = first_skb
+        while skb != None:
+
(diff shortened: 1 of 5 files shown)


drgn/helpers/linux/net.py
```diff
@@ -260,3 +270,20 @@ def skb_shinfo(skb: Object) -> Object:
...
         return cast("struct skb_shared_info *", skb.end)
+
+
+def is_pp_page(page: Object) -> bool:
+    """
+    Check if given page is a page_pool page.
+
+    :param page: ``struct page *``
+    """
+    PP_SIGNATURE = 0xDEAD000000000040
```
osandov, line 281: This value apparently varies by architecture. You can copy this into this file:

```python3
def _poison_pointer_delta(prog: Program) -> int:
    # The value of POISON_POINTER_DELTA depends on
    # CONFIG_ILLEGAL_POINTER_VALUE, which varies by architecture and kernel
    # version. To avoid hard-coding values, derive the value from
    # TIMER_ENTRY_STATIC, which we can find in any statically-defined timer.
    # This still requires hard-coding an offset, but that offset is the same on
    # all architectures and kernel versions since Linux kernel commit
    # b8a0255db958 ("include/linux/poison.h: use POISON_POINTER_DELTA for
    # poison pointers") (in v4.5).
    try:
        return prog.cache["POISON_POINTER_DELTA"]
    except KeyError:
        pass
    TIMER_ENTRY_STATIC = prog["poll_spurious_irq_timer"].entry.next.value_()
    POISON_POINTER_DELTA = TIMER_ENTRY_STATIC - 0x300
    prog.cache["POISON_POINTER_DELTA"] = POISON_POINTER_DELTA
    return POISON_POINTER_DELTA
```

Then define it as

```python
PP_SIGNATURE = _poison_pointer_delta(page.prog_) + 0x40
```

drgn/helpers/linux/net.py
```diff
@@ -260,3 +270,20 @@ def skb_shinfo(skb: Object) -> Object:
...
+    PP_MAGIC_MASK = ~0x3
+
+    try:
+        return (page.pp_magic & PP_MAGIC_MASK) == PP_SIGNATURE
+    except FaultError:
+        return False
```
osandov, line 287: In my opinion, `FaultError` handling belongs in the caller, not in this helper.

contrib/pp_leak/scan_tcp_socks.py
```diff
@@ -0,0 +1,91 @@
...
+            continue
+
+        first_skb = sk.sk_receive_queue.next
+        skb = first_skb
+        while skb != None:
+
+            try:
+                # Check linear part of skb:
+                page = virt_to_page(skb.data)
+                if page.pp_magic == PP_SIGNATURE and page.pp.user.detach_time:
```
osandov, line 74: Can this use your new helper?

osandov (changes requested):
Sorry for the delay, thanks for these! A few minor comments. I also wrote a test case for the helper: https://gist.github.com/osandov/c9b1d0fbd86a5c07ebc2e200ba4f4738. I can apply that after merging or you can fold it in, your choice.

drgn/helpers/linux/net.py
dtatulea: Ack. Will do.

drgn/helpers/linux/net.py
```diff
@@ -260,3 +270,20 @@ def skb_shinfo(skb: Object) -> Object:
         return cast("struct skb_shared_info *", skb.head + skb.end)
     else:
         return cast("struct skb_shared_info *", skb.end)
+
+
+def is_pp_page(page: Object) -> bool:
```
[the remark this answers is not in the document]
dtatulea: Ack. Will add.

drgn/helpers/linux/net.py
dtatulea: I don't disagree. But half of the convenience of this function was to handle the FaultError. I can pull it back in the callers. Your call.

contrib/pp_leak/scan_tcp_socks.py
dtatulea: Yes, I missed this one.

dtatulea: Yes.

dtatulea:
Fixed the requested changes. LMK if you need me to squash and force push.

osandov:
Thanks, this looks good. Please do squash and force push. It also looks like some of your committer emails don't match your signed-off-by emails.

dtatulea:
> Thanks, this looks good. Please do squash and force push. It also looks like some of your committer emails don't match your signed-off-by emails.

Forced push with fixed Signed-off-by tags.

osandov:
Thank you!

PR #496, opened 2025-05-06
drgn.helpers.linux.fs: skip cursors in for_each_mount()

brenns10:
Details are in the commit message. I'm open to a more robust method of detecting cursors, but this did the job and was simple enough.

A follow up for this is to see if I can convert the mount flags to an enum. They are currently held in an `int` so it seems reasonable, but there's only space for 4 more bits. C23 cannot come fast enough for the Linux kernel 🚀

### DIFF
--- a/drgn/helpers/linux/fs.py
+++ b/drgn/helpers/linux/fs.py
@@ -374,7 +374,8 @@ def for_each_mount(
         )
     for mnt in mounts:
         if (
-            (src is None or mount_src(mnt) == src)
+            mnt.mnt.mnt_sb  # skip cursors (v5.8 - v6.8)
+            and (src is None or mount_src(mnt) == src)
             and (dst is None or mount_dst(mnt) == dst)
             and (fstype is None or mount_fstype(mnt) == fstype)
         ):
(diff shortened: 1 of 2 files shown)


osandov:
Merged, thanks for the excellent test case and the fix!

> A follow up for this is to see if I can convert the mount flags to an enum.

Christian Brauner was sympathetic to torvalds/linux@b2b4483b5d05026218127fc8f38c69adf69c235b, so this would be great.

> They are currently held in an `int` so it seems reasonable, but there's only space for 4 more bits.

You probably already know this, but enums can expand to be larger than an `int` (in an implementation-defined manner), so that shouldn't be a deal breaker.

brenns10:
> You probably already know this, but enums can expand to be larger than an int (in an implementation-defined manner), so that shouldn't be a deal breaker.

I was not aware. I think I was thinking specifically of [CTF v3](https://raw.githubusercontent.com/wiki/oracle/binutils-gdb/files/ctf-spec.pdf) which sadly doesn't represent enums larger than an int. Thankfully, BTF patches that hole with "enum64" and so it shouldn't be too much of an issue for the future.

PR #498, opened 2025-05-10
libdrgn: python: fix reference leak for held types

brenns10:
Python "LazyObjects" (e.g. TypeMember) contain a callable or an object, either of which may contain a reference to the Program. When cached on the program, these references form a cycle. Add tp_traverse methods to help detect these cycles and avoid resource leakages.

### DIFF
--- a/libdrgn/python/object.c
+++ b/libdrgn/python/object.c
@@ -1635,6 +1635,12 @@ static PyMappingMethods DrgnObject_as_mapping = {
 	.mp_subscript = (binaryfunc)DrgnObject_subscript,
 };
 
+static int DrgnObject_traverse(DrgnObject *self, visitproc visit, void *arg)
+{
+	Py_VISIT(DrgnObject_prog(self));
+	return 0;
+}
+
 PyTypeObject DrgnObject_type = {
 	PyVarObject_HEAD_INIT(NULL, 0)
 	.tp_name = "_drgn.Object",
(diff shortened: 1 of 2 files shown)


osandov:
Wow, awesome investigation. Apparently I got mixed up and thought that immutable objects don’t need `tp_traverse`, but that only applies to `tp_clear`. There are probably more types missing `tp_traverse`, but I’ll merge this after taking another look in the morning and audit the rest after.

brenns10:
Thanks and yeah, I think we may need a few more of these. I also wonder if it might be worth having some debug build configurations for detecting program leaks? EG increment a counter at program alloc time, decrement at dealloc, and have an atexit() handler for detecting if the counter is non-zero? Purely a debug build configuration of course.

Or maybe Python has something like this built-in. It would be worth including for tests in CI or something like that.

osandov:
That's a great idea. Actually, we currently track all of the live programs for our `logging` hack, so we don't even need a debug build: https://github.com/osandov/drgn/blob/f8bedb963def7436ec8920397c40ffcf915ef52d/libdrgn/python/program.c#L159

As a proof-of-concept, I wrote a `table_for_each()` helper (which reminds me I should revive your `drgn.helpers.drgn` PR) and compared the set of programs before and after running tests: https://gist.github.com/osandov/b4cdab8e61476dd5cfa306dab5d4c7cc

Before your change, I get 1484 leaks :scream: This PR brings it down to 19!

osandov:
Just landed a fix for the rest of the leaks. Thanks again for catching this!

brenns10:
Awesome, and thanks for the rest of the fixes, I was definitely at the edge of my Python C API knowledge here :)

PR #499, opened 2025-05-13
Add contrib/pstack.py script (and helper for iterating over process threads)

brenns10:
I went ahead and dropped the `pstack.py` script from the `.gnu_debugdata` pull request over at #466, because it's not totally relevant. I've also been doing a lot of work on the pstack contrib script, so the copy there was outdated anyway. This version supports two modes:

```
# Directly print stack traces to stdout
drgn contrib/pstack.py pstack [select tasks...]

# Dump necessary userspace stack pages to files
drgn contrib/pstack.py dump $OUTDIR [select tasks...]
# Print stack traces using the information
python -m contrib.pstack print $OUTDIR
```

The second mode is better for use in the kdump environment, because it does not try to use the userspace ELF files directly, it just dumps a few pages.

### DIFF
--- a/drgn/helpers/linux/pid.py
+++ b/drgn/helpers/linux/pid.py
@@ -19,13 +19,14 @@ from _drgn import (
 from drgn import IntegerLike, Object, Program, cast, container_of
 from drgn.helpers.common.prog import takes_object_or_program_or_default
 from drgn.helpers.linux.idr import idr_for_each
-from drgn.helpers.linux.list import hlist_for_each_entry
+from drgn.helpers.linux.list import hlist_for_each_entry, list_for_each_entry
 
 __all__ = (
     "find_pid",
     "find_task",
     "for_each_pid",
     "for_each_task",
+    "for_each_task_in_group",
     "pid_task",
 )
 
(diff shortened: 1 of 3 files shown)


brenns10:
I'm actually quite pleased at the compression of the stack data, it ends up being that the stacks saved via `pstack.py dump` can be smaller than the resulting output of `pstack.py print`. I guess it's because we're adding information from the ELF files, and representing the information in a less dense way. EG:
```
stephen at augustus in ~/repos/drgn (pstack)
$ python -m drgn -k contrib/pstack.py dump out.3 -p $(pgrep firefox)

stephen at augustus in ~/repos/drgn (pstack)
$ ls -lh out.3/
total 188K
-rw-r--r--. 1 stephen stephen 170K May 13 23:11 22630.gz
-rw-r--r--. 1 stephen stephen  13K May 13 23:11 22630-meta.json.gz

stephen at augustus in ~/repos/drgn (pstack)
$ python -m contrib.pstack print out.3 | wc -c
296220
````

Though this could be a bit cherry-picked since Rust stack traces tend to have function names that are about a mile long. In any case, it's cool. Definitely worth trying on Fedora due to the `.eh_frame` availability, and if you pull in the `.gnu_debugdata` support then it's even better :)

File: drgn/helpers/linux/pid.py [hunk 1]
```diff
@@ -102,3 +103,41 @@ def for_each_task(prog: Program, ns: Optional[Object]) -> Iterator[Object]:
...
+        )
+    else:
+        # Since commit 8e1f385104ac0 ("kill task_struct->thread_group") from
+        # 6.7, the thread_group list is gone, replaced by a list inside the
+        # task.signal struct. This has an explicit list_head (unlike the
+        # thread_group which just linked each task together with no explicit
+        # head node).
+        for other in list_for_each_entry(
+            "struct task_struct",
+            task.signal.thread_head.address_of_(),
```
osandov, line 137: `task->signal->thread_head` has existed since torvalds/linux@0c740d0afc3bff0a097ad03a1c8df92757516f5c in v3.14, is there a reason we can't use that unconditionally?

PR #506, opened 2025-06-17
Enhance get_dmesg() and print_dmesg()

YassineLr:
Add option for human-readable timestamps to get_dmesg() and print_dmesg() by using timekeeping helpers

### DIFF
--- a/drgn/helpers/linux/printk.py
+++ b/drgn/helpers/linux/printk.py
@@ -15,8 +15,11 @@ from typing import TYPE_CHECKING, Dict, List, NamedTuple, Optional, Tuple
 if TYPE_CHECKING:
     from _typeshed import SupportsWrite
 
+from datetime import datetime
+
 from drgn import Object, Program, cast, sizeof
 from drgn.helpers.common.prog import takes_program_or_default
+from drgn.helpers.linux.timekeeping import ktime_get_coarse_ns, ktime_get_coarse_real_ns
 
 __all__ = (
     "get_dmesg",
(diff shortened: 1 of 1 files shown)


drgn/helpers/linux/printk.py
```diff
@@ -243,7 +246,7 @@ def get_printk_records(prog: Program) -> List[PrintkRecord]:
 
 
 @takes_program_or_default
-def get_dmesg(prog: Program) -> bytes:
+def get_dmesg(prog: Program, human_readable_time: bool = False) -> bytes:
```
osandov, line 249: I'd prefer making this parameter keyword-only:
```suggestion
def get_dmesg(prog: Program, *, human_readable_time: bool = False) -> bytes:
```
It should also be documented (i.e., `:param human_readable_time: description...` in the docstring).

drgn/helpers/linux/printk.py
```diff
@@ -259,21 +262,44 @@ def get_dmesg(prog: Program) -> bytes:
...
+        lines.append(b"")  # So we get a trailing newline.
     return b"\n".join(lines)
 
 
 @takes_program_or_default
-def print_dmesg(prog: Program, *, file: "Optional[SupportsWrite[str]]" = None) -> None:
+def print_dmesg(
+    prog: Program,
+    human_readable_time: bool = False,
+    *,
```
osandov, line 300: Same here, keyword-only:
```suggestion
    *,
    human_readable_time: bool = False,
```

osandov (changes requested):
This is awesome, thank you! A few small comments.

osandov (approved):

osandov:
Merged, thanks!

PR #508, opened 2025-06-18
libdrgn: always create all modules in load_debug_info()

brenns10:
I'm not certain this is the best way to solve this, but I hate filing issues without an attempted fix.

As far as I can tell, the choice to not create all modules when default=False is a performance optimization. So I think it could be dropped given that it impacts correctness when relying on built-in ORC for unwinding.


brenns10:
Thanks! And sorry for spoiling that optimization :(

PR #516, opened 2025-07-02
Add get_task_rss helper

YassineLr:


drgn/helpers/linux/rss.py
```diff
@@ -0,0 +1,148 @@
...
+    shared memory pages (those which aren't associated with on-disk files, but
+    belonging to shared memory mappings). This function returns a tuple
+    containing each category, but the common behavior is to use the "total"
+    value which sums them up.
+
+    :param task: ``struct task_struct *`` for which to compute RSS
+    :param cache: if provided, we can use this to cache the mapping of
+      "mm_struct" to RSS. This helps avoid re-computing the RSS value for
+      processes with many threads, but note that it could result in out of date
+      values on a live system.
```
osandov, line 92: I would simply drop the `cache` parameter. drgn itself doesn't usually cache things for you (`identify_address()` is a notable exception). drgn-tools can wrap this with its own cache if needed.

drgn/helpers/linux/rss.py
```diff
@@ -0,0 +1,148 @@
...
+    MM_ANONPAGES = prog.constant("MM_ANONPAGES").value_()
+    MM_SWAPENTS = prog.constant("MM_SWAPENTS").value_()
+    try:
+        MM_SHMEMPAGES = prog.constant("MM_SHMEMPAGES").value_()
+    except LookupError:
+        MM_SHMEMPAGES = -1
+
+    # Start with the counters from the mm_struct
+    filerss = anonrss = shmemrss = swapents = 0
+
```
osandov, line 116: These initializations seem redundant. You can delete these and then initialize them with ` = ` instead of ` += ` in the `else` branch below.

```suggestion
```

drgn/helpers/linux/rss.py
```diff
@@ -0,0 +1,148 @@
...
+        shmemrss = 0
+        if MM_SHMEMPAGES >= 0:
+            shmemrss = percpu_counter_sum(rss_stat[MM_SHMEMPAGES].address_of_())
+    else:
+        # Prior to this, the "rss_stat" was a structure containing counters that
+        # were cached on each task_struct and periodically updated into the
+        # mm_struct. We start with the counter values from the mm_struct and
+        # then sum up the cached copies from each thread.
+        filerss += rss_stat.count[MM_FILEPAGES].counter.value_()
+        anonrss += rss_stat.count[MM_ANONPAGES].counter.value_()
```
osandov, line 133: Is there a case missing for `MM_SWAPENTS` here?

drgn/helpers/linux/rss.py
```diff
@@ -0,0 +1,148 @@
...
+    if not task.mm:
+        return TaskRss(0, 0, 0, 0)
+
+    prog = task.prog_
+    rss_stat = task.mm.rss_stat
+
+    MM_FILEPAGES = prog.constant("MM_FILEPAGES").value_()
+    MM_ANONPAGES = prog.constant("MM_ANONPAGES").value_()
+    MM_SWAPENTS = prog.constant("MM_SWAPENTS").value_()
+    try:
```
osandov, line 109: This needs a comment explaining why it's optional. Something like

```suggestion
    # This counter was added in Linux kernel commit eca56ff906bd ("mm, shmem:
    # add internal shmem resident memory accounting") (in v4.5).
    try:
```

drgn/helpers/linux/rss.py
```diff
@@ -0,0 +1,148 @@
...
+            "thread_node",
+        ):
+            # We've already yielded "task" (or not, depending on the caller's
+            # preference) so skip it here.
+            if other != task:
+                yield other
+
+
+@takes_program_or_default
+def get_task_rss(
```
osandov, line 74: It looks like the closest thing the kernel has to this helper is the `get_mm_rss()` function: https://github.com/torvalds/linux/blob/master/include/linux/mm.h#L2609. Maybe we can rename this to `get_task_rss_info()` or something like that to indicate that it returns a tuple instead of just a single counter? That leaves the possibility open for adding `get_mm_rss()`, `get_mm_rss_info()`.

drgn/helpers/linux/rss.py
```diff
@@ -0,0 +1,148 @@
+# Copyright (c) 2025 Oracle and/or its affiliates
+# SPDX-License-Identifier: LGPL-2.1-or-later
+
+
+"""
+The ``drgn.helpers.linux.rss`` module provides get_task_rss helper to get resident set size
```
osandov, line 6: This helper can just go in `drgn.helpers.linux.mm`.

osandov (changes requested):
Oops, my fix for the mypy error on 3.8 introduced a new one on 3.13. I just force-pushed a new fix.

PR #517, opened 2025-07-03
Add get_task_rss_info helper

YassineLr:


drgn/helpers/linux/mm.py
```diff
@@ -1435,3 +1437,123 @@ def in_direct_map(prog: Program, addr: IntegerLike) -> bool:
...
+        if MM_SHMEMPAGES >= 0:
+            shmemrss = rss_stat.count[MM_SHMEMPAGES].counter.value_()
+
+        for gtask in for_each_task_in_group(task, include_self=True):
+            filerss += gtask.rss_stat.count[MM_FILEPAGES].value_()
+            anonrss += gtask.rss_stat.count[MM_ANONPAGES].value_()
+            swapents += gtask.rss_stat.count[MM_SWAPENTS].value_()
+            if MM_SHMEMPAGES >= 0:
+                shmemrss += gtask.rss_stat.count[MM_SHMEMPAGES].value_()
+    rss = TaskRss(filerss, anonrss, shmemrss, swapents)
```
osandov, line 1557: There's no need for this variable assignment anymore, just return it directly.
```suggestion
    return TaskRss(filerss, anonrss, shmemrss, swapents)
```

osandov (changes requested):
Thanks! A couple of comments inline and a couple more things:

I just merged Stephen's `for_each_task_in_group()` helper, so you can rebase on that and drop your copy.

This also needs a test case. Maybe create a process with `fork_and_stop()`, parse its `/proc/pid/meminfo`, and compare that with the output of the helper, maybe with some allowed error tolerance to allow for racing with swap.

YassineLr:
I added a test with a 3-page error tolerance (just rough estimate) . Thanks!

YassineLr:
thanks for the suggested changes! I went through them and made the updates.
I also wanted to ask if helpers like get_vmem() (that returns virtual memory size of the task), is_thread(), is_user(), and is_kthread() would be useful to add ?

PR #519, opened 2025-07-10
Add Ps crash command

YassineLr:


YassineLr:
Hey, I added some helper functions but couldn’t find kernel equivalents for example they usually access virtual memory size directly from task_struct rather than wrapping it in a helper.
I haven’t added any tests yet -- waiting to see if you think these helpers are worth keeping. Thank you !

drgn/helpers/linux/pid.py
```diff
@@ -123,3 +125,39 @@ def for_each_task_in_group(
     ):
         if other != task or include_self:
             yield other
+
+
+def is_kthread(t: Object) -> bool:
```
osandov, line 130: I think `task_is_kthread()` would be a better name for this one, `task` would be a better name for the argument, and the helper would make more sense in `drgn.helpers.linux.kthread`.

drgn/helpers/linux/pid.py
```diff
@@ -123,3 +125,39 @@ def for_each_task_in_group(
...
+    """
+    Check if a task is kernel thread.
+    """
+    if not t.mm and not task_state_to_char(t) == "Z":
+        return True
+    return False
```
osandov, line 136: I'm torn about this implementation. It makes sense, but it's not obvious to me that it will always be correct. Even though I don't like hard-coding constants, I'm leaning towards:

```suggestion
    # This hasn't changed since Linux kernel commit 246bb0b1deb2 ("kill
    # PF_BORROWED_MM in favour of PF_KTHREAD") (in v2.6.27).
    PF_KTHREAD = 0x200000
    return bool(t.flags.value_() & PF_KTHREAD)
```

Since this seems to have come from drgn-tools, @SiddhiK29 or @brenns10 do you have any thoughts?

drgn/helpers/linux/pid.py
```diff
@@ -123,3 +125,39 @@ def for_each_task_in_group(
...
+
+def is_user(t: Object) -> bool:
+    """
+    Check if a task is user thread.
+    """
+    if t.mm:
+        return True
+    elif task_state_to_char(t) == "Z":
+        return True
+    return False
```
osandov, line 147: This is just `not task_is_kthread(t)`, so I'd prefer to drop it.

drgn/helpers/linux/pid.py
```diff
@@ -123,3 +125,39 @@ def for_each_task_in_group(
...
+
+
+def is_group_leader(t: Object) -> bool:
+    """
+    Check if a task is thread group leader.
+    """
+    if t.pid.value_() == t.tgid.value_():
+        return True
+    return False
```
osandov, line 156: The equivalent in the function in the kernel is [`thread_group_leader()`](https://github.com/torvalds/linux/blob/347e9f5043c89695b01e66b3ed111755afcf1911/include/linux/sched/signal.h#L700-L703), which has a drastically different implementation:

```c
static inline bool thread_group_leader(struct task_struct *p)
{
	return p->exit_signal >= 0;
}
```

I think we should base our helper on that, and put it in `drgn.helpers.linux.sched`.

drgn/helpers/linux/pid.py
```diff
@@ -123,3 +125,39 @@ def for_each_task_in_group(
...
+
+
+def get_command(task: Object) -> str:
+    """
+    :returns: name of the command
+    """
+    return escape_ascii_string(task.comm.string_())
```
osandov, line 163: This one is trivial, so I'd rather not add it as a helper. It can go in `ps.py` if it makes your life easier, or you can just fold it into the caller. The `--drgn` output can just suggest `task.comm`.

drgn/helpers/linux/mm.py
```diff
@@ -1523,3 +1526,16 @@ def get_task_rss_info(prog: Program, task: Object) -> TaskRss:
...
+    """
+    Return virtual memory size of the task
+    """
+    prog = task.prog_
+    page_size = prog["PAGE_SIZE"].value_()
+    try:
+        vmem = (task.mm.total_vm.value_() * page_size) // ByteToKB
+    except drgn.FaultError:
+        vmem = 0
+    return vmem
```
osandov, line 1541: A few comments on this one:

* We can probably name this after [`task_vsize()`](https://github.com/torvalds/linux/blob/347e9f5043c89695b01e66b3ed111755afcf1911/fs/proc/task_mmu.c#L86-L89).
* Kilobytes are a pretty arbitrary unit to return from a helper. Let's just return bytes (like `task_vsize()` does, and also delete the definition of `ByteToKB`).
* Also, if the intention of the `FaultError` handling is to handle kernel threads with a `NULL` `mm`, it'd make more sense to check for that explicitly instead.

osandov (changes requested):
I only looked at the helpers this time around, a few comments there. I know you're just adapting these from drgn-tools, so thanks for your patience!

drgn/helpers/linux/mm.py
brenns10: I feel like I remember searching for a constant like this and not finding it!

I'm not against switching over. I do feel like it's easier to change a task constant than it is to change behavior, so there's an argument to be made that this implementation is less likely to become incorrect over time.

But I guess there's nothing that says that `task_state_to_char(t)` for a kthread _couldn't_ return Z at some point in the future, it's not like that's an invariant specified anywhere in the kernel.

YassineLr:
Hey @osandov @brenns10 Apologies for the delay. I’ve addressed the requested changes, please let me know if there’s anything else you’d like me to adjust.

YassineLr:
Hey @osandov  @brenns10 , sorry for the delay. I’ve addressed the requested changes and looking forward to your review.

drgn/helpers/linux/mm.py
```diff
@@ -1590,3 +1591,17 @@ def get_task_rss_info(prog: Program, task: Object) -> TaskRss:
...
+    Return virtual memory size of the task in bytes.
+
+    :param task: ``struct task_struct *``
+    """
+    prog = task.prog_
+    page_size = prog["PAGE_SIZE"].value_()
+    vmem = 0
+    if task.mm:
+        vmem = task.mm.total_vm.value_() * page_size
+    return vmem
```
osandov: Read the `mm` once to avoid the extra read round-trip and avoid a potential race where it changes:
```suggestion
    mm = task.mm.read_()
    if not mm:
        return 0
    return mm.total_vm.value_() * page_size
```

drgn/helpers/linux/sched.py
```diff
@@ -209,3 +210,12 @@ def task_since_last_arrival_ns(task: Object) -> int:
     rq_clock = task_rq(task).clock.value_()
 
     return rq_clock - arrival_time
+
+
+def thread_group_leader(t: Object) -> bool:
```
osandov: The parameter is documented as `task`, which is a better name anyways, so rename `t` -> `task`.

tests/linux_kernel/helpers/test_kthread.py
```diff
@@ -15,3 +22,13 @@ def test_kthread_data(self):
...
+
+    def test_task_is_kthread(self):
+        with fork_and_stop() as pid:
+            task = find_task(self.prog, pid)
+            is_kthread = task_is_kthread(task)
+
+            text = Path(f"/proc/{pid}/cmdline").read_text()
+            # kernel threads have an empty cmdline
+            expected = text == ""
+            self.assertEqual(is_kthread, expected)
```
osandov: A process you create from userspace will never be a kthread. An easier and more thorough test is checking the current process (which can't be a kthread) and `init_task` (which must be a kthread):
```suggestion
        self.assertFalse(task_is_kthread(find_task(self.prog, os.getpid())))
        self.assertTrue(task_is_kthread(prog["init_task"].address_of_()))
```

drgn/helpers/linux/mm.py
```diff
@@ -84,6 +83,8 @@
     "vma_find",
     "vmalloc_to_page",
     "vmalloc_to_pfn",
+    "get_task_rss_info",
```
osandov: It seems like this was accidentally moved.

drgn/commands/_builtin/crash/ps.py
```diff
@@ -0,0 +1,294 @@
...
+@crash_command(
+    description="process information",
+    long_description="display process status information",
+    arguments=(
+        mutually_exclusive_group(
+            argument(
+                "-u",
+                dest="user",
+                action="store_true",
+                default=False,
```
osandov: `default` is automatically `False` when `action="store_true"`, so you can drop this (and all of the other ones).
```suggestion
```

drgn/commands/_builtin/crash/ps.py
```diff
@@ -0,0 +1,294 @@
...
+        argument(
+            "-G",
+            dest="group_leader",
+            action="store_true",
+            default=False,
+            help="display only the thread group leader in a thread group",
+        ),
+        argument(
+            "arg",
+            nargs="?",
```
osandov: Crash seems to support passing multiple of these and includes the task if it matches any of them.

drgn/commands/_builtin/crash/ps.py
```diff
@@ -0,0 +1,294 @@
...
+        (
+            CellFormat("PID", ">"),
+            CellFormat("PPID", ">"),
+            CellFormat("CPU", "^"),
+            CellFormat("TASK", "^"),
+            CellFormat("ST", ">"),
+            CellFormat("%MEM", ">"),
+            CellFormat("%VSZ", ">"),
+            CellFormat("RSS", ">"),
+            CellFormat("COMM", "<"),
```
osandov: "<" is the default for strings, so this one is unnecessary to wrap

```suggestion
            "COMM",
```

drgn/commands/_builtin/crash/ps.py
```diff
@@ -0,0 +1,294 @@
...
+        elif args.arg[0] == "comm":
+            tasks = [
+                t for t in tasks if escape_ascii_string(t.comm.string_()) == args.arg[1]
+            ]
+    if args.user:
+        tasks = sorted(filter(lambda t: not task_is_kthread(t), tasks))
+    elif args.kernel:
+        tasks = sorted(filter(task_is_kthread, tasks))
+    if args.group_leader:
+        tasks = list(filter(thread_group_leader, tasks))
```
osandov: We don't need to sort here yet since the `show_*` functions already sort, right?

Also, list comprehensions are more Pythonic (and probably slightly more performant than `list(filter(...))`.

```suggestion
    if args.user:
        tasks = [task for task in tasks if not task_is_kthread(t)]
    elif args.kernel:
        tasks = [task for task in tasks if task_is_kthread(t)]
    if args.group_leader:
        tasks = [task for task in tasks if thread_group_leader(t)]
```

drgn/commands/_builtin/crash/ps.py
```diff
@@ -0,0 +1,294 @@
...
+        builder.add_from_import(
+            "drgn.helpers.linux.sched", "task_state_to_char", "task_cpu"
+        )
+        if args.last_run or args.last_run_timestamp:
+            builder.add_from_import(
+                "drgn.helpers.linux.sched", "task_since_last_arrival_ns"
+            )
+            builder.add_import("datetime")
+
+        builder.append("total_mem = int(totalram_pages(prog))\n")
```
osandov: `totalram_pages()` already returns an `int`, so the conversion is unnecessary.
```suggestion
        builder.append("total_mem = totalram_pages(prog)\n")
```

drgn/commands/_builtin/crash/ps.py
```diff
@@ -0,0 +1,294 @@
...
+                builder.append("tasks = filter(task_is_kthread, for_each_task(prog))\n")
+        else:
+            builder.append("tasks = for_each_task(prog)\n")
+
+        if args.group_leader:
+            builder.add_from_import(
+                "drgn.helpers.linux.sched",
+                "thread_group_leader",
+            )
+            builder.append("tasks = filter(thread_group_leader, tasks)\n")
```
osandov: Rather than applying these as filters, I think it'd look nicer to make them checks in the for loop. E.g., something like

```python
for task in for_each_task():
    if task_is_kthread(task):
        continue
    if not thread_group_leader(task):
        continue
```

It also needs to consider the pid/task/command checks.

drgn/commands/_builtin/crash/ps.py
```diff
@@ -0,0 +1,294 @@
...
+
+        builder.append("for task in tasks:\n")
+        builder.append("    cpu = task_cpu(task)\n")
+        builder.append("    state = task_state_to_char(task)\n")
+        builder.append("    task_rss = get_task_rss_info(task)\n")
+        builder.append("    total_rss = task_rss.total\n")
+        builder.append("    pct_mem = total_rss * 100 / total_mem\n")
+        builder.append("    vmem = task_vsize(task)\n")
+        builder.append("    command = escape_ascii_string(task.comm.string_())\n")
```
osandov: No need to convert this to a string here, just

```suggestion
        builder.append("    command = task.comm\n")
```
(And you can remove the `add_from_import` for `escape_ascii_string`).

drgn/commands/_builtin/crash/ps.py
```diff
@@ -0,0 +1,294 @@
...
+from drgn.commands import argument, drgn_argument, mutually_exclusive_group
+from drgn.commands.crash import CrashDrgnCodeBuilder, crash_command
+from drgn.helpers.common.format import (
+    CellFormat,
+    double_quote_ascii_string,
+    escape_ascii_string,
+    print_table,
+)
+from drgn.helpers.linux.kthread import task_is_kthread
+from drgn.helpers.linux.mm import get_task_rss_info, task_vsize, totalram_pages
```
osandov: FYI I renamed `get_task_rss_info()` to `task_rss_info()`, so you'll want to rebase and update that.

PR #522, opened 2025-07-15
Migrate runq related helpers from drgn-tools

richl9:


drgn/helpers/linux/runq.py
```diff
@@ -0,0 +1,81 @@
...
+
+    try:
+        arrival_time = task.last_run.value_()
+    except AttributeError:
+        try:
+            arrival_time = task.timestamp.value_()
+        except AttributeError:
+            arrival_time = task.sched_info.last_arrival.value_()
+
+    return arrival_time
```
brenns10, line 47: See [this comment](https://github.com/osandov/drgn/pull/519#discussion_r2205958751). I think we don't need this helper. It can be simplified to `task.sched_info.last_arrival` and inlined into the user below.

drgn/helpers/linux/runq.py
```diff
@@ -0,0 +1,81 @@
+# Copyright (c) 2025, Oracle and/or its affiliates.
+# SPDX-License-Identifier: LGPL-2.1-or-later
+
+"""
+runq
+-------------
+
+The ``drgn.helpers.linux.runq`` module provides helpers for working with the
+Linux runq.
```
brenns10, line 9: A few nits and suggestions on the docstring, since this does appear in the public docs.

- Need to match the underline length to the title above it. I think it triggers a Sphinx warning if not.
- In both the title and the sentence below, I'd suggest not using "runq". That's just a crash command name, not the name of a subsystem or structure. Instead, "run queues".

```suggestion
Run Queues
----------

The ``drgn.helpers.linux.runq`` module provides helpers for working with the
Linux run queues.
```

drgn/helpers/linux/runq.py
```diff
@@ -0,0 +1,81 @@
...
+
+def get_current_run_time(prog: Program, cpu: int) -> int:
+    """
+    Get running duration of the current task on some cpu
+
+    :param prog: drgn program
+    :param cpu: cpu index
+    :returns: duration in ns granularity
+    """
+    return task_lastrun2now(cpu_curr(prog, cpu))
```
brenns10, line 81: I think this helper is a bit too trivial as well, we could just inline it into whatever is using it.

drgn/helpers/linux/runq.py
```diff
@@ -0,0 +1,81 @@
...
+
+    :param prog: drgn program
+    :param task: ``struct task_struct *``
+    :returns: duration in ns granularity
+    """
+    prog = task.prog_
+    arrival_time = get_task_arrival_time(task)
+    rq_clock = runq_clock(prog, task_cpu(task))
+
+    return rq_clock - arrival_time
```
brenns10, line 70: This needs a unit test. Take a look at the function `fork_and_stop()` in `tests/linux_kernel/__init__.py` - you can use that to create a task in the live kernel, and you could use this helper and at least assert that the result is above or below certain thresholds.

brenns10 (commented):
Thanks Richard!

The context of this PR is that these helpers were used by both the `runq` and `ps` commands, so we thought it best to merge them first, prior to the commands themselves.

@osandov do you have an opinion about a new `runq.py` file, versus putting them in `sched.py`?

drgn/helpers/linux/runq.py
```diff
@@ -0,0 +1,47 @@
...
+def runq_clock(prog: Program, cpu: int) -> int:
+    """
+    Get clock of cpu runqueue ``struct rq``
+
+    :param prog: drgn program
+    :param cpu: cpu index
+    :returns: cpu runqueue clock in ns granularity
+    """
+    rq = per_cpu(prog["runqueues"], cpu)
+    return rq.clock.value_()
```
osandov, line 24: This helper is fairly simple, which I wouldn't mind if there was an equivalent function in the kernel, but that doesn't seem to be the case.

What if instead we add `cpu_rq()` and `task_rq()` helpers based on the existing kernel macros: https://github.com/torvalds/linux/blob/c7de79e662b8681f54196c107281f1e63c26a3db/kernel/sched/sched.h#L1346-L1348

Then, instead of:

```python3
rq_clock = runq_clock(prog, task_cpu(task))
```

You'd do:

```python3
rq_clock = task_rq(task).clock.value_()
```

Which is still an improvement over open-coding the whole thing.

drgn/helpers/linux/runq.py
```diff
@@ -0,0 +1,47 @@
...
+
+    The return duration will cover task's last run time on cpu and also
+    the time staying in current status, usually the time slice for task
+    on cpu will be short, so this can roughly tell how long this task
+    has been staying in current status.
+    For task status in "RU" status, if it's still on cpu, then this return
+    the duration time this task has been running, otherwise it roughly tell
+    how long this task has been staying in runqueue.
+
+    :param prog: drgn program
```
osandov, line 39: This doesn't have a `prog` parameter.

drgn/helpers/linux/runq.py
```diff
@@ -0,0 +1,47 @@
...
+    :param cpu: cpu index
+    :returns: cpu runqueue clock in ns granularity
+    """
+    rq = per_cpu(prog["runqueues"], cpu)
+    return rq.clock.value_()
+
+
+def task_lastrun2now(task: Object) -> int:
+    """
+    Get the duration from task last run timestamp to now
```
osandov, line 29: I had to read the kernel code to figure out whether this timestamp was for when a task last started running or stopped running. I'd be even more explicit here with something like "Get the number of nanoseconds since a task last started running."

drgn/helpers/linux/runq.py
```diff
@@ -0,0 +1,47 @@
...
+
+    The return duration will cover task's last run time on cpu and also
+    the time staying in current status, usually the time slice for task
+    on cpu will be short, so this can roughly tell how long this task
+    has been staying in current status.
+    For task status in "RU" status, if it's still on cpu, then this return
```
osandov, line 35: Is the "RU" status equivalent to the "R" state most tools use? If so, I'd use the more familiar terminology.

tests/linux_kernel/helpers/test_runq.py
```diff
@@ -0,0 +1,17 @@
...
+from drgn.helpers.linux.runq import task_lastrun2now
+from tests.linux_kernel import LinuxKernelTestCase, fork_and_stop
+
+
+class TestRunq(LinuxKernelTestCase):
+    def sleep_10ms():
+        time.sleep(0.01)
+
+    def test_task_lastrun2now(self):
+        with fork_and_stop(self.sleep_10ms) as pid:
```
osandov, line 15: If the forking process is the one sleeping, then it will sleep for 0.01 seconds, then wake up and run again, then stop, so I don't think this tests what you want. I think you want the main process to do the sleep so that the forked process stays stopped for at least that long:

```suggestion
        with fork_and_stop() as pid:
            time.sleep(0.01)
```

tests/linux_kernel/helpers/test_runq.py
```diff
@@ -0,0 +1,17 @@
...
+            task = find_task(self.prog, pid)
+            assert task_lastrun2now(task) >= 0.01 * 1e9
```
osandov, line 17: Use `self.assertGreaterEqual` to get better error messages when this fails.
```suggestion
            self.assertGreaterEqual(task_lastrun2now(task), 10000000)
```

osandov (changes requested):
Thanks, a few comments below. I think this would make more sense to keep in `drgn.helpers.linux.sched` instead of adding a new module.

The ideal naming for `task_lastrun2now()` is tricky since there's no equivalent in the kernel, so if I think of a better name, I'll let you know, but otherwise this is fine.

tests/linux_kernel/helpers/test_runq.py
richl9: Yes, this makes more sense.

richl9:
Thanks @osandov for the comments! I have addressed them.

drgn/helpers/linux/sched.py
```diff
@@ -163,3 +164,44 @@ def loadavg(prog: Program) -> Tuple[float, float, float]:
     avenrun = prog["avenrun"]
     vals = [avenrun[i].value_() / (1 << 11) for i in range(3)]
     return (vals[0], vals[1], vals[2])
+
+
+def cpu_rq(prog: Program, cpu: int) -> Object:
```
osandov, line 169: This should use `drgn.helpers.common.prog.takes_program_or_default` and `drgn.IntegerLike` to make the `prog` argument optional:

```suggestion
@takes_program_or_default
def cpu_rq(prog: Program, cpu: IntegerLike) -> Object:
```

drgn/helpers/linux/sched.py
```diff
@@ -163,3 +164,44 @@ def loadavg(prog: Program) -> Tuple[float, float, float]:
...
+    on cpu will be short, so this can roughly tell how long this task
+    has been staying in current status.
+    For task status in "R" status, if it's still on cpu, then this return
+    the duration time this task has been running, otherwise it roughly tell
+    how long this task has been staying in runqueue.
+
+    :param task: ``struct task_struct *``
+    :returns: duration in ns granularity
+    """
+    arrival_time = task.sched_info.last_arrival.value_()
```
osandov, line 204: Running the tests revealed that this only exists if `CONFIG_SCHED_INFO=y` (selected by `CONFIG_SCHEDSTATS` or `CONFIG_TASK_DELAY_ACCT`). I'm rebuilding our test kernels now to enable this option: https://github.com/osandov/drgn/commit/05ed1c947969c73bf3610ed37190775fa0ed683e, https://github.com/osandov/drgn/actions/runs/16479745871. But please mention this requirement in the docstring.

osandov (changes requested):
A final couple of minor updates. We'll need to wait for the vmtest build in https://github.com/osandov/drgn/actions/runs/16479745871 to finish so we can actually run the tests.

osandov (commented):
Code looks good now, thanks. After thinking about the naming for a bit longer, how about `task_since_last_arrival_ns()`? That reflects the actual timestamp and unit being used more accurately, although it's a bit verbose, so I'm happy to consider alternatives.

I'd also streamline the comment a bit, something like:

```diff
diff --git a/drgn/helpers/linux/sched.py b/drgn/helpers/linux/sched.py
index 748065e8..e082f07d 100644
--- a/drgn/helpers/linux/sched.py
+++ b/drgn/helpers/linux/sched.py
@@ -187,22 +187,18 @@ def task_rq(task: Object) -> Object:
     return cpu_rq(task.prog_, task_cpu(task))
 
 
-def task_lastrun2now(task: Object) -> int:
+def task_since_last_arrival_ns(task: Object) -> int:
     """
     Get the number of nanoseconds since a task last started running.
 
-    The return duration will cover task's last run time on cpu and also
-    the time staying in current status, usually the time slice for task
-    on cpu will be short, so this can roughly tell how long this task
-    has been staying in current status.
-    For task status in "R" status, if it's still on cpu, then this return
-    the duration time this task has been running, otherwise it roughly tell
-    how long this task has been staying in runqueue.
+    Assuming that time slices are short, this is approximately the time that
+    the task has been in its current status (running, queued, or blocked).
 
-    Note that task.sched_info only exists if CONFIG_SCHED_INFO=y, enabled by CONFIG_SCHEDSTATS or CONFIG_TASK_DELAY_ACCT.
+    This is only supported if the kernel was compiled with
+    ``CONFIG_SCHEDSTATS`` or ``CONFIG_TASK_DELAY_ACCT``.
 
     :param task: ``struct task_struct *``
-    :returns: duration in ns granularity
+    :returns: Duration in nanoseconds.
     """
     arrival_time = task.sched_info.last_arrival.value_()
     rq_clock = task_rq(task).clock.value_()
diff --git a/tests/linux_kernel/helpers/test_sched.py b/tests/linux_kernel/helpers/test_sched.py
index 5d56ceca..09241e28 100644
--- a/tests/linux_kernel/helpers/test_sched.py
+++ b/tests/linux_kernel/helpers/test_sched.py
@@ -13,8 +13,8 @@ from drgn.helpers.linux.sched import (
     idle_task,
     loadavg,
     task_cpu,
-    task_lastrun2now,
     task_on_cpu,
+    task_since_last_arrival_ns,
     task_state_to_char,
     task_thread_info,
 )
@@ -104,8 +104,8 @@ class TestSched(LinuxKernelTestCase):
         self.assertEqual(len(values), 3)
         self.assertTrue(all(v >= 0.0 for v in values))
 
-    def test_task_lastrun2now(self):
+    def test_task_since_last_arrival_ns(self):
         with fork_and_stop() as pid:
             time.sleep(0.01)
             task = find_task(self.prog, pid)
-            self.assertGreaterEqual(task_lastrun2now(task), 10000000)
+            self.assertGreaterEqual(task_since_last_arrival_ns(task), 10000000)
```

If you're okay with that, you can either fold it into your PR and force push or I can apply it after merging, just let me know.

richl9:
Hi @osandov thanks for the comments. Yes, I agree that the new name will be more accurate. Feel free to apply the changes after merging.

PR #523, opened 2025-07-16
format thread_info value as hex in sys.py

qiyeliu:
Make the thread_info address output consistent with the task address output by formatting it as hexadecimal (using :x) in the string representation.
```

TASK: ffff8a7c9c5cb500  [THREAD_INFO: 18446614866536740096] 
->
TASK: ffff8a7cf10e3500  [THREAD_INFO: ffff8a7cf10e3500]
```


osandov:
Good catch, thanks!

Issue #527, opened 2025-07-17
%crash warning couldn't get DATE/UPTIME: 'tk_core'


qiyeliu:
Directly run the crash command of drgn, and there are two warning.

```
[root@localhost drgn]# drgn 
drgn 0.0.32+unknown (using Python 3.11.6, elfutils 0.190, without debuginfod (dlopen), without libkdumpfile,)
For help, type help(drgn).
>>> import drgn
>>> from drgn import FaultError, NULL, Object, alignof, cast, container_of, execscript, implicit_convert, ofe
>>> from drgn.helpers.common import *
>>> from drgn.helpers.linux import *
>>> 
>>> %crash
warning: couldn't get DATE: 'tk_core'
warning: couldn't get UPTIME: 'tk_core'
      KERNEL: /root/liuye/linux-next-cscope/vmlinux
    DUMPFILE: /proc/kcore
        CPUS: 8
LOAD AVERAGE: 0.79, 0.26, 0.08
       TASKS: 390
    NODENAME: localhost.localdomain
     RELEASE: 6.16.0-rc4-next-20250701-00002-g7d3b6b00ca67-dirty
     VERSION: #1 SMP PREEMPT_DYNAMIC Tue Jul  8 15:00:14 CST 2025
     MACHINE: x86_64
      MEMORY: 14.5G
         PID: 14777
     COMMAND: "drgn"
        TASK: ffff8d11b03c3500  [THREAD_INFO: ffff8d11b03c3500]
         CPU: 7
       STATE: R (running) (ACTIVE)
%crash> 

```

osandov:
Fixed by 133d40a7da4532312418006a36450ce9873d0d83.

Issue #28, opened 2019-12-04
Use drgn-specific exceptions


osandov:
@prakashsurya pointed out that drgn throwing generic Python exceptions makes it hard to differentiate between a Python programming error and drgn-specific errors, like out of date helper code.We might want to add drgn-specific exceptions.

One concern with, e.g., changing `ValueError` to `drgn.ValueError` is that existing scripts out there may already be depending on the existing exceptions. One way to handle this would be to make the drgn-specific exceptions inherit from the generic exception; existing code will continue working, and newer code can catch the more specific exception.

prakashsurya:
Thanks for opening the bug!

Here's a concrete example from sdb, where we thought the lack of a drgn-specific exception might hide non-drgn `ValueError` execeptions; [link](https://github.com/delphix/sdb/blob/master/sdb/commands/stacks.py#L230):
```
            try:
                for frame in sdb.get_prog().stack_trace(task):
                    stack_key.append(frame.pc)
            except ValueError:
                #
                # Unwinding the stack of a running/runnable task will
                # result in an exception. Since we expect some tasks to
                # be running, we silently ignore this case, and move on.
                #
                # Unfortunately, the exception thrown in this case is a
                # generic "ValueError" exception, so we may wind up
                # masking other "ValueError" exceptions that are not due
                # to unwinding the stack of a running task.
                #
                # We can't check the state of the task here, and verify
                # it's in the "R" state, since that state can change in
                # between the point where the "ValueError" exception was
                # originally raised, and here where we'd verify the
                # state of the task; i.e. it could have concurrently
                # transitioned from running to some other state.
                #
                pass
```

PR #528, opened 2025-07-17
python: raise drgn.ObjectNotFoundError instead of KeyError/LookupError

osandov:
Add a new `drgn.ObjectNotFoundError` exception type to use for failed `prog[]`, `prog.object()`, `prog.variable()`, `prog.constant()`, and `prog.function()` lookups. This has a few benefits:

1. It allows catching object lookup errors without the risk of catching a stray error from a dict or attribute lookup. See #28.
2. It allows us to have a more descriptive error message. See #527.
3. It makes it possible to get the name that could not be found from the exception.

E.g., instead of:

```pycon
>>> prog['asdf']
  File "<python-input-0>", line 1, in <module>
    prog['asdf']
    ~~~~^^^^^^^^
KeyError: 'asdf'
```

We get:

```pycon
>>> prog['asdf']
  File "<python-input-0>", line 1, in <module>
    prog['asdf']
    ~~~~^^^^^^^^
_drgn.ObjectNotFoundError: could not find 'asdf'
```

It is a subclass of `KeyError` (and thus also of `LookupError`), so this should be backwards compatible. The implementation is inspired by `AttributeError` in CPython.

If we end up liking this, we might want to do the same for object member lookups.


osandov:
@brenns10 I _think_ this is safe, but I wanted to double check if you can think of any issues this would cause, especially in drgn-tools.

brenns10:
I can't come up with any reason why this would cause an issue, given that it's a subclass of `KeyError`. I'd say go for it, but thanks for checking!

PR #533, opened 2025-08-06
drgn: add runq command

richl9:
Pushed a new branch for the runqueue helper with the earlier feedback applied. Just wanted to get a quick review to see if any structural changes are needed before I continue:)  One thing is that maybe we want to move parse_range_list to somewhere else?

### DIFF
--- a/drgn/commands/_builtin/crash/_sched.py
+++ b/drgn/commands/_builtin/crash/_sched.py
@@ -1,4 +1,5 @@
 # Copyright (c) Meta Platforms, Inc. and affiliates.
+# Copyright (c) 2026, Oracle and/or its affiliates.
 # SPDX-License-Identifier: LGPL-2.1-or-later
 
 """Scheduler-related crash commands."""
@@ -6,22 +7,32 @@
 import argparse
 import sys
 import textwrap
-from typing import Any, List, Sequence
+from typing import Any, Iterable, List, Sequence
 
 from drgn import Object, Program, offsetof
 from drgn.commands import _repr_black, argument, drgn_argument, mutually_exclusive_group
 from drgn.commands.crash import (
+    Cpuspec,
     CrashDrgnCodeBuilder,
     _crash_foreach_subcommand,
     _guess_type,
     _parse_type_name_and_member,
     _prefer_object_lookup,
+    _print_task_header,
     _TaskSelector,
     crash_command,
+    parse_cpuspec,
     print_task_header,
 )
-from drgn.helpers.common.format import CellFormat, print_table
+from drgn.helpers.common.format import CellFormat, escape_ascii_string, print_table
 from drgn.helpers.linux.pid import for_each_task_in_group
+from drgn.helpers.linux.sched import (
+    cfs_rq_for_each_entity,
+    cpu_rq,
+    rq_for_each_rt_task,
+    sched_entity_to_task,
+    task_group_name,
+)
 from drgn.helpers.linux.signal import (
     decode_sigaction_flags_value,
     decode_sigset,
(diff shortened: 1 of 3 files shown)


drgn/helpers/linux/runqueue.py
```diff
@@ -0,0 +1,173 @@
...
+                "prio": t.prio.value_(),
+                "pid": t.pid.value_(),
+                "task_addr": int(t),
+                "command": escape_ascii_string(t.comm.string_()),
+            }
+        )
+    return {"cfs_root": cfs_root, "tasks": results}
+
+
+def run_queue(prog: Program, args: argparse.Namespace) -> List[Dict[str, Any]]:
```
osandov, line 102: This doesn't make much sense as a helper. It's basically the implementation of `_crash_cmd_runq` just modified to spit out a dictionary.

It's expected that a complex command like `runq` will have complex logic for iterating over whatever is needed and printing it, so that part can go back inside of `_crash_cmd_runq`. The point of @brenns10's [comment](https://github.com/osandov/drgn/pull/513#discussion_r2178257742) on the previous PR was that any non-trivial inspection of the internals of a subsystem should go in small, self-contained helper functions. For example, it's a great idea to add helpers for iterating over tasks in a runqueue, and it's totally fine to call those directly from the command function.

tests/linux_kernel/__init__.py
```diff
@@ -213,7 +213,7 @@ def smp_enabled():
     return bool(re.search(r"\bSMP\b", os.uname().version))
 
 
-def parse_range_list(s):
+def parse_range_list(s) -> Set[int]:
```
osandov, line 216: You can drop this since it won't be needed anymore once you use `parse_cpuspec()` instead.

osandov (commented):
Thanks, I tried to clear up what @brenns10 and I had in mind and had a few more comments here and there. Let me know if that still doesn't make sense.

Also, I think you meant to say "command" instead of "helper" in your commit and PR titles.

drgn/helpers/linux/runqueue.py
```diff
@@ -0,0 +1,173 @@
...
+def _get_runqueue_timestamps(runqueue: Object) -> int:
+    """
+    Get runqueue clock timestamp.
+
+    :param runque: Object
+    :returns: rq timestamp
+    """
+    # Try common fields in order; not all will exist on all kernels
+    rq_ts = 0
+    for name in ("clock", "most_recent_timestamp", "timestamp_last_tick"):
```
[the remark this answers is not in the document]
richl9: Just cleaned them.

richl9:
HI @osandov , thanks for the comments. Anymore big change suggestions before I wrap up the details?

drgn/helpers/linux/runqueue.py
richl9: That makes sense. I kept get_rq_per_cpu helper though just to get per cpu runqueue since I think it may be qualified as a "kernel internal inspection". Let me know what do you think.

drgn/helpers/linux/runqueue.py
```diff
@@ -0,0 +1,68 @@
...
+
+    :param runqueue: Object
+    :return: Iterator of (``struct task_struct``, int) tuples
+    """
+    runq = runqueue.address_of_()
+    for t in list_for_each_entry(
+        "struct task_struct", runq.cfs_tasks.address_of_(), "se.group_node"
+    ):
+        if t == runqueue.curr:
+            continue
```
osandov, line 47: Why is this check necessary? It seems more appropriate for the caller to ignore the current task if that's what it wants. If you get rid of the check, then this whole function body can be replaced by:

```python3
    return list_for_each_entry(
        "struct task_struct", runqueue.cfs_tasks.address_of_(), "se.group_node"
    )
```

drgn/helpers/linux/runqueue.py
```diff
@@ -0,0 +1,68 @@
...
+        yield container_of(t, "struct task_struct", "rt")
+
+
+def get_rq_per_cpu(prog: Program, cpus: List[int] = []) -> Iterator[Tuple[int, Object]]:
```
osandov, line 51: I still think this one should go in `drgn/commands/_builtin/crash/runq.py` instead of as a public helper.

drgn/helpers/linux/runqueue.py
```diff
@@ -0,0 +1,68 @@
...
+    :param runqueue: Object
+    :return: Iterator of (``struct task_struct``, int) tuples
+    """
+    runq = runqueue.address_of_()
+    for t in list_for_each_entry(
+        "struct task_struct", runq.cfs_tasks.address_of_(), "se.group_node"
+    ):
+        if t == runqueue.curr:
+            continue
+        yield container_of(t, "struct task_struct", "rt")
```
osandov, line 48: This `container_of()` is for real-time, which doesn't seem right for this function.

osandov (commented):
Thanks, this structure is much better. I commented a couple more tweaks for the structure and some more comments on the helpers themselves.

drgn/commands/_builtin/crash/_runq.py
```diff
@@ -0,0 +1,192 @@
...
+@crash_command(
+    description="Display the tasks on the run queues of each cpu.",
+    arguments=(
+        mutually_exclusive_group(
+            argument("-t", action="store_true", dest="show_timestamps"),
+            argument("-T", action="store_true", dest="show_lag"),
+            argument("-m", action="store_true", dest="pretty_runtime"),
+            argument("-g", action="store_true", dest="group"),
+        ),
+        argument("-c", type=str, default="a", dest="cpus"),
```
osandov: These all need help strings.

drgn/commands/_builtin/crash/_runq.py
```diff
@@ -0,0 +1,192 @@
...
+    prog: Program, name: str, args: argparse.Namespace, **kwargs: Any
+) -> None:
+    table_format = args.show_timestamps or args.show_lag or args.pretty_runtime
+    table: List[List[Any]] = []
+    headers: List[Any] = []
+
+    runq_clocks: Dict[int, int] = {}
+    cpus = parse_cpuspec(args.cpus).cpus(prog)
+    for i, (cpu, runqueue) in enumerate(get_rq_per_cpu(prog, cpus)):
+        curr_task = runqueue.curr[0].address_of_()
```
osandov: What is the `[0].address_of_()` for? Instead, I think we want to read it:

```suggestion
        curr_task = runqueue.curr.read_()
```

drgn/commands/_builtin/crash/_runq.py
```diff
@@ -0,0 +1,192 @@
...
+            print(f"  CFS RB_ROOT: {hex(cfs_root)}")
+            is_cfs_queue = False
+            if cfs_tasks:
+                for task in cfs_tasks:
+                    if task == runqueue.curr:
+                        continue
+                    is_cfs_queue = True
+                    print(
+                        " " * 4,
+                        f"[{task.prio.value_():3d}] PID: {task.pid.value_():<6d} TASK: {hex(int(task))}  "
```
osandov:
```suggestion
                        f"[{task.prio.value_():3d}] PID: {task.pid.value_():<6d}  TASK: {hex(int(task))}  "
```

drgn/commands/_builtin/crash/_runq.py
```diff
@@ -0,0 +1,192 @@
...
+                }
+                sorted_lags = sorted(lags.items(), key=lambda item: item[1])
+                for c, lag in sorted_lags:
+                    print(f"CPU {c}: {lag / 1e9:.2f} secs")
+                return
+            else:
+                continue
+
+        if table_format:
+            row = [
```
osandov: The format of these options is fairly different from the crash output, was that an intentional choice. That's okay if there's an intentional reason, but it'd be nice to imitate crash's format if not.

drgn/helpers/linux/runqueue.py
```diff
@@ -0,0 +1,68 @@
...
+from drgn.helpers.linux.list import list_for_each_entry
+from drgn.helpers.linux.percpu import per_cpu
+
+
+def get_rt_runq_tasks(runqueue: Object) -> Iterator[Object]:
+    """
+    Get RT runqueue tasks in rt scheduler.
+
+    :param runqueue: Object
+    :return: Iterator of ``struct task_struct``
```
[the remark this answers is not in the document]
osandov: This still needs to be updated to "Iterator of ``struct task_struct *`` objects."

osandov (changes requested):
Okay finally got back to this. In addition to the inline comments, a few high-level comments:

1. I don't think there will be enough runqueue-specific stuff to merit its own module, so let's move the new helpers to `drgn.helpers.linux.sched`.
2. The `--drgn` option is the main motivation for porting these, so please add that.
3. This needs some basic test cases. See `tests/linux_kernel/crash_commands` for examples.

richl9:
> Hey, thanks for the update! I was actually just in the middle of implementing the `-g` option, so I will combine our work when I'm done.

Thank you Omar.

osandov:
I ended up doing a big refactor in the process of adding `-g`. Thanks for taking it most of the way!

PR #534, opened 2025-08-08
crash: add ptov command

qiyeliu:
Add ptov command for drgn.

### DIFF
--- a/drgn/commands/_builtin/crash/ptov.py
+++ b/drgn/commands/_builtin/crash/ptov.py
@@ -1,0 +1,114 @@
+# Copyright (c) 2025, Kylin Software, Inc. and affiliates.
+# SPDX-License-Identifier: LGPL-2.1-or-later
+
+import argparse
+from typing import Any
+
+from drgn import Object, Program
+from drgn.commands import argument, drgn_argument
+from drgn.commands.crash import CrashDrgnCodeBuilder, crash_command, parse_cpuspec
+from drgn.helpers.common.format import print_table
+from drgn.helpers.linux.mm import phys_to_virt
+from drgn.helpers.linux.percpu import per_cpu_ptr
+
+
+@crash_command(
+    description="physical or per-CPU to virtual",
+    long_description="""This command translates a hexadecimal physical address into a
+                        kernel virtual address. Alternatively, a hexadecimal per-cpu
+                        offset and cpu specifier will be translated into kernel virtual
+                        addresses for each cpu specified.""",
+    arguments=(
+        argument(
+            "address",
+            metavar="address|offset:cpuspec",
+            nargs="+",
+            help="hexadecimal physical address or hexadecimal per-CPU offset and CPU specifier",
+        ),
+        drgn_argument,
+    ),
+)
+def _crash_cmd_ptov(
+    prog: Program, name: str, args: argparse.Namespace, **kwargs: Any
+) -> None:
+    if args.drgn:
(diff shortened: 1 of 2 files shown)


drgn/commands/_builtin/crash/ptov.py
```diff
@@ -0,0 +1,79 @@
...
+        # Translate per-CPU offsets
+        print("PER-CPU OFFSET: {:x}".format(offset))
+        print("  CPU    VIRTUAL")
+        ptr = Object(prog, "unsigned long", offset)
+        for cpu in cpus.cpus(prog):
+            virt = per_cpu_ptr(ptr, cpu)
+            print("  [{:d}]  {:016x}".format(cpu, virt.value_()))
+    else:
+        # Handle physical address case
+        address_str = args.address[0]
```
osandov, line 74: The crash command (and your `argument()` specification above) support multiple addresses, but this only handles the first.

drgn/commands/_builtin/crash/ptov.py
```diff
@@ -0,0 +1,79 @@
...
+        phys = int(address_str, 16)
+        virt = phys_to_virt(phys)
+        virt_int = virt.value_()
+        print("VIRTUAL           PHYSICAL")
+        print("{:016x}  {:x}".format(virt_int, phys), flush=True)
```
osandov, line 79: If you use `drgn.helpers.common.format.print_table()`, then you don't need to manually align things.

drgn/commands/_builtin/crash/ptov.py
```diff
@@ -0,0 +1,79 @@
...
+    offset_str, cpu_spec = address.split(":", 1)
+    offset = int(offset_str, 16)
+    cpus = parse_cpuspec(cpu_spec)
+    print("PER-CPU OFFSET: {:x}".format(offset))
+    print("  CPU    VIRTUAL")
+    ptr = Object(prog, "unsigned long", offset)
+    for cpu in cpus.cpus(prog):
+        virt = per_cpu_ptr(ptr, cpu)
+        print("  [{:d}]  {:016x}".format(cpu, virt.value_()))
+                  """
```
osandov, line 48: Don't worry about having the generated code print the actual table. It should just do whatever it takes to get the point across for what helpers to use, etc. So for this, calling `per_cpu_ptr()` on the address is good enough.

drgn/commands/_builtin/crash/ptov.py
```diff
@@ -0,0 +1,79 @@
...
+) -> None:
+    if args.drgn:
+        if ":" in args.address[0]:
+            print("from drgn.commands.crash import parse_cpuspec")
+            print(f'address = ["{args.address[0]}"]')
+            print(
+                """
+    offset_str, cpu_spec = address.split(":", 1)
+    offset = int(offset_str, 16)
+    cpus = parse_cpuspec(cpu_spec)
```
osandov, line 41: Rather than having the generated code call `parse_cpuspec()`, parse the cpuspec in the actual command code and use `drgn.commands.crash.CrashDrgnCodeBuilder.append_cpuspec()`.

osandov (changes requested):
Thanks so much for working on this!  I left a few comments. Once the details are worked out, this will also need a test case.

qiyeliu:
@osandov 
I have updated the code according to the review suggestions. Please take the time to review it and let me know if there are any changes that need to be made. Thank you.

PR #536, opened 2025-08-14
crash: add files command

YassineLr:


YassineLr:
Here is a first version. If you have any recommendations or suggestions for changes, please let me know as I continue working on its compatibility with the files crash output.

PR #542, opened 2025-09-16
libdrgn: linux kernel: restore THREAD_SIZE object finder

brenns10:
It looks like the linker symbols, in addition to `union thread_union`, are enough to resurrect the `THREAD_SIZE` object for all architectures! I updated the docs and included a test that uses the kmod to properly get the value and verify it. I got a clean vmtest on all architectures, at least for this helper. 

I did encounter some errors in aarch64 tiny/alternative for kernels <= 5.14, on `TestFsRefs.test_btrfs_subvolume` and `TestFsRefs::test_super_block_on_block_device`. I'll share some logs on that in the morning in a Github issue.

### DIFF
--- a/libdrgn/linux_kernel.c
+++ b/libdrgn/linux_kernel.c
@@ -34,6 +34,7 @@
 #include "log.h"
 #include "platform.h"
 #include "program.h"
+#include "symbol.h"
 #include "type.h"
 #include "util.h"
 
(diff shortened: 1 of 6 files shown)


libdrgn/linux_kernel.c
```diff
@@ -195,6 +197,92 @@ static struct drgn_error *linux_kernel_get_page_mask(struct drgn_program *prog,
...
+	} else if (!drgn_error_catch(&err, DRGN_ERROR_LOOKUP)) {
+		return err;
+	}
+
+#define SYMBOL_START_END(symname_start, symname_end) do { \
+		struct drgn_symbol _cleanup_symbol_ *sym_start = NULL; \
+		struct drgn_symbol _cleanup_symbol_ *sym_end = NULL; \
+		err = drgn_program_find_symbol_by_name(prog, symname_start, &sym_start); \
+		if (drgn_error_catch(&err, DRGN_ERROR_LOOKUP)) \
+			break; \
```
osandov, line 260: This `break` is deranged, but I like it :stuck_out_tongue:

libdrgn/linux_kernel.c
```diff
@@ -195,6 +197,92 @@ static struct drgn_error *linux_kernel_get_page_mask(struct drgn_program *prog,
...
+	struct drgn_qualified_type qualified_type, thread_union;
+	qualified_type.qualifiers = 0;
+	err = drgn_program_find_primitive_type(prog, DRGN_C_TYPE_UNSIGNED_LONG,
+					       &qualified_type.type);
+	if (err)
+		return err;
+
+	if (prog->thread_size_cached)
+		return drgn_object_set_unsigned(ret, qualified_type, \
+						prog->thread_size_cached, 0); \
```
osandov, line 213: Unnecessary backslashes here.

osandov (changes requested):
I'm so happy to have this back :smile: Just one nitpick along with the one you noticed.

osandov:
Oh, let's drop my old smoke test, too, the kmod-based one is much better.

brenns10:
I would _love_ an equivalent to the `-Wunused-*` compiler flags, but for mypy or another static checker, if it could be done reliably :)

osandov (approved):

PR #543, opened 2025-09-16
linux/mm: make decode_page_flags() handle memdesc_flags_t

qiyeliu:
Since Linux kernel 6.18 (linux-next，“mm: introduce memdesc_flags_t”)
    struct page->flags is no longer an unsigned long, but a memdesc_flags_t 
    containing the actual flags in '.f'. This helper now handles both the legacy 
    unsigned long flags and the new memdesc_flags_t.

    Add get_page_flags() in drgn helpers to automatically handle both
    cases by trying page.flags.f first and falling back to page.flags.

    This fixes TypeError in drgn 0.0.32+ when running on newer kernels.

### DIFF
--- a/drgn/helpers/linux/mm.py
+++ b/drgn/helpers/linux/mm.py
@@ -121,7 +121,25 @@ __all__ = (
     "PageWriteback",
     "PageXenRemapped",
     "PageYoung",
+    "get_page_flags",
 )
+
+
+def get_page_flags(page: Object) -> int:
+    """
+    Return the integer representation of a page's flags.
+    Handles both legacy unsigned long flags and memdesc_flags_t (flags.f).
+
+    :param page: ``struct page *``
+    """
+
+    # Since Linux kernel 6.18 (linux-next，"mm: introduce memdesc_flags_t")
+    # struct page->flags is no longer an unsigned long, but a memdesc_flags_t
+    # containing the actual flags in '.f'.
+    try:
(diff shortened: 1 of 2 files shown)


drgn/helpers/linux/mm.py
```diff
@@ -124,6 +124,21 @@
 )
 
 
+def _get_page_flags(page: Object) -> int:
```
osandov, line 127: This should probably be a public helper, so drop the leading underscore and add it to `__all__`.

drgn/helpers/linux/mm.py
```diff
@@ -124,6 +124,21 @@
...
+    """
+    Return the integer representation of a page's flags.
+    Handles both legacy unsigned long flags and memdesc_flags_t (flags.f).
+
+    # Since Linux kernel commit 53fbef56e07df ("mm: introduce memdesc_flags_t")
```
osandov, line 132: This comment should be outside of the docstring.

osandov (changes requested):
Thanks! I hadn't hit this in my tests yet. I assume this is a change in linux-next that isn't in mainline yet? If so, the commit hashes may not be finalized yet, so we may want to update the commit references to remove the hash indicate that it is in linux-next for 6.18.

Also, the DCO check is complaining because your commit author email is different from the signed-off-by email.

osandov (approved):

PR #547, opened 2025-09-17
Aarch64 exception frame unwinding

brenns10:
This is a hacky draft to demonstrate the idea of what we can do to unwind through aarch64 interrupt/exception frames.

There were two issues:

1. Near the bottom of the exception stack, a CFI row is returned, and drgn can only get one register (ra_sign-state) from it. So the register state is essentially empty. But because the CFI row was found and used, the stack frame gets appended, and we never try the fallback. My hack around this for now is to require a minimum of 2 registers present in the register state, and when the minimum is not met, return `&drgn_not_found` so we try the fallback. Obviously this needs to be refined.
2. This gets us to the bottom of the exception stack. If the resulting frame pointer points at two NULL slots on the stack pointer, then there may be a metadata field which we can read to detect that this is actually a pt_regs, and we can determine the pt_regs from that. The approach would be different on older kernels.

To help demonstrate I included patches to panic in an IRQ on aarch64. That way you can test it easily with:

```
python -m vmtest.vm --build-drgn --insert-test-kmod -k aarch64:6.16.*default bash
...
$ python3 -m vmtest.enter_kdump
...
$ python3 -m drgn -c /proc/vmcore -s $DRGN_TEST_KMOD -e 'print(prog.crashed_thread().stack_trace())'
```

The result without the WIP patch at the end:
```
#0  crash_setup_regs (./arch/arm64/include/asm/kexec.h:45:3)
#1  __crash_kexec (./kernel/crash_core.c:119:4)
#2  panic (./kernel/panic.c:401:3)
#3  drgn_test_crash_func (./drgn_test/drgn_test.c:1652:2)
#4  drgn_test_crash_irq_work_fn (./drgn_test/drgn_test.c:1677:2)
#5  irq_work_single (./kernel/irq_work.c:221:2)
#6  irq_work_run_list (./kernel/irq_work.c:252:3)
#7  irq_work_run (./kernel/irq_work.c:261:2)
#8  do_handle_IPI (./arch/arm64/kernel/smp.c:989:3)
#9  ipi_handler (./arch/arm64/kernel/smp.c:1016:2)
#10 handle_percpu_devid_irq (./kernel/irq/chip.c:857:9)
#11 generic_handle_irq_desc (./include/linux/irqdesc.h:173:2)
#12 handle_irq_desc (./kernel/irq/irqdesc.c:676:2)
#13 generic_handle_domain_irq (./kernel/irq/irqdesc.c:732:9)
#14 gic_handle_irq (./drivers/irqchip/irq-gic.c:370:3)
#15 call_on_irq_stack (./arch/arm64/kernel/entry.S:895)
#16 do_interrupt_handler (./arch/arm64/kernel/entry-common.c:310:3)
#17 ???
```
And the result with it:
```
#0  crash_setup_regs (./arch/arm64/include/asm/kexec.h:45:3)
#1  __crash_kexec (./kernel/crash_core.c:119:4)
#2  panic (./kernel/panic.c:401:3)
#3  drgn_test_crash_func (./drgn_test/drgn_test.c:1652:2)
#4  drgn_test_crash_irq_work_fn (./drgn_test/drgn_test.c:1677:2)
#5  irq_work_single (./kernel/irq_work.c:221:2)
#6  irq_work_run_list (./kernel/irq_work.c:252:3)
#7  irq_work_run (./kernel/irq_work.c:261:2)
#8  do_handle_IPI (./arch/arm64/kernel/smp.c:989:3)
#9  ipi_handler (./arch/arm64/kernel/smp.c:1016:2)
#10 handle_percpu_devid_irq (./kernel/irq/chip.c:857:9)
#11 generic_handle_irq_desc (./include/linux/irqdesc.h:173:2)
#12 handle_irq_desc (./kernel/irq/irqdesc.c:676:2)
#13 generic_handle_domain_irq (./kernel/irq/irqdesc.c:732:9)
#14 gic_handle_irq (./drivers/irqchip/irq-gic.c:370:3)
#15 call_on_irq_stack (./arch/arm64/kernel/entry.S:895)
#16 do_interrupt_handler (./arch/arm64/kernel/entry-common.c:310:3)
#17 __el1_irq (./arch/arm64/kernel/entry-common.c:583:2)
#18 el1_interrupt (./arch/arm64/kernel/entry-common.c:598:3)
#19 el1h_64_irq_handler (./arch/arm64/kernel/entry-common.c:603:2)
#20 el1h_64_irq+0x6c/0x70 (./arch/arm64/kernel/entry.S:596)
#21 gic_ipi_send_mask (./drivers/irqchip/irq-gic.c:847:2)
#22 __ipi_send_mask (./kernel/irq/ipi.c:285:3)
#23 smp_cross_call (./arch/arm64/kernel/smp.c:1023:2)
#24 arch_irq_work_raise (./arch/arm64/kernel/smp.c:868:2)
#25 irq_work_raise (./kernel/irq_work.c:84:2)
#26 __irq_work_queue_local (./kernel/irq_work.c:112:3)
#27 __irq_work_queue_local (./kernel/irq_work.c:88:13)
#28 irq_work_queue (./kernel/irq_work.c:124:2)
#29 irq_work_queue (./kernel/irq_work.c:116:6)
#30 drgn_test_crash_store (./drgn_test/drgn_test.c:1699:2)
#31 kobj_attr_store (./lib/kobject.c:840:9)
#32 sysfs_kf_write (./fs/sysfs/file.c:145:9)
#33 kernfs_fop_write_iter (./fs/kernfs/file.c:334:9)
#34 new_sync_write (./fs/read_write.c:593:8)
#35 vfs_write (./fs/read_write.c:686:9)
#36 ksys_write (./fs/read_write.c:738:9)
#37 __do_sys_write (./fs/read_write.c:749:9)
#38 __se_sys_write (./fs/read_write.c:746:1)
#39 __arm64_sys_write (./fs/read_write.c:746:1)
#40 __invoke_syscall (./arch/arm64/kernel/syscall.c:35:9)
#41 invoke_syscall (./arch/arm64/kernel/syscall.c:49:9)
#42 el0_svc_common (./arch/arm64/kernel/syscall.c:132:2)
#43 do_el0_svc (./arch/arm64/kernel/syscall.c:151:2)
#44 el0_svc (./arch/arm64/kernel/entry-common.c:767:2)
#45 el0t_64_sync_handler (./arch/arm64/kernel/entry-common.c:786:3)
#46 el0t_64_sync+0x198/0x19c (./arch/arm64/kernel/entry.S:600)
```

### DIFF
--- a/tests/linux_kernel/kmod/drgn_test.c
+++ b/tests/linux_kernel/kmod/drgn_test.c
@@ -1770,6 +1770,8 @@ DEFINE_KMODIFY_TEST_ARGS(
 // unwinding through them.
 #ifdef __x86_64__
 #define DRGN_TEST_NMI_CRASH
+#elif defined(__aarch64__)
+#define DRGN_TEST_IRQ_CRASH
 #endif
 
 


osandov:
Like you already mentioned, the "two registers" heuristic probably isn't the right one, but I think this is on the right track. So just to double check my understanding, there is actually a CFI row in the DWARF, but it only includes the `ra_sign_state`? Or the CFI row includes some additional registers but only `ra_sign_state` succeeds? A couple of options that come to mind are:

* Instead of checking for any register, checking specifically whether we got the return address register (`ret_addr_regno`). The tricky part with that is I think in theory a CFI row that intentionally omits the return address register is supposed to indicate the end of the stack, although I don't know whether that is true in practice. So maybe the check should be "if the CFI row included a rule for the return address register, but we couldn't unwind it, fall back".
* Architecture-specific logic that specifically disregards `ra_sign_state`.

I'm leaning towards the former, but it depends on what exactly is in the DWARF.

Re: "tests: linux_kernel: panic within IRQ on aarch64", you're probably going to run into https://lore.kernel.org/all/ec2d02358eee53f095ad9a7d9c96af8f4d0d4eeb.1741636880.git.osandov@osandov.com/, where the IRQ stack in `/proc/vmcore` is bogus. I'll revive that patch with the changes the maintainers requested.

brenns10:
> So just to double check my understanding, there is actually a CFI row in the DWARF, but it only includes the `ra_sign_state`? Or the CFI row includes some additional registers but only `ra_sign_state` succeeds?

According to my logging, it's the latter.  There is a CFI row which includes rules for 14 registers with types other than `DRGN_CFI_RULE_UNDEFINED`. Yet it only succeeds in getting the value for `ra_sign_state`. So it sounds like your preferred approach should work. I'll give that a shot.

I haven't yet seen the IRQ stack issue you mentioned but I can see how it would be an issue, thanks for the heads up.

osandov:
Patch sent: https://lore.kernel.org/linux-arm-kernel/93a6f8e47cce118493e250b719a927f517be12cf.1758316750.git.osandov@osandov.com/.

brenns10:
(It's passing everywhere except aarch64 -tiny, due to the vmap issue you mentioned).

brenns10:
Ok, I have added your patches (and backports) and bumped the vmtest version for aarch64. I built all the aarch64 tiny variants (which were the only ones that failed previously), and they all pass! So I think this is ready for review. If you want to look at the vmtest patches first, before getting into the IRQ unwind heuristics discussed above, I can break things out in whichever way you prefer.

Once all this is done, I believe I'll be ready to submit the branch for the bt command. In terms of unwinding through interrupts, categorizing the stacks, and printing registers at exception boundaries, drgn will be at parity with crash for aarch64 and x86_64. The bt command won't have all the flags yet, and there will still be some room for improvement on ppc64, s390x, and arm, but it will be a great place to begin.

libdrgn/arch_aarch64.c
```diff
@@ -68,6 +68,100 @@ static void demangle_cfi_registers_aarch64(struct drgn_program *prog,
...
+
+	err = get_stackframe_offset(prog, &stackframe_offset);
+	if (err)
+		return err;
+
+	uint64_t pt_regs_addr = fp - stackframe_offset;
+
+	if (!frame[0] && !frame[1]) {
+		// Since c2c6b27b5aa14 ("arm64: stacktrace: unwind exception
+		// boundaries") - v6.13, a signal frame with both values zero
```
osandov, line 119: "Signal frame" confused me here. I think it'd be more accurate to say something like "a frame record with the fp and lr both zero".

libdrgn/arch_aarch64.c
```diff
@@ -68,6 +68,100 @@ static void demangle_cfi_registers_aarch64(struct drgn_program *prog,
...
+		// "metadata" which tells us whether a pt_regs is pushed to the
+		// stack.
+		uint64_t frame_meta_type;
+		err = drgn_program_read_memory(prog, &frame_meta_type, fp + 16,
+					       sizeof(frame_meta_type), false);
+		if (err)
+			return err;
+
+		// Frame metadata code 2 indicates a pt_regs
+		frame_meta_type = drgn_platform_bswap(&prog->platform) ? bswap_64(frame_meta_type) : frame_meta_type;
```
osandov, line 130: You can use `drgn_program_read_u64()` and then you don't need to bswap yourself.

libdrgn/arch_aarch64.c
```diff
@@ -68,6 +68,100 @@ static void demangle_cfi_registers_aarch64(struct drgn_program *prog,
...
+		// indicates that we can read the next word on the stack for
+		// "metadata" which tells us whether a pt_regs is pushed to the
+		// stack.
+		uint64_t frame_meta_type;
+		err = drgn_program_read_memory(prog, &frame_meta_type, fp + 16,
+					       sizeof(frame_meta_type), false);
+		if (err)
+			return err;
+
+		// Frame metadata code 2 indicates a pt_regs
```
osandov, line 129: Could you please mention the name of the metadata value in the kernel for reference (`FRAME_META_TYPE_PT_REGS`)?

osandov (changes requested):
The first two commits look good, so I'll go ahead and merge those.

brenns10:
Thank you! I've gone ahead and moved the `__irq_regs` logic into this branch, rebased on main, and applied your feedback. Running aarch64 tests now, but I don't expect any change.

My only question I'd bring up for the `__irq_regs` approach is whether you even think it's necessary to use the `cpu_present_mask`, since it involves a lot of added C helper code (and I especially don't love the special case of converting 32-bit bitfields to 64). If we could just unconditionally check the `__irq_regs` value for each CPU up to `nr_cpu_ids`, then that would simplify everything by quite a bit. I'm trying to determine whether percpu regions are guaranteed to be setup for all such CPU IDs. If not, we could handle it by just catching the faults. I'll let you know what I find.

brenns10:
So, it seems that percpu global variables are always present for all possible CPUs. However, it's not necessarily guaranteed that every CPU ID less than `nr_cpu_ids` is a "possible" CPU -- the set of CPUs could be sparse. So I guess we need to keep the bitmap :(

PR #551, opened 2025-10-06
vmtest: install e2fsprogs in rootfs

brenns10:
It's necessary for fsrefs and swap tests, but when I recently rebuilt a rootfs, it was not present.


brenns10:
Yeah, I don't know, but it actually happened to me twice so I'm pretty confident! (I keep rebuilding because I had been using `ntfs3` for the vmtest directory... which was a mistake 😢)

Issue #538, opened 2025-08-27
Crash command: bt


brenns10:
https://crash-utility.github.io/help_pages/bt.html

This has lots of options. The basic functionality is very important, but there is a long tail of extra options that are lower priority. One thing that crash does very well that we want is delineation of NMI/IRQ stacks.

brenns10:
Regarding (2) above, it looks like the value crash bt prints in brackets seems to be intended to be the stack pointer:
 https://github.com/crash-utility/crash/blob/62486400d35b258e4e3c40c4bf0daedc231f835a/x86_64.c#L3321-L3322

I modified drgn to print the CFA and it was much further away from what crash printed out than the SP. Honestly I'm not too motivated to track down the 8-byte difference here, so I'll still keep the sp. I think it's not terribly hard to expose the CFA so I may keep that commit around in my branch and we can include or drop it as we see fit.

osandov:
That all sounds good to me. My guess for the 8-byte difference is that crash uses the stack pointer after the return address has been pushed, whereas drgn's stack pointer is the value before pushing/after popping the return address. I'm also inclined to keep it as is.

osandov:
Re: `kernel_stack_trace()`, I am totally down with the use case of providing additional, kernel-specific context. I still have to give it a closer look to have an opinion on the exact API.

What I'm less keen on is the suggested future direction of using it as a way to add architecture-specific heuristics for looking for more stack traces. I'd rather have those be handled directly from `prog.stack_trace()`, even if it means adding custom logic in libdrgn or a plugin mechanism for stack traces.

Re: `THREAD_SIZE`, I want that as a special object asap :)

brenns10:
Noting a few other decisions that I'm making, but obviously are subject to change.

**`bt -l` support:** crash prints file name and line number on a separate line from the frame. This looks dumb, but it's usually necessary because crash prints the _full_ file path, from the root directory, which can be quite long:
```
crash> bt -l 1
PID: 1        TASK: ffff9f8b403db780  CPU: 6    COMMAND: "systemd"
 #0 [ffffad114008fbf8] __schedule at ffffffff893aeeba
    /usr/src/debug/kernel-6.12.0/linux-6.12.0-103.40.4.3.el10uek/kernel/sched/core.c: 5367
 #1 [ffffad114008fc60] schedule at ffffffff893af207
[...]

PR #555, opened 2025-10-09
Crash command: bt

brenns10:
Thanks for the review on the aarch64 unwinding things! That unblocks the bt command, which is implemented in this PR. No rush on this as you know.

The main functionality is exposed via two helpers:
- `kernel_stack_trace()` returns a `LinuxKernelStack()`. This behaves very similarly to a `StackTrace` - it can be iterated and indexed, and printed with `str()`. However, it has additional "segments" that correspond to portions of the stack which are separated by interrupted frames. Each segment is associated with a stack "kind", which the helper can categorize.
- `print_registers()` prints the registers dictionary in a way similar to crash. It isn't exactly the same, because crash actually is looking at the `struct pt_regs`, which frequently contains one or two more items that aren't part of the generic register set. (**It seems like it would be possible to expose the `struct pt_regs` for an interrupted `StackFrame`.** Is that something you'd like to see me include and work into this?)

For the actual command, I've tried to stay close to the crash command, without being too attached to things. I documented a lot of the choices I made in #538, so I won't exhaustively enumerate everything.

Some notable features that I've either included or omitted:

- The `-d` and `-V` flags are new, and not present in crash. I am still on the fence about whether the `-d` flag is really worthwhile. I am very confident that `-V` will be popular.
- The `-r` flag is not yet implemented. It seems lower priority but not hard to do.
- There are several flags that seem low priority, but still useful - `-v` to check for stack overflow evidence, `-e` and `-E` to search for possible exception stack frames (I'm not sure how these are implemented though), and `-t / -T` to display any text symbols found on the task stack memory. Also, `-R` will only print the stack trace if it contains a given function, which seems useful for something like `foreach bt -R dput`, printing every stack currently executing `dput`. I think these are all great, but maybe worth putting into a separate "lower priority bt flags" task. Since at this point, there's actually a lot of crash commands implemented, and `bt` really should be there now.

### DIFF
--- a/drgn/commands/_builtin/crash/_bt.py
+++ b/drgn/commands/_builtin/crash/_bt.py
@@ -1,0 +1,430 @@
+# Copyright (c) 2026, Oracle and/or its affiliates.
+# SPDX-License-Identifier: LGPL-2.1-or-later
+import argparse
+from typing import Any, Iterable, Literal, Optional
+
+from drgn import Architecture, FaultError, MainModule, Object, Program, StackFrame
+from drgn.commands import argument, drgn_argument, mutually_exclusive_group
+from drgn.commands._builtin.crash._rd import _print_memory
+from drgn.commands.crash import (
+    Cpuspec,
+    CrashDrgnCodeBuilder,
+    _crash_foreach_subcommand,
+    _TaskSelector,
+    crash_command,
+    parse_cpuspec,
+    print_task_header,
+)
+from drgn.helpers.common.stack import print_registers
+from drgn.helpers.linux.sched import cpu_curr
+from drgn.helpers.linux.stack import LinuxKernelStack, StackKind, kernel_stack_trace
+
+
+def _sp_align(trace: LinuxKernelStack) -> int:
+    # Some architectures may have shorter address widths than their register
+    # width. We want to align the stack frames nicely, and not waste space on
+    # extra zeros. Determine the width for the stack pointer field empirically
+    # rather than setting it to 8 or 16 based on IS_64_BIT. Subtract 2 to
+    # account for the "0x" prefix. Swallow the LookupError in case the sp is not
+    # known.
+    width = 2
+    for segment in trace.segments:
+        for frame in segment.frames:
+            try:
+                width = max(width, len(hex(frame.sp)))
+            except LookupError:
+                pass
+    return width - 2
+
+
+def _print_frame(
+    prog: Program,
+    index: int,
+    sp: Optional[int],
+    frame: StackFrame,
+    sp_width: int,
+    drgn_style: bool = False,
+    show_file_line: bool = False,
+) -> None:
+    index_str = f"#{index}"
+
+    if drgn_style:
+        try:
+            source_info = " ({}:{}:{})".format(*frame.source())
+        except LookupError:
+            source_info = ""
+        print(f"{index_str:>3s} {frame.name}{source_info}")
+    else:
+        mod_text = ""
+        try:
+            module = prog.module(frame.pc - int(frame.pc > 0 and not frame.interrupted))
+            if not isinstance(module, MainModule):
+                mod_text = f" [{module.name}]"
+        except LookupError:
+            pass
+        if frame.is_inline:
+            print(
+                f"{index_str:>3s} {'(inline)':{sp_width + 2}s} {frame.name}{mod_text}"
+            )
+        elif sp is None:
+            print(
+                f"{index_str:>3s} [{'?' * sp_width}] {frame.name} at {frame.pc:x}{mod_text}"
+            )
+        else:
+            print(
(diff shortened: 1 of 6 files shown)


brenns10:
This is once again rebased. In terms of updates:

- Added tests for the `-f`, `-F`, and `-FF` arguments.
- Fixed a very frustrating issue with s390x stack pointers being 160 bytes offset from the relevant stack data.
- Added the missing `[module name]` annotations for kernel module function names.

I did need to resolve some minor conflicts, especially in the `cli.py` file where we check return types to see whether to format them with `str` or `repr`. It occurs to me that my `DisplayStr` idea may not be the best way to implement this (e.g. it cannot be used with NamedTuples, which are really useful return types).

Do you think it would make sense to have the CLI check `hasattr(value, "_repr_pretty_")` and then use that to format return values, rather than checking against a hard-coded list of types, or my mix-in `DisplayStr` class? I know it's a Jupyter/IPython thing so there are already non-drgn types out there which implement it. But I think that there may be value in making it easier for non-drgn code to declare a pretty-printer that works in the CLI.

**Edit:** and this had a fully clean test run on all supported architectures.

osandov (commented):
I didn't get to reviewing the bulk of this before I head out for Thanksgiving, so I'll just share my comments on the first three commits. Feel free to wait until I look at the rest or address these now.

Re: `DisplayStr`, Using `_repr_pretty_` would require us to emulate IPython's pretty-printer object. A class decorator that adds to some global tuple internal to `drgn` would work for `NamedTuple`s:

```python3
@drgn.cli.display_str
class MyClass:
    ...
```

Would that be better? You could also just call the decorator (`drgn.cli.display_str(YourClass)`), meaning you could register types that you don't necessarily have control over. The implementation would be slower, but we're talking about printing to a terminal at that point anyways.

drgn/helpers/linux/stack.py
```diff
@@ -0,0 +1,461 @@
...
+        segments: List[StackSegment],
+    ) -> None:
+        self.task = task
+        self.cpu = cpu
+        self.on_cpu = on_cpu
+        self.segments = segments
+
+    def __str__(self) -> str:
+        total_frames = sum(len(s.frames) for s in self.segments)
+        framew = 1 if total_frames < 10 else 2
```
osandov: I think this is off by one, since total_frames == 10 would be 0..9. FWIW, `StackTrace.__str__()` always uses a width of 2, even if the trace has less than 10 frames, so maybe you can just use 2 regardless.

drgn/helpers/linux/stack.py
```diff
@@ -0,0 +1,461 @@
...
+
+__all__ = (
+    "StackKind",
+    "StackSegment",
+    "LinuxKernelStack",
+    "kernel_stack_trace",
+)
+
+
+class StackKind(enum.Enum):
```
osandov: This needs a class docstring.

osandov (changes requested):
It's awesome to see this coming together. I gave the helpers the most scrutiny, plus a few more comments throughout.

brenns10:
All architectures & flavors passed, with a few exceptions not related to bt:
```
Failed: aarch64: 6.19.0-rc4-vmtest39.0alternative
        ppc64: 6.7.12-vmtest39.1default, 6.7.12-vmtest39.0tiny, 6.6.119-vmtest39.0alternative

stephen at envy in ~/repos/drgn (crash_bt)
$ python -m vmtest.summarize
Total test runs: 492
TOTAL  FAILED  TEST
-------------------
3      3       tests/linux_kernel/helpers/test_block.py::TestBlock::test_rq_helpers
1      1       tests/linux_kernel/crash_commands/test_timer.py::TestTimer::test_hrtimer

Details by failing test
------------------------
tests/linux_kernel/helpers/test_block.py::TestBlock::test_rq_helpers  (total 3)
  FAILED (3):
    - arches: ppc64
    - vers: 6.6, 6.7
    - flavors: alternative, default, tiny

tests/linux_kernel/crash_commands/test_timer.py::TestTimer::test_hrtimer  (total 1)
  FAILED (1):
    - arches: aarch64
    - vers: 6.19
    - flavors: alternative
```
I will take a look at those outside this PR.

osandov:
I rebased and made a few minor edits:

In "tests: linux_kernel: panic within IRQ on arm & ppc64", I got rid of the error if the architecture is unrecognized to make things easier for RISC-V when that comes along and changed `__PPC64__` to `__powerpc64__`, which is the macro that we check elsewhere in the file.

In "drgn.helpers.common: add print_registers() helper", I changed the module name in the subject to "drgn.helpers.common.stack" and made `__all__` switch to one name on each line.

In "cli: add a decorator to display types with str()", I changed the example code in the docstring to use the full name `@drgn.cli.display_str`.

I'm running a full vmtest run. Assuming that passes, I think this'll be good to go.

osandov:
More updates:

* Pad `_print_rd` with zeroes like crash does for `bt` (but not `rd`).
* Add `@skip_unless_have_stack_tracing` to test cases that need it.

PR #556, opened 2025-10-10
crash: port vm command

qiyeliu:
Add a new 'vm' command to the crash command set.

The 'vm' command shows virtual memory information of a given task or process, including:
 - basic memory statistics (MM, PGD, RSS, TOTAL_VM)
 - a list of vm_area_struct regions with start/end addresses, flags, and file paths
 - support for both normal and DRGN code generation modes
 - graceful handling of kernel threads (no mm_struct) and invalid arguments

If no argument is provided, the current task context is used.

### DIFF
--- a/drgn/commands/_builtin/crash/vm.py
+++ b/drgn/commands/_builtin/crash/vm.py
@@ -1,0 +1,161 @@
+# Copyright (c) 2025, Kylin Software, Inc. and affiliates.
+# SPDX-License-Identifier: LGPL-2.1-or-later
+
+import argparse
+from typing import Any, List, Sequence
+
+from drgn import Object, Program
+from drgn.commands import argument, drgn_argument
+from drgn.commands.crash import (
+    CrashDrgnCodeBuilder,
+    crash_command,
+    crash_get_context,
+    print_task_header,
+)
+from drgn.helpers.common.format import CellFormat, print_table
+from drgn.helpers.linux.mm import for_each_vma, task_rss, vma_name
+
+
+# Generate DRGN code mode output
+def _generate_drgn_code(prog: Program, args: argparse.Namespace) -> None:
+    code = CrashDrgnCodeBuilder(prog)
+    code.add_from_import("drgn", "Object")
+    code.add_from_import(
+        "drgn.helpers.linux.mm", "for_each_vma", "task_rss", "vma_name"
+    )
+    code.add_from_import("drgn.helpers.linux.sched", "task_cpu")
+    code.add_from_import("drgn.helpers.linux.pid", "find_task")
+
+    # Build task list
+    code.append("tasks = []\n")
+    # Handle the case where no arguments are provided
+    if not args.pids_or_tasks:
+        code.append_crash_context(None)
+        code.append("tasks.append(task)\n")
+    else:
+        # Iterate over provided arguments and handle invalid inputs gracefully
+        for pid_or_task in args.pids_or_tasks:
+            code.append_crash_context(pid_or_task)
+            code.append("tasks.append(task)\n")
+
+    # Add processing logic
+    code.append(
+        """\
+for task in tasks:
+"""
+    )
+    code.append_task_header(indent="    ")
+    code.append(
+        """\
+    mm = task.mm.read_()
+    if mm:
+        # Get memory statistics
+        pgd = mm.pgd
+        rss = task_rss(task).total
+        total_vm = mm.total_vm * prog["PAGE_SIZE"]
+        # Get VMA info
+        for vma in for_each_vma(mm):
+            vma_addr = vma
+            vm_start = vma.vm_start
+            vm_end = vma.vm_end
+            vm_flags = vma.vm_flags
+
+            # Get file path
+            file_path = vma_name(vma)
+"""
+    )
+    code.print()
+
+
+# Print virtual memory info for a single task
+def _print_task_vm_info(prog: Program, task: Object) -> None:
+    print_task_header(task)
+
+    mm = task.mm.read_()
+    if not mm:
+        kernel_stats_rows = [["MM", "PGD", "RSS", "TOTAL_VM"], ["0", "0", "0k", "0k"]]
+        print_table(kernel_stats_rows)
+        return
+
+    # Show memory statistics
+    rss_info = task_rss(task)
+    total_vm_kb = mm.total_vm.value_() * prog["PAGE_SIZE"].value_() // 1024
+
+    stats_rows: List[List[CellFormat]] = [
+        [
+            CellFormat("MM", "^"),
+            CellFormat("PGD", "^"),
+            CellFormat("RSS", "^"),
+            CellFormat("TOTAL_VM", "^"),
+        ],
+        [
+            CellFormat(mm.value_(), "^x"),
+            CellFormat(mm.pgd.value_(), "^x"),
+            CellFormat(f"{rss_info.total}k", "^"),
+            CellFormat(f"{total_vm_kb}k", "^"),
+        ],
+    ]
+    print_table(stats_rows)
+    # Show detailed VMA info
+    rows: List[Sequence[Any]] = [
+        [
+            CellFormat("VMA", "^"),
+            CellFormat("START", "^"),
+            CellFormat("END", "^"),
+            CellFormat("FLAGS", "<"),
+            CellFormat("FILE", "<"),
+        ]
+    ]
+    for vma in for_each_vma(mm):
+        file_path = vma_name(vma)
+        rows.append(
+            [
+                CellFormat(vma.value_(), "^x"),
+                CellFormat(vma.vm_start.value_(), "^x"),
+                CellFormat(vma.vm_end.value_(), "^x"),
+                CellFormat(vma.vm_flags.value_(), "<x"),
+                file_path,
+            ]
+        )
+
+    print_table(rows)
+
+
+@crash_command(
+    description="virtual memory",
+    long_description="""This command displays basic virtual memory information of a context,
+consisting of a pointer to its mm_struct and page directory, its RSS and
+total virtual memory size; and a list of pointers to each vm_area_struct,
+its starting and ending address, vm_flags value, and file pathname. If no
+arguments are entered, the current context is used.
+""",
+    arguments=(
+        argument(
+            "pids_or_tasks",
+            metavar="pid|task",
+            nargs="*",
+            type="pid_or_task",
+            help="one or more process PIDs or hexadecimal task_struct pointers",
+        ),
+        drgn_argument,
+    ),
+)
+def _crash_cmd_vm(
+    prog: Program, name: str, args: argparse.Namespace, **kwargs: Any
+) -> None:
+
+    if args.drgn:
+        _generate_drgn_code(prog, args)
+        return
+
+    if not args.pids_or_tasks:
+        args.pids_or_tasks.append(None)
+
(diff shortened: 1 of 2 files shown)


drgn/commands/_builtin/crash/vm.py
```diff
@@ -0,0 +1,233 @@
...
+            f"{mm.value_():#x}",
+            f"{mm.pgd.value_():#x}",
+            f"{rss_info.total}k",
+            f"{total_vm_kb}k",
+        ],
+    ]
+    print_table(stats_rows)
+
+    # Show detailed VMA info
+    rows = [["VMA", "START", "END", "FLAGS", "FILE"]]
```
osandov: Similar here, VMA, START, and END are centered, then FLAGS should be left-justified.

drgn/commands/_builtin/crash/vm.py
```diff
@@ -0,0 +1,233 @@
...
+    code.append(
+        """\
+for task in tasks:
+    # Get basic task info
+    pid = int(task.pid)
+    task_addr = task.value_()
+    cpu = task_cpu(task)
+    command = task.comm.string_().decode()
+
+    mm=task.mm
```
osandov: Same comment as for the non-drgn case (and style nit).
```suggestion
    mm = task.mm.read_()
```

drgn/commands/_builtin/crash/vm.py
```diff
@@ -0,0 +1,233 @@
...
+from drgn.helpers.linux.fs import inode_path
+from drgn.helpers.linux.mm import for_each_vma, get_task_rss_info
+from drgn.helpers.linux.sched import task_cpu
+
+
+def _get_file_path(vma: Object, mm: Object) -> str:
+    """Get the file path name corresponding to the VMA"""
+    if vma.vm_file:
+        try:
+            path_bytes = inode_path(vma.vm_file.f_inode)
```
osandov: `inode_path()` is relative to the containing filesystem. Use `d_path(vma.vm_file.f_path)` for an absolute path.

drgn/commands/_builtin/crash/vm.py
```diff
@@ -0,0 +1,233 @@
...
+        and mm.context.vdso.value_() >= start
+        and mm.context.vdso.value_() < end
+    ):
+        return "[vdso]"
+
+    return "[anon]"
+
+
+def _get_tasks_from_args(prog: Program, pids_or_tasks: List[str]) -> List[Object]:
+    """Get task list from command line arguments"""
```
osandov: A drgn code style quirk is that we only use docstrings for public functions (mainly because our documentation generation system uses the existence of a docstring to decide whether to include a function). Internal functions like this should just use a normal `#` comment.

drgn/commands/_builtin/crash/vm.py
```diff
@@ -0,0 +1,233 @@
...
+
+
+def _print_task_vm_info(prog: Program, task: Object) -> None:
+    """Print virtual memory info for a single task"""
+    pid = int(task.pid)
+    cpu = task_cpu(task)
+    command = task.comm.string_().decode()
+    task_addr = task.value_()
+
+    print(f'PID: {pid}\tTASK: {task_addr:#x}\tCPU: {cpu}\tCOMMAND: "{command}"')
```
osandov: I just added `drgn.commands.crash.print_task_header()` that you can use for this.

drgn/helpers/linux/mm.py
```diff
@@ -1595,6 +1597,56 @@ def vma_find(mm: Object, addr: IntegerLike) -> Object:
...
+        return "[stack]"
+    elif (
+        hasattr(mm.context, "vdso")
+        and mm.context.vdso
+        and mm.context.vdso.value_() >= start
+        and mm.context.vdso.value_() < end
+    ):
+        return "[vdso]"
+
+    return "[anon]"
```
osandov: This isn't necessarily correct for special VMAs that don't have a `vma->vm_ops->name`, right? E.g., if you have a special device mmap, then it's not anonymous memory but it's also not any of the other cases. Maybe we should return `None` in this case, similar to how `get_vma_name()` returns a `NULL` name.

drgn/commands/_builtin/crash/vm.py
```diff
@@ -0,0 +1,179 @@
...
+
+def _print_task_vm_info(prog: Program, task: Object) -> None:
+    """Print virtual memory info for a single task"""
+    print_task_header(task)
+
+    mm = task.mm.read_()
+    if not mm:
+        kernel_stats_rows = [["MM", "PGD", "RSS", "TOTAL_VM"], ["0", "0", "0k", "0k"]]
+        print_table(kernel_stats_rows)
+        print()
```
osandov: We already print a blank line in the caller, so I don't think this is needed.
```suggestion
```

drgn/commands/_builtin/crash/vm.py
```diff
@@ -0,0 +1,179 @@
...
+        _generate_drgn_code(prog, args)
+        return
+
+    # Non-drgn mode
+    tasks = _get_tasks_from_args(prog, args.pids_or_tasks)
+
+    for task in tasks:
+        if task:  # Add null check
+            _print_task_vm_info(prog, task)
+            print()
```
osandov: `crash_get_context()` already does all of the work here, so I think you can drop `_get_tasks_from_args()` and simplify this to:

```suggestion
    if not args.pids_or_tasks:
        args.pids_or_tasks.append(None)

    for task_arg in args.pids_or_tasks:
        try:
            task = crash_get_context(prog, task_arg)
        except Exception as e:
            print("vm:", e)
        else:
            _print_task_vm_info(prog, task)
        print()
```

drgn/commands/_builtin/crash/vm.py
```diff
@@ -0,0 +1,179 @@
...
+        code.append("tasks.append(task)\n")
+    else:
+        # Iterate over provided arguments and handle invalid inputs gracefully
+        for pid_or_task in args.pids_or_tasks:
+            code.append(f"# Processing input: {pid_or_task!r}\n")
+            try:
+                code.append_crash_context(pid_or_task)
+                code.append("tasks.append(task)\n")
+            except ValueError as e:
+                code.append(f"# Skipping invalid input: {pid_or_task!r} ({e})\n")
```
osandov: I don't think the comments are necessary, and I don't think `append_crash_context` can raise `ValueError`, so this can be:
```suggestion
            code.append_crash_context(pid_or_task)
            code.append("tasks.append(task)\n")
```

osandov (changes requested):
Thanks! I'm really excited for the `vma_name()` helper, that just needs a few tweaks. Some minor nitpicks on the command, too.

drgn/helpers/linux/mm.py
```diff
@@ -1595,6 +1597,56 @@ def vma_find(mm: Object, addr: IntegerLike) -> Object:
         return cast("struct vm_area_struct *", mtree_load(mt, addr))
 
 
+def vma_name(vma: Object, mm: Object) -> str:
```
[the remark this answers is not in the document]
qiyeliu: About  "[vdso]"
```
def vma_name(vma: Object) -> str:
    vm_file = vma.vm_file.read_()
    if vm_file:
        return os.fsdecode(d_path(vm_file.f_path))

    # TODO : vma->vm_ops->name and arch_vma_name

    # Handle special memory regions
    if not vma.vm_mm:
        return "[vdso]"

    start = vma.vm_start.value_()
    end = vma.vm_end.value_()

    if start < vma.vm_mm.brk.value_() and end > vma.vm_mm.start_brk.value_():
        return "[heap]"

    if start <= vma.vm_mm.start_stack.value_() and end >= vma.vm_mm.start_stack.value_():
        return "[stack]"

    if vma.vm_mm:
        anon_name = None
        try:
            anon_name = vma.vm_mm.anon_vma_name.read_()
        except AttributeError:
            pass

        if anon_name:
            return f"[anon:{anon_name.name.string_().decode()}]"
    # If no conditions are met, return None
    return ""

```
```
/proc/xx/maps
7fc99ec2c000-7fc99ec30000 r--p 00000000 00:00 0                          [vvar]
7fc99ec30000-7fc99ec32000 r--p 00000000 00:00 0                          [vvar_vclock]
7fc99ec32000-7fc99ec34000 r-xp 00000000 00:00 0                          [vdso]
7fc99ec34000-7fc99ec60000 r-xp 00000000 fd:00 67112826                   /usr/lib64/ld-linux-x86-64.so.2
7fc99ec60000-7fc99ec62000 r--p 0002c000 fd:00 67112826                   /usr/lib64/ld-linux-x86-64.so.2
```

%crash >   vm   xx

```
ffff8ae86532f000  7fc99ec2c000  7fc99ec30000  c044411   
ffff8ae83ba51600  7fc99ec30000  7fc99ec32000  c044411   
ffff8ae83c8a1900  7fc99ec32000  7fc99ec34000  8040075   
ffff8ae7c3ef59c0  7fc99ec34000  7fc99ec60000  8000075   /usr/lib64/ld-linux-x86-64.so.2
ffff8ae83569ef00  7fc99ec60000  7fc99ec62000  8100071   /usr/lib64/ld-linux-x86-64.so.2
ffff8ae829e59540  7fc99ec62000  7fc99ec64000  8100073   /usr/lib64/ld-linux-x86-64.so.2

```
Regarding vdso, the fields can be displayed using the code above via /proc/xx/maps, but not using the vm command.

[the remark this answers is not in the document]
osandov: Oh, it looks like on x86, the `[vdso]` name comes from `special_mapping_name()`. So it'd need to be something like:

```python3
vm_ops = vma.vm_ops.read_()
if vm_ops:
    vm_ops_name = vm_ops.name.read_()
    if vm_ops_name == prog["special_mapping_name"]:
        return cast("struct vm_special_mapping *", vma.vm_private_data).name.string_().decode()
```

drgn/helpers/linux/mm.py
```diff
@@ -1595,6 +1598,84 @@ def vma_find(mm: Object, addr: IntegerLike) -> Object:
...
+
+    if mm.start_stack and start <= mm.start_stack.value_() <= end:
+        return "[stack]"
+
+    try:
+        anon_name = mm.anon_vma_name.read_()
+        if anon_name and anon_name.name:
+            return f"[anon:{os.fsdecode(anon_name.name.string_())}]"
+    except AttributeError:
+        pass
```
osandov: A few things here:

1. The anonymous name is `vma->anon_name`, not `mm->anon_vma_name`.
2. `anon_name->name` is a flexible array, which can't be `NULL`, so we don't need to check that.
3. We should limit the scope of the `try` block to only the `vma.anon_name` access so that we don't mask any other failed member accesses (in case members get renamed in the future, for example).
4. We should document the reason that we need the check.

```suggestion
    # anon_name was added in Linux kernel commit 9a10064f5625 ("mm: add a field
    # to store names for private anonymous memory") (in v5.17). It also only
    # exists if CONFIG_ANON_VMA_NAME=y.
    try:
        anon_name = vma.anon_name.read_()
    except AttributeError:
        pass
    else:
        if anon_name:
            return f"[anon:{os.fsdecode(anon_name.name.string_())}]"
```

I'll work on enabling `CONFIG_ANON_VMA_NAME` in drgn's vmtest kernels so we can test this codepath, but that doesn't block landing this.

osandov (changes requested):
After these final few comments I think this will be ready to land, thanks again.

PR #559, opened 2025-10-17
crash: add bpf command for displaying eBPF programs and maps

suchit07-git:
Introduce a new crash command 'bpf' to list eBPF programs and maps from the kernel. The command enumerates over all loaded BPF programs and maps, showing each program’s ID, addresses, type, tag, and the maps it uses.

### DIFF
--- a/drgn/commands/_builtin/crash/_bpf.py
+++ b/drgn/commands/_builtin/crash/_bpf.py
@@ -1,0 +1,130 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# SPDX-License-Identifier: LGPL-2.1-or-later
+
+# ebpf-related commands.
+
+import argparse
+from typing import Any, List, Sequence
+
+from drgn import Program
+from drgn.commands import drgn_argument
+from drgn.commands.crash import CrashDrgnCodeBuilder, crash_command
+from drgn.helpers.common.format import CellFormat, print_table
+from drgn.helpers.linux.bpf import (
+    bpf_map_for_each,
+    bpf_prog_for_each,
+    bpf_prog_used_maps,
+)
+
+
+@crash_command(
+    description="display all eBPF programs and maps",
+    arguments=(drgn_argument,),
+)
+def _crash_cmd_bpf(
+    prog: Program, name: str, args: argparse.Namespace, **kwargs: Any
+) -> None:
+    if args.drgn:
+        code = CrashDrgnCodeBuilder(prog)
+        code.add_from_import(
+            "drgn.helpers.linux.bpf",
+            "bpf_map_for_each",
+            "bpf_prog_for_each",
+            "bpf_prog_used_maps",
+        )
+        code.append(
+            """\
+for bpf_prog in bpf_prog_for_each(prog):
+    aux = bpf_prog.aux
+    prog_id = aux.id
+    prog_type = bpf_prog.type
+    tag = bpf_prog.tag
+    for used_map in bpf_prog_used_maps(bpf_prog):
+        map_id = used_map.id
+
+for bpf_map in bpf_map_for_each(prog):
+    map_id = bpf_map.id
+    map_type = bpf_map.map_type
+    try:
+        map_flags = bpf_map.map_flags
+    except AttributeError:
+        # map_flags is only available since Linux 4.6.
+        pass
+    """
+        )
+        code.print()
+        return
+
+    prog_rows: List[Sequence[Any]] = [
+        (
+            CellFormat("ID", "^"),
+            CellFormat("BPF_PROG", "^"),
+            CellFormat("BPF_PROG_AUX", "^"),
+            CellFormat("BPF_PROG_TYPE", "^"),
+            CellFormat("TAG", "^"),
+            CellFormat("USED_MAPS", "^"),
+        )
+    ]
+
+    for bpf_prog in bpf_prog_for_each(prog):
+        prog_id = bpf_prog.aux.id.value_()
+        prog_type_name = bpf_prog.type.format_(type_name=False).split("BPF_PROG_TYPE_")[
+            -1
+        ]
+
+        tag = bpf_prog.tag
+        prog_tag = "".join(f"{b.value_():02x}" for b in tag)
+
+        used_maps = []
+        for map in bpf_prog_used_maps(bpf_prog):
+            used_maps.append(map.id.value_())
+        used_maps_str = ",".join(str(m) for m in used_maps)
+
+        prog_rows.append(
+            (
+                prog_id,
+                CellFormat(bpf_prog.value_(), "^x"),
+                CellFormat(bpf_prog.aux.value_(), "^x"),
+                CellFormat(prog_type_name, "^"),
+                CellFormat(prog_tag, "^"),
+                CellFormat(used_maps_str, "^"),
+            )
+        )
+
+    print_table(prog_rows)
+    print()
+
+    map_rows: List[Sequence[Any]] = [
+        (
+            CellFormat("ID", "^"),
+            CellFormat("BPF_MAP", "^"),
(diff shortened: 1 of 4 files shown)


suchit-k:
@osandov, following up on this PR just in case it you haven’t had a chance to look at this PR yet.

suchit07-git:
Thanks for the heads up @osandov. Sorry if I bugged you.

drgn/commands/_builtin/crash/_bpf.py
```diff
@@ -0,0 +1,271 @@
...
+from typing import Any, Generator, Optional, Tuple
+
+from drgn import Program
+from drgn.commands.crash import crash_command
+from drgn.helpers.common.type import enum_type_to_class
+from drgn.helpers.linux import (  # type: ignore[attr-defined]
+    bpf_map_for_each,
+    bpf_prog_for_each,
+    hlist_for_each_entry,
+)
```
osandov: Import these from the specific submodules where they're defined (`drgn.helpers.linux.bpf` and `drgn.helpers.linux.list`) to avoid the need for the `type: ignore`.

drgn/commands/_builtin/crash/_bpf.py
```diff
@@ -0,0 +1,271 @@
...
+    )
+    BpfLinkType = enum_type_to_class(prog.type("enum bpf_link_type"), "BpfLinkType")
+    BpfProgTrampType = enum_type_to_class(
+        prog.type("enum bpf_tramp_prog_type"), "BpfProgTrampType"
+    )
+
+    print(
+        f"{'ID':>4}  {'BPF_PROG':<18} {'BPF_PROG_AUX':<18} {'BPF_PROG_TYPE':<18} "
+        f"{'TAG':<18} {'USED_MAPS'}"
+    )
```
osandov: I'd prefer use of `drgn.helpers.common.format.print_table()` rather than manual formatting. There are lots of example in `drgn/commands`.

suchit07-git:
Hi @osandov. Thanks for the comments. I’ll make the required changes and let you know once they’re done. Apologies for the inconvenience.

drgn/helpers/linux/bpf.py
```diff
@@ -180,3 +183,59 @@ def cgroup_bpf_prog_for_each_effective(
             if not prog:
                 break
             yield prog
+
+
+def bpf_prog_used_maps(bpf_prog: Object) -> Iterator[Object]:
+    """Yield maps used by a BPF program."""
```
osandov: The convention for helpers is to document the C type of object arguments/return values:
```suggestion
    """
    Yield maps used by a BPF program.

    :param bpf_prog: ``struct bpf_prog *``
    :return: Iterator of ``struct bpf_map *`` objects.
    """
```

drgn/helpers/linux/bpf.py
```diff
@@ -180,3 +183,59 @@ def cgroup_bpf_prog_for_each_effective(
...
+    for i in range(bpf_prog.aux.used_map_cnt.value_()):
+        yield bpf_prog.aux.used_maps[i]
```
osandov: This requires reading `bpf_prog.aux` twice, so as a small optimization, you can read it ahead of time. We can also use a slicing trick to optimize this further:


```suggestion
    aux = bpf_prog.aux.read_()
    return iter(aux.used_maps[:aux.used_map_cnt])
```

drgn/commands/_builtin/crash/_bpf.py
```diff
@@ -0,0 +1,135 @@
...
+
+@crash_command(
+    description="display all eBPF programs and maps",
+)
+def _crash_cmd_bpf(
+    prog: Program, name: str, args: argparse.Namespace, **kwargs: Any
+) -> None:
+    prog_rows: List[Sequence[Any]] = [
+        (
+            "ID",
```
osandov: It looks like crash centers this heading:


```suggestion
            CellFormat("ID", "^"),
```

drgn/commands/_builtin/crash/_bpf.py
```diff
@@ -0,0 +1,135 @@
...
+        used_maps = []
+        for map in bpf_prog_used_maps(bpf_prog):
+            used_maps.append(map.id.value_())
+        used_maps_str = ", ".join(str(m) for m in used_maps)
+
+        prog_rows.append(
+            (
+                prog_id,
+                CellFormat(prog_addr, "^"),
+                CellFormat(aux_addr, "^"),
```
osandov: Delete the manual formatting for `prog_addr` and `aux_addr` above and do it all with `CellFormat` instead:

```suggestion
                CellFormat(bpf_prog.value_(), "^x"),
                CellFormat(bpf_prog.aux.value_(), "^x"),
```

drgn/commands/_builtin/crash/_bpf.py
```diff
@@ -0,0 +1,135 @@
...
+
+        bpf_map_type = prog.type("enum bpf_map_type")
+        map_type = bpf_map.map_type.value_()
+
+        map_type_name = ""
+        if bpf_map_type.enumerators:
+            for name, value in bpf_map_type.enumerators:
+                if value == map_type:
+                    map_type_name = name
+        map_type_name = map_type_name.split("BPF_MAP_TYPE_")[-1]
```
osandov: Same enum trick applies here.

drgn/commands/_builtin/crash/_bpf.py
```diff
@@ -0,0 +1,136 @@
...
+        map_type_name = bpf_map.map_type.format_(type_name=False).split(
+            "BPF_MAP_TYPE_"
+        )[-1]
+
+        # The 'map_flags' field was added in Linux kernel commit 6c9059817432
+        # ("bpf: pre-allocate hash map elements") in version 4.6. Older kernels
+        # may not have this field, so we catch LookupError and default to 0.
+        try:
+            map_flags = f"{bpf_map.map_flags.value_():08x}"
+        except LookupError:
```
osandov: This should be `AttributeError`:
```suggestion
        except AttributeError:
```

suchit07-git:
Also, after this I plan to work on printing detailed information for BPF programs and maps using their IDs.

osandov:
Okay I took a closer look at the test failures.

* The kprobe program issue was caused by stricter checks before Linux 5.0. I just landed a fix here: 02ecc2c31f7d8373daaefc665dadcb6e44a53177. So it'd probably be good to add the kprobe program back to the test.

* Using `bpf_prog_get_info_by_fd()` technically works to get the ID of a map, but I just added `bpf_map_get_info_by_fd()` for that purpose: 84dcc0cb54dfe1d9471150718fc1647c76183649.

* Most importantly, I forgot that none of the ID stuff even exists before torvalds/linux@dc4bb0e2356149aee4cdae061936f3bbdd45595c and torvalds/linux@f3f1c054c288bb6e503005e6d73611151ed20e91 in v4.13. Crash also bails on v4.9: `bpf: command not supported or applicable on this architecture or kernel`. So there's no point trying to support it. The BPF helper tests already have a workaround for this: https://github.com/osandov/drgn/blob/84dcc0cb54dfe1d9471150718fc1647c76183649/tests/linux_kernel/helpers/test_bpf.py#L166-L175

  So you could probably add something like this to your test class (untested):

  ```python3
      @classmethod
      def setUpClass(cls):
          super().setUpClass()
          # This command isn't supported before Linux kernel commits dc4bb0e23561
          # ("bpf: Introduce bpf_prog ID") and f3f1c054c288 ("bpf: Introduce bpf_map
          # ID") (in v4.13).
          try:
              next(bpf_prog_ids())
          except OSError as e:
              if e.errno != errno.EINVAL:
                  raise
              raise unittest.SkipTest("kernel does not support BPF object IDs")
  ```

PR #560, opened 2025-10-20
contrib: Add script to display mutex owner and waiters

mfleming:
Knowing which tasks are holding or waiting on a mutex can be particularly helpful during a debugging session, e.g. when simple commands like cat /proc/$foo start to block forever because some other task is holding the mutex and will not release it:
```
   mfleming@machine:~$ sudo ./mutex-owner event_mutex
   Mutex details for 'event_mutex'
   ========================================
           Owner: prometheus-ebpf-340210
           Waiters:
                   wc-346534
                   cat-2945595
```

### DIFF
--- a/contrib/mutex_owner.py
+++ b/contrib/mutex_owner.py
@@ -1,0 +1,40 @@
+#!/usr/bin/env drgn
+# Copyright (c) Cloudflare
+# SPDX-License-Identifier: LGPL-2.1-or-later
+
+"""List the task holding a mutex and which tasks are waiting."""
+
+import sys
+import drgn
+from drgn import FaultError, NULL, Object, cast, container_of, execscript, offsetof, reinterpret, sizeof, stack_trace
+from drgn.helpers.common import *
+from drgn.helpers.linux import *
+
+def main(mutex_name):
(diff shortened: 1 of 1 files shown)


contrib/mutex_owner.py
```diff
@@ -0,0 +1,43 @@
...
+
+def main(mutex_name):
+	print("Mutex details for '%s'" % mutex_name)
+	print("="*40)
+
+	m = prog[mutex_name]
+	addr = cast("unsigned long", m.owner.counter) & ~0x7
+	if addr != 0:
+		t = cast("struct task_struct *", addr)
+		owner = "{}-{}".format(comm_to_str(t.comm.value_()), int(t.pid))
```
osandov: The typical way to get a string from a character array is:

```suggestion
		owner = "{}-{}".format(t.comm.string_().decode(), int(t.pid))
```

Same thing below, then you can delete `comm_to_str`.

osandov (commented):
Thanks! We should probably have a helper for getting the mutex owner in the future. I left one minor suggestion.

osandov (approved):
Thank you!

PR #561, opened 2025-10-22
crash: port rd command

brenns10:
This implements _almost_ every option for crash's rd command. With the `identify_address_all()` API it's now trivial to make the symbolized output identical to crash's. There are a lot of formatting options, which I tried to cover with a generic unit test suite that is independent from the kernel tests. And there are also linux kernel crash command tests.

The `_print_memory()` API will get used by the `bt` command to print the formatted memory between stack frames. I didn't think it merited being an exported helper API because it's so crash-specific. If necessary that can always be changed.

The only options missing are:

- `-f` to print dumpfile offsets. This is just too low-level for drgn, and I don't see an easy way to include it. I also have never used it myself.
- `-m` to print Xen host machine memory. In 2025 I'm not sure how relevant this is, and I don't think we actually support the Xen dumps necessary for this to even make sense.

Userspace memory is supported too, using the trick we discussed on matrix.

### DIFF
--- a/drgn/commands/_builtin/crash/_rd.py
+++ b/drgn/commands/_builtin/crash/_rd.py
@@ -1,0 +1,547 @@
+# Copyright (c) 2025, Oracle and/or its affiliates.
+# SPDX-License-Identifier: LGPL-2.1-or-later
+import argparse
+import operator
+import string
+from typing import Any, Dict, Literal, Optional
+
+from drgn import FaultError, IntegerLike, PlatformFlags, Program
+from drgn.commands import argument, drgn_argument, mutually_exclusive_group
+from drgn.commands.crash import CrashDrgnCodeBuilder, crash_command, crash_get_context
+from drgn.helpers.common.format import escape_ascii_string
+from drgn.helpers.common.memory import IdentifiedSymbol, identify_address_all
+from drgn.helpers.linux.common import IdentifiedSlabObject
+from drgn.helpers.linux.mm import access_process_vm
+
+
+def _crash_annotate(
+    prog: Program,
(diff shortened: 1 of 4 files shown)


osandov (changes requested):
This is great, thank you. Just a few nitpicks.

brenns10:
No, this is in the `kmodify` helpers:
```
ERROR: test_string_args (tests.linux_kernel.helpers.test_kmodify.TestCallFunction.test_string_args) [array pointer]
  File "/home/stepbren/repos/drgn/tests/linux_kernel/helpers/test_kmodify.py", line 183, in test_string_args
    self.assert_called(
  File "/home/stepbren/repos/drgn/tests/linux_kernel/helpers/test_kmodify.py", line 34, in assert_called
    return_value = call_function(self.prog["drgn_kmodify_test_" + name], *args)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/stepbren/repos/drgn/drgn/helpers/common/prog.py", line 115, in wrapper
    return f(args[0].prog_, *args, **kwds)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/stepbren/repos/drgn/drgn/helpers/experimental/kmodify.py", line 1501, in call_function
    implicit_convert(
TypeError: cannot convert 'char *' to incompatible type 'char *'

```

It is only reproducing on my Oracle Linux laptop, not my Arch desktop or the Debian rootfs. I'm tracking it down a bit further but will file an issue if it is actually a bug.

PR #575, opened 2025-11-25
crash: port log command

NagappanRMPL:
crash: port log command.
Please review and let me know your comments.

### DIFF
--- a/drgn/commands/_builtin/crash/_sys.py
+++ b/drgn/commands/_builtin/crash/_sys.py
@@ -56,6 +56,7 @@ from drgn.helpers.linux.mm import totalram_pages
 from drgn.helpers.linux.panic import panic_message, panic_task
 from drgn.helpers.linux.percpu import per_cpu
 from drgn.helpers.linux.pid import for_each_task
+from drgn.helpers.linux.printk import print_dmesg
 from drgn.helpers.linux.sched import (
     get_task_state,
     loadavg,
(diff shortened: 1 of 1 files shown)


drgn/commands/_builtin/crash/_log.py
```diff
@@ -0,0 +1,55 @@
...
+            dest="time_stamp",
+	    action="store_true",
+            help="Dump kernel dmesg in human readable time",
+        ),
+        argument(
+            "-t",
+            dest="no_time_stamp",
+            action="store_true",
+            help="Dump kernel dmesg without timestamp",
+        ),
```
osandov: A couple of things here:

1. It doesn't make sense to specify both `-t` and `-T`, so these should be in a `mutually_exclusive_group()`.
2. You can make use of `store_const` to set the proper `timestamps` argument directly instead of needing the conditionals below. Something like:

```suggestion
        mutually_exclusive_group(
            argument(
                "-T",
                dest="timestamps",
                action="store_const",
                const="human",
                default=True,
                help="Dump kernel dmesg in human readable time",
            ),
            argument(
                "-t",
                dest="timestamps",
                action="store_false",
                default=argparse.SUPPRESS,
                help="Dump kernel dmesg without timestamp",
            ),
        ),
```

Then all you need do is call `print_dmesg(prog, timestamps=args.timestamps)`.

NagappanRMPL:
> Thank you! I merged this and also added support for the `--drgn` option and some basic test cases: [7e3049f](https://github.com/osandov/drgn/commit/7e3049ffb9a6bfd03996328c45aff5f420d61c2e).
> 
> I also created a followup draft issue for the remaining options: [drgn Crash Compatibility Mode](https://github.com/users/osandov/projects/2?pane=issue&itemId=144403701). Are you interested in working on any of those (particularly `-d`, `-m`, and/or `-c`)? It's no problem if not, just let me know.

@osandov - yes, I am interested to further implement `-d`, `-m`, and/or `-c`. I will go through inputs given in https://github.com/users/osandov/projects/2/views/1?pane=issue&itemId=144403701. Thanks!

PR #585, opened 2026-01-02
Swaitq work

imran-kn:
Helpers to work with simple wait queues. Can be used later for completion variable helpers.

### DIFF
--- a/tests/linux_kernel/helpers/test_swait.py
+++ b/tests/linux_kernel/helpers/test_swait.py
@@ -1,0 +1,25 @@
+# Copyright (c) 2026, Oracle and/or its affiliates.
+# SPDX-License-Identifier: LGPL-2.1-or-later
+
+from drgn.helpers.linux.swait import swait_active, swait_for_each_task
+from tests.linux_kernel import LinuxKernelTestCase, skip_unless_have_test_kmod
+
+
+@skip_unless_have_test_kmod
+class TestSimpleWaitqueue(LinuxKernelTestCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.swaitq = cls.prog["drgn_test_swaitq"].address_of_()
+        cls.empty_swaitq = cls.prog["drgn_test_empty_swaitq"].address_of_()
+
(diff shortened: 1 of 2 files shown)


drgn/helpers/linux/swait.py
```diff
@@ -0,0 +1,44 @@
...
+from drgn import Object
+from drgn.helpers.linux.list import list_empty, list_for_each_entry
+
+__all__ = (
+    "swaitqueue_active",
+    "swaitqueue_for_each_task",
+)
+
+
+def swaitqueue_active(wq: Object) -> bool:
```
osandov: The equivalent kernel function is named `swait_active()`, so let's use the same name. (It's inconsistent with `waitqueue_active()`, but I prefer to copy the kernel's inconsistencies rather than invent our own names for the sake of consistency).

drgn/helpers/linux/swait.py
```diff
@@ -0,0 +1,44 @@
...
+    """
+    Return whether a simple wait queue has any waiters.
+
+    :param wq: ``struct swait_queue_head *``
+    """
+    return not list_empty(wq.task_list.address_of_())
+
+
+def swaitqueue_for_each_task(wq: Object) -> Iterator[Object]:
```
osandov: Similar here, the kernel seems to prefer the name `swait_*`, so let's name this `swait_for_each_task()`.

drgn/helpers/linux/swait.py
```diff
@@ -0,0 +1,44 @@
...
+    """
+    Iterate over all tasks waiting on a simple wait queue.
+
+    :param wq: ``struct swait_queue_head *``
+    :return: Iterator of ``struct task_struct *`` objects.
+    """
+    for entry in list_for_each_entry(
+        wq.prog_.type("struct swait_queue"), wq.task_list.address_of_(), "task_list"
```
osandov: No need for the explicit type lookup, `list_for_each_entry()` does it automatically:
```suggestion
        "struct swait_queue", wq.task_list.address_of_(), "task_list"
```

osandov (approved):

PR #586, opened 2026-01-07
Completion work

imran-kn:
Helpers to work with completion variables.

### DIFF
--- a/tests/linux_kernel/helpers/test_completion.py
+++ b/tests/linux_kernel/helpers/test_completion.py
@@ -1,0 +1,25 @@
+# Copyright (c) 2026, Oracle and/or its affiliates.
+# SPDX-License-Identifier: LGPL-2.1-or-later
+
+from drgn.helpers.linux.completion import completion_done, completion_for_each_task
+from tests.linux_kernel import LinuxKernelTestCase, skip_unless_have_test_kmod
+
+
+@skip_unless_have_test_kmod
+class TestCompletion(LinuxKernelTestCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.completion = cls.prog["drgn_test_completion"].address_of_()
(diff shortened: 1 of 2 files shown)


drgn/helpers/linux/completion.py
```diff
@@ -0,0 +1,52 @@
...
+)
+
+
+def completion_done(completion: Object) -> bool:
+    """
+    Test if a completion has any waiters.
+
+    :param completion: ``struct completion *``
+    """
+    return completion.done.value_()
```
osandov: `struct completion::done` an `unsigned int`, so this would return a Python `int`. This should be:
```suggestion
    return bool(completion.done)
```

drgn/helpers/linux/completion.py
```diff
@@ -0,0 +1,52 @@
...
+    # wait_queue_head_t to struct swait_queue_head.
+    # Differentiate between these 2 cases by looking for task_list
+    # member under completion::wait.
+    try:
+        _ = completion.wait.task_list
+        for task in swait_for_each_task(completion.wait.address_of_()):
+            yield task
+    except AttributeError:
+        for task in waitqueue_for_each_task(completion.wait.address_of_()):
+            yield task
```
osandov: A few things here:
1. Usually I do suggest EAFP, but in this case since you aren't using `task_list` directly, a `hasattr()` check makes more sense.
2. You can return the underlying iterators directly instead of adding a for ... in ... yield.
3. We should document the reason for the conditional.
```suggestion
    wait = completion.wait.address_of_()
    # completion->wait is a simple wait queue since Linux kernel commit
    # a5c6234e1028 ("completion: Use simple wait queues") (in v5.7). Detect
    # that by checking for a task_list member.
    if hasattr(wait, "task_list"):
        return swait_for_each_task(wait)
    else:
        return waitqueue_for_each_task(wait)
```

osandov (changes requested):
Good catch! One final comment, it should be good otherwise.

Issue #463, opened 2025-01-22
API for creating a custom Linux kernel program


brenns10:
Currently, it's not documented anywhere, but the only way to create a `Program` that has the `IS_LINUX_KERNEL` flag set is to go through `drgn_program_set_core_dump_fd_internal()`.
https://github.com/osandov/drgn/blob/64d82ddf6e1aa63cae7e4e395657100a9628e26c/libdrgn/program.c#L327-L691

This means that it's not possible to create a Program that represents the kernel with a custom memory reader. This was the goal of #246, which enabled a really interesting use case, even if setting program flags may not be the preferred way to do so.

It seems to me that as of now, it wouldn't be terribly hard to support this. As far as I can tell, three things would be necessary:

1. The VMCOREINFO would need to have been provided at the creation of the program, along with the platform.
2. A Linux kernel object finder should be created and added to the Program.
3. A page table memory reader should be created and added to the Program.
4. The `IS_LINUX_KERNEL` flag should be set.

With that, I believe that the next time `Program.loaded_modules()` is called, the kernel module iterator would be activated, and assuming the memory readers work, drgn should be able to proceed as normal.

I was thinking one interesting way to achieve this could be a `Program.linux_kernel_main_module()` function which does the above and then returns the kernel main module. However, I'm not sure that it's the right API, so I feel like that's an area to discuss.

brenns10:
Notes from our discussion in the January meeting:

* This dovetails with the gdbstub support. We'll need a good way to initialize Linux kernel programs for gdbstub targets.
* Another use case @osandov had was a script for creating a `struct pt_regs` from an oops message and doing some unwinding. It would have been nice to create a program, add a vmlinux, and then start using it as a kernel program without memory readers.
* The complexity of Program initialization makes it a bit difficult to support these things. Currently a new, blank Program is technically a "userspace core dump" according to the program flags. Really, it should probably be created in some uninitialized state, but that would introduce some backward incompatibility.
* We discussed adding a `is_linux_kernel` flag to the `Program` constructor which could set the corresponding flag, object finder, and page table memory reader.
* We could also just reuse the `vmcoreinfo` argument in the program constructor as a way to signal that we should set the `IS_LINUX_KERNEL` flag. The downside is that, if you don't have a vmcoreinfo, you would have to forge it or pass an empty string. (We currently require that vmcoreinfo at least contain an osrelease and page size, as well as `swapper_pg_dir`. We may want to reduce that restriction.)
[...]

PR #587, opened 2026-01-07
libdrgn: Add set_linux_kernel_custom()

sdimitro:
## Context

There are cases where a `drgn` user wants to signify that the program being debugged is a Linux Kernel target. Currently there's no way of doing that. Linux kernel targets are only created by `set_kernel()` and `set_core_dump()` so if your target is not a live system nor a proper ELF dump/kdump you can't do anything about it so you're stuck (address translation doesn't work even if you pass a valid platform+vmcoreinfo_data and vmlinux is not loaded at the right offset).

The API introduced here provides a way to set the kernel target flag in drgn accordingly.

## References To Past Discussions

https://github.com/osandov/drgn/pull/246

https://github.com/osandov/drgn/issues/463

(also search for discussions on Matrix)

## Alternatives That I Considered

(1) Making the Program Flags mutable per the referenced PR above - this seems to go against the immutable creation idea.
(2) If anyone passes the `vmcoreinfo` in the `drgn.Program()` constructor then just assume that this is a linux kernel target and set the `is_linux_kernel` flag - I hesitated on this one because I'm afraid of breaking any existing scripts/behavior
(3) Pass an `is_linux_kernel=True` flag in the `drgn.Program()` as discussed in issue 463 - I didn't implement this because it seemed like Omar had some concerns.

### DIFF
--- a/_drgn.pyi
+++ b/_drgn.pyi
@@ -816,6 +816,24 @@ class Program:
         """
         ...
 
+    def set_linux_kernel_custom(self, vmcoreinfo: Union[str, bytes]) -> None:
+        """
+        Set the program to a custom Linux kernel target.
+
+        This enables debugging a Linux kernel via a custom memory transport
+        (e.g., RDMA, TCP/IP, or VMM introspection). It sets up page table
+        walking for virtual address translation. It does not load any
+        debugging symbols; see :meth:`load_default_debug_info()`.
+
+        Physical memory segments must be registered via
+        :meth:`add_memory_segment()` with ``physical=True`` before reading
+        memory. Platform must be provided when creating the :class:`Program`.
+
+        :param vmcoreinfo: Raw vmcoreinfo data. If vmcoreinfo was already set
+            when creating the :class:`Program`, this is ignored.
+        """
+        ...
+
     def set_pid(self, pid: int) -> None:
         """
         Set the program to a running process.
--- a/libdrgn/program.c
+++ b/libdrgn/program.c
@@ -733,6 +733,65 @@ drgn_program_set_kernel(struct drgn_program *prog)
 }
 
 LIBDRGN_PUBLIC struct drgn_error *
+drgn_program_set_linux_kernel_custom(struct drgn_program *prog,
+				     const char *vmcoreinfo,
+				     size_t vmcoreinfo_size)
+{
+	struct drgn_error *err;
+
+	if (!prog->has_platform) {
+		return drgn_error_create(DRGN_ERROR_INVALID_ARGUMENT,
+			"platform must be set before calling set_linux_kernel_custom()");
+	}
+
+	if (prog->flags & DRGN_PROGRAM_IS_LINUX_KERNEL)
+		return NULL;
+
+	// Parse vmcoreinfo if not already set via Program constructor
+	bool had_vmcoreinfo = prog->vmcoreinfo.raw != NULL;
+	if (!had_vmcoreinfo) {
+		err = drgn_program_parse_vmcoreinfo(prog, vmcoreinfo,
+						    vmcoreinfo_size);
+		if (err)
+			return err;
+	}
+
+	/*
+	 * Register a virtual memory reader that uses page table walking.
+	 * This translates virtual addresses to physical using swapper_pg_dir
+	 * from vmcoreinfo, then reads physical memory from user-registered
+	 * segments.
+	 */
+	if (prog->platform.arch->linux_kernel_pgtable_iterator_next) {
+		err = drgn_program_add_memory_segment(prog, 0, UINT64_MAX,
+						      read_memory_via_pgtable,
+						      prog, false);
+		if (err)
+			goto out_vmcoreinfo;
+	}
+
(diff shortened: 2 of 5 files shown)


libdrgn/program.c
```diff
@@ -733,6 +733,50 @@ drgn_program_set_kernel(struct drgn_program *prog)
...
+LIBDRGN_PUBLIC struct drgn_error *
+drgn_program_set_remote_kernel(struct drgn_program *prog)
+{
+	struct drgn_error *err;
+
+	if (!prog->vmcoreinfo.swapper_pg_dir) {
+		return drgn_error_create(DRGN_ERROR_INVALID_ARGUMENT,
+			"vmcoreinfo with swapper_pg_dir must be set before "
+			"calling set_remote_kernel()");
+	}
```
osandov: Let's keep `drgn_program_parse_vmcoreinfo()` as the source of truth for what we consider a valid vmcoreinfo and just check `if (!prog->vmcoreinfo.raw)` here.

libdrgn/program.c
```diff
@@ -733,6 +733,50 @@ drgn_program_set_kernel(struct drgn_program *prog)
 	return drgn_program_set_core_dump(prog, "/proc/kcore");
 }
 
+LIBDRGN_PUBLIC struct drgn_error *
+drgn_program_set_remote_kernel(struct drgn_program *prog)
```
osandov: Since this can be used for different use cases, I'd suggest the name `drgn_program_set_linux_kernel_custom()` or something along those lines (and rewording the documentation to emphasize the custom aspect rather than the remote aspect).

_drgn.pyi
```diff
@@ -748,6 +748,22 @@ class Program:
...
+
+        This enables debugging a remote kernel via a custom memory transport
+        (e.g., RDMA, TCP/IP, or VMM introspection). It sets up page table
+        walking for virtual address translation. It does not load any
+        debugging symbols; see :meth:`load_default_debug_info()`.
+
+        Physical memory segments must be registered via
+        :meth:`add_memory_segment()` with ``physical=True`` before reading
+        memory. Platform and vmcoreinfo must be provided when creating the
+        :class:`Program`.
```
osandov: `drgn.Program(vmcoreinfo=...)` is a wart, so I'd rather not make this new API require you to use it. Instead, I'd prefer for this method to take a mandatory `vmcoreinfo` argument and use that.

That would also make it easier to define third-party functions for other remote targets (e.g., `set_kernel_openocd()`) that take a pre-created `drgn.Program` and use `set_linux_kernel_custom()` under the hood.

Note that passing `vmcoreinfo` to `drgn.Program()` is intended for overriding the usual method of finding vmcoreinfo, so this method should ignore the passed vmcoreinfo if it was already set via `drgn.Program`. I.e., `if (!prog->vmcoreinfo.raw) ... drgn_program_parse_vmcoreinfo()`.

libdrgn/program.c
```diff
@@ -733,6 +733,50 @@ drgn_program_set_kernel(struct drgn_program *prog)
...
+			"calling set_remote_kernel()");
+	}
+
+	if (!prog->has_platform) {
+		return drgn_error_create(DRGN_ERROR_INVALID_ARGUMENT,
+			"platform must be set before calling set_remote_kernel()");
+	}
+
+	if (prog->flags & DRGN_PROGRAM_IS_LINUX_KERNEL)
+		return NULL;
```
osandov: I would like for this to check that the program hasn't already been initialized instead of silently ignoring this:

```suggestion
	err = drgn_program_check_initialized(prog);
	if (err)
		return err;
```

But that checks whether any memory segments have already been added, so it needs some tweaks to allow you to still register physical memory segments first. Your version is for now, and I can tweak it afterwards.

libdrgn/program.c
```diff
@@ -733,6 +733,50 @@ drgn_program_set_kernel(struct drgn_program *prog)
...
+	}
+
+	prog->flags |= DRGN_PROGRAM_IS_LINUX_KERNEL;
+
+	err = drgn_program_finish_set_kernel(prog);
+	if (err) {
+		prog->flags &= ~DRGN_PROGRAM_IS_LINUX_KERNEL;
+		return err;
+	}
+
```
osandov: This also needs to run plugins at this point:
```suggestion

	drgn_call_plugins_prog("drgn_prog_set", prog);
```

libdrgn/program.c
```diff
@@ -733,6 +733,50 @@ drgn_program_set_kernel(struct drgn_program *prog)
...
+		if (err)
+			return err;
+	}
+
+	prog->flags |= DRGN_PROGRAM_IS_LINUX_KERNEL;
+
+	err = drgn_program_finish_set_kernel(prog);
+	if (err) {
+		prog->flags &= ~DRGN_PROGRAM_IS_LINUX_KERNEL;
+		return err;
```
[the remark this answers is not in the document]
sdimitro: Would you like me to add some TODO comment about this so we don't lose context? I can also take a stab at it if you have any initial ideas

PR #589, opened 2026-01-09
drgn.helpers.linux.net: add netdev_ipv4_addrs

pkannoju:
This helper will be used to obtain the list of IPv4 addresses associated with a network device.

### DIFF
--- a/drgn/helpers/linux/net.py
+++ b/drgn/helpers/linux/net.py
@@ -9,8 +9,9 @@ The ``drgn.helpers.linux.net`` module provides helpers for working with the
 Linux kernel networking subsystem.
 """
 
+import ipaddress
 import operator
-from typing import Iterator, Optional, Union
+from typing import Iterator, List, Optional, Union
 
 from drgn import NULL, IntegerLike, Object, Program, Type, cast, container_of, sizeof
 from drgn.helpers.common.prog import (
@@ -37,6 +38,7 @@ __all__ = (
     "sk_nulls_for_each",
     "skb_shinfo",
     "is_pp_page",
+    "netdev_ipv4_addrs",
 )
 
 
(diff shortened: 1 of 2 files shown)


drgn/helpers/linux/net.py
```diff
@@ -325,3 +316,20 @@ def is_pp_page(page: Object) -> bool:
     # Between that and Linux kernel commit c07aea3ef4d4 ("mm: add a signature
     # in struct page") (in v5.14), there is no way to identify page_pool pages.
     raise NotImplementedError("page_pool pages cannot be identified before Linux 5.14")
+
+
+def netdev_ipv4_addr_list(dev: Object) -> List[ipaddress.IPv4Address]:
```
osandov: If you don't mind me bikeshedding the name, how about just `netdev_ipv4_addrs()`?

drgn/helpers/linux/net.py
```diff
@@ -325,3 +316,20 @@ def is_pp_page(page: Object) -> bool:
...
+    Get the list of IPV4 addresses associated with a network device.
+
+    :param dev: ``struct net_device *``
+    """
+    ips = []
+    prog = dev.prog_
+    if dev.ip_ptr:
+        ifa = dev.ip_ptr.ifa_list
+        while ifa:
+            addr_bytes = prog.read(ifa.ifa_address.address_of_(), 4)
```
osandov: Since you don't need the address as an `Object` here, you can just use `address_` (which is an `int`) instead of `address_of_()`:
```suggestion
            addr_bytes = prog.read(ifa.ifa_address.address_, 4)
```

pkannoju (commented):
Didn't "git pull" before commit and hence "netdev_name" was not present in my branch. Hence the conflict. Have resolved the conflict by accepting the both the changes.

drgn/helpers/linux/net.py
```diff
@@ -325,3 +327,20 @@ def is_pp_page(page: Object) -> bool:
...
+    Get the list of IPV4 addresses associated with a network device.
+
+    :param dev: ``struct net_device *``
+    """
+    ips = []
+    prog = dev.prog_
+    if dev.ip_ptr:
+        ifa = dev.ip_ptr.ifa_list
+        while ifa:
+            addr_bytes = prog.read(ifa.ifa_address.address_(), 4)
```
osandov: `address_` is an attribute, not a method. mypy might complain about it being None, which isn't possible in this scenario, so you can add a `type: ignore` annotation. (I'm not sure if arg-type is the right error code. If not, just copy the error code that mypy gives instead.)

```suggestion
            addr_bytes = prog.read(
                ifa.ifa_address.address_,  # type: ignore[arg-type]  # address can't be None.
                4,
            )
```

osandov (changes requested):
Thanks! A couple of final comments. The merge commit seems to be causing conflicts, so try getting rid of it and rebasing on `main` instead.

osandov:
Thank you!

pkannoju:
Thank you very much @osandov .

PR #590, opened 2026-01-13
drgn.helpers.linux.net: add netdev_ipv6_addrs

pkannoju:
This helper will be used to obtain the list of IPv6 addresses associated with a network device.


osandov (approved):
Thanks again!

PR #594, opened 2026-02-03
libdrgn: enable PCRE2 for Oracle Linux 8

brenns10:
PCRE2_MATCH_INVALID_UTF was added in PCRE 10.34, and Oracle Linux 8 has 10.32. This flag seems to be an enhancement but not required functionality. Don't use it when drgn is built against an old version, so we can still enable PCRE2 there.

---

While OL8 & friends are Python 3.6, they have a 3.12 interpreter available in the appstream repositories, so it's good to keep things buildable there too.

I did try another way of doing this, if you prefer, with this in `libdrgn/configure.ac`:
```
AC_MSG_CHECKING([for PCRE2_MATCH_INVALID_UTF])
AC_COMPILE_IFELSE(
	[AC_LANG_PROGRAM(
		[[
			#define PCRE2_CODE_UNIT_WIDTH 8
			#include <pcre2.h>
		]],
		[[uint32_t test = PCRE2_MATCH_INVALID_UTF;]]
	)],
	[
		AC_MSG_RESULT([yes])
		AC_DEFINE([HAVE_PCRE2_MATCH_INVALID_UTF], [1], [Define if PCRE2_MATCH_INVALID_UTF exists])
	],
	[
		AC_MSG_RESULT([no])
	]
)
```
But then I saw there were defines for the PCRE2 major/minor versions!

I tested this on Oracle Linux 8 rather quickly - prior to the patch, build fails, but with the patch, it succeeds, and the unit tests pass. I didn't go through to test specific unicode regexes, but I can do that tomorrow if you have concerns. I just wanted to post this so I could unstick our OL8 CI builds :)

### DIFF
--- a/tests/test_search_memory.py
+++ b/tests/test_search_memory.py
@@ -904,6 +904,7 @@ class TestSearchMemoryRegex(TestCase):
             [(0x3FFFFFFE, b"foo")],
         )
 
+    @unittest.skipUnless(drgn._with_pcre2_utf, "PCRE2 does not support UTF-8")
     def test_str_valid_utf8(self):
         # 'ñ' is \xc3\xb1 in UTF-8. '±' is \xc2\xb1 in UTF-8. A Unicode search
         # for the former shouldn't match the \xb1 byte in the latter.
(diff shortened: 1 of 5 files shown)


tests/test_search_memory.py
```diff
@@ -920,6 +922,7 @@ def test_invalid_utf8(self):
             list(prog.search_memory_regex(r"[a-z]+")), [(0x1002, "abcdef")]
         )
 
+    @unittest.skipUnless(drgn._with_pcre2_utf, "PCRE2 does not support UTF-8")
```
osandov: This test should just use a bytes pattern instead.

PR #604, opened 2026-02-18
Fix linux_helper_direct_mapping_offset on x86_64 non-root

brenns10:
Playing with the search command on my local development machine, I got "FaultError: address not mapped: 0". Digging in, it came from linux_kernel_direct_mapping_x86_64, because the kallsyms lookup had returned zero. This was because I was running as non-root, having opened kcore via the sudo helper, and my kernel returns 0's in /proc/kallsyms for non-root users.

There are two issues here. First, linux_kernel_direct_mapping_x86_64() was only written for the live case, and now that it's used for all kernel programs, it needs to check whether /proc/kallsyms is a valid way to lookup symbols. Second, the symbol lookup needs to fail when it reads a zero and is running as non-root. Some kernel configurations may allow root kallsyms, so we catch the error on the assumption that the symbol address will not be 0.

Fixes: d9f441a1 ("libdrgn: use arch-specific hook for direct map offset")

### DIFF
--- a/libdrgn/linux_kernel.c
+++ b/libdrgn/linux_kernel.c
@@ -101,6 +101,17 @@ invalid:
 	}
 	free(line);
 	fclose(file);
+
+	// Depending on sysctl kptr_restrict and perf_event_paranoid, only a
+	// user with CAP_SYSLOG can access the kallsyms addresses. We're not
+	// guaranteed to be running as root even in the live case. Given that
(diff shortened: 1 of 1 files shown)


osandov (changes requested):

osandov (approved):
Assuming tests pass, this looks good to go.

PR #607, opened 2026-02-20
libdrgn: riscv: Add support for stack undwinding

thecharlesjenkins:
Introduce riscv register definitions and add the helper functions to support stack unwinding.

### DIFF
--- a/libdrgn/arch_riscv.h
+++ b/libdrgn/arch_riscv.h
@@ -8,12 +8,237 @@
 #include <elf.h>
 #include <string.h>
 
+#include "cfi.h"
+#include "error.h"
 #include "platform.h" // IWYU pragma: associated
+#include "program.h"
+#include "register_state.h"
 
 /*
  * The ABI specification can be found at:
  * https://github.com/riscv-non-isa/riscv-elf-psabi-doc
  */
+
+static const struct drgn_cfi_row default_dwarf_cfi_row_riscv = DRGN_CFI_ROW(
+	// The psABI defines the CFA as the value of sp in the calling frame.
+	[DRGN_REGISTER_NUMBER(x2)] = { DRGN_CFI_RULE_CFA_PLUS_OFFSET },
+	// Callee-saved registers default to DW_CFA_same_value. This isn't
+	// explicitly documented in the psABI, but it seems to be the consensus.
+	DRGN_CFI_SAME_VALUE_INIT(DRGN_REGISTER_NUMBER(x8)),
+	DRGN_CFI_SAME_VALUE_INIT(DRGN_REGISTER_NUMBER(x9)),
+	DRGN_CFI_SAME_VALUE_INIT(DRGN_REGISTER_NUMBER(x18)),
+	DRGN_CFI_SAME_VALUE_INIT(DRGN_REGISTER_NUMBER(x19)),
+	DRGN_CFI_SAME_VALUE_INIT(DRGN_REGISTER_NUMBER(x20)),
+	DRGN_CFI_SAME_VALUE_INIT(DRGN_REGISTER_NUMBER(x21)),
+	DRGN_CFI_SAME_VALUE_INIT(DRGN_REGISTER_NUMBER(x22)),
+	DRGN_CFI_SAME_VALUE_INIT(DRGN_REGISTER_NUMBER(x23)),
+	DRGN_CFI_SAME_VALUE_INIT(DRGN_REGISTER_NUMBER(x24)),
+	DRGN_CFI_SAME_VALUE_INIT(DRGN_REGISTER_NUMBER(x25)),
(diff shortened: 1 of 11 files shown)


libdrgn/arch_riscv.h
```diff
@@ -0,0 +1,341 @@
...
+	err = drgn_program_read_memory(prog, frame, fp.value, 2 * reg_size,
+				       false);
+
+	if (err) {
+		if (err->code == DRGN_ERROR_FAULT) {
+			drgn_error_destroy(err);
+			err = &drgn_stop;
+		}
+		return err;
+	}
```
osandov, line 99: Slightly terser way of writing this:

```suggestion
	if (drgn_error_catch(&err, DRGN_ERROR_FAULT))
		return &drgn_stop;
	else if (err)
		return err;
```

(`drgn_error_catch()` is somewhat new, so some older code in drgn that you may have been referring to doesn't use it.)

libdrgn/arch_riscv.h
```diff
@@ -0,0 +1,341 @@
...
+	uint16_t inst_packet;
+	err = drgn_program_read_memory(prog, &inst_packet, ra.value - 2, sizeof(inst_packet),
+				 false);
+	if (err) {
+		if (err->code == DRGN_ERROR_FAULT) {
+			drgn_error_destroy(err);
+			err = &drgn_stop;
+		}
+		return err;
+	}
```
osandov, line 161: Same comment as above, you can use `drgn_error_catch()`.

libdrgn/arch_riscv.h
```diff
@@ -0,0 +1,341 @@
...
+{
+	struct drgn_error *err;
+
+	// The ABI does not require the return address to be in x1, but it generally
+	// is by convention.
+	struct optional_uint64 ra = drgn_register_state_get_u64(prog, regs, x1);
+	if (!ra.has_value)
+		return &drgn_stop;
+
+	struct drgn_register_state *tmp = drgn_register_state_dup(regs);
```
osandov, line 146: This allocation is leaked if the memory read below fails. The easiest fix is to move the allocation to after the read.

libdrgn/arch_riscv.h
```diff
@@ -0,0 +1,341 @@
...
+	}
+
+	struct drgn_register_state *regs =
+		drgn_register_state_create(pc, true);
+	if (!regs)
+		return &drgn_enomem;
+
+	drgn_register_state_set_from_buffer(regs, pc, buf);
+	drgn_register_state_set_from_buffer(regs, x2,
+					    (const char *)buf + 2 * reg_size);
```
osandov, line 62: Is there a reason that we don't get `x1`, `x3`, and `x4`, too?

libdrgn/arch_riscv.h
```diff
@@ -0,0 +1,341 @@
...
+	if (err) {
+		if (err->code == DRGN_ERROR_FAULT) {
+			drgn_error_destroy(err);
+			err = &drgn_stop;
+		}
+		return err;
+	}
+
+	// Instruction is 4 bytes if opcode ends in 0b11 and 2 bytes otherwise
+	uint32_t inst_size = (inst_packet & 0b11) == 0b11 ? 4 : 2;
```
osandov, line 164: I had to read up on RISC-V variable length instructions, so bear with me here. If I understand correctly, the least significant bits of the instruction indicate whether it is 16 or 32 bits, right? If that's the case, then on a little-endian target, won't this mistakenly check bits 16-17 of a 4-byte instruction? So maybe this should be going back 4 bytes instead? Assuming that 4-byte instructions must be 4-byte aligned and 2-byte instructions must be 2-byte aligned, I _think_ that's guaranteed to work even if we check 2 instructions back in the 2-byte case.

There's also the issue that `drgn_program_read_memory()` doesn't swap bytes if the host and target use different byte orders. `drgn_program_read_u32()` and co. byte swap automatically.

How is this? Forgive me if I got it totally wrong.

```suggestion
	uint32_t inst_packet;
	err = drgn_program_read_u32(prog, ra.value - 4, false, &inst_packet);
	if (drgn_error_catch(&err, DRGN_ERROR_FAULT))
		return &drgn_stop;
	else
		return err;

	// Instruction is 4 bytes if opcode ends in 0b11 and 2 bytes otherwise
	uint32_t inst_size = (inst_packet & 0b11) == 0b11 ? 4 : 2;
```

osandov (changes requested):
I've been waiting for this for a long time, thank you! I have a couple of minor style nits and a couple of minor bugs, but overall this looks awesome.

osandov:
FYI, I merged "libdrgn: Reflect Makefile changes in architecture docs". I'd be happy to finish up this PR on my end, but I'd appreciate your take on my `get_initial_registers_from_struct_riscv()` `x[134]` and variable length instruction questions.

PR #609, opened 2026-02-26
helpers/linux: add sysfs lookup and directory traversal helpers

### LINKED ISSUE #443: Add sysfs helpers

muskangoyal0606:
Add helper functions for resolving and traversing sysfs (kernfs) nodes from a drgn program.

Introduce sysfs_lookup_node() to resolve a sysfs path to a ``struct kernfs_node *`` starting from ``sysfs_root`` with proper handling of absolute and relative paths. Add sysfs_lookup_kobject() to return the corresponding ``struct kobject *`` for a path, returning the parent kobject for attribute files. Add sysfs_lookup() to return the containing ``struct device *`` when the kobject represents a device, and sysfs_listdir() to list child entries of a sysfs directory.

These helpers abstract common sysfs traversal logic and simplify inspection of sysfs objects from kernel memory.

Fixes: #443

Signed-off-by: Muskan Goyal <muskan@linux.ibm.com>

### DIFF
--- a/drgn/helpers/linux/sysfs.py
+++ b/drgn/helpers/linux/sysfs.py
@@ -1,0 +1,128 @@
+# (C) Copyright IBM Corp. 2026
+# SPDX-License-Identifier: LGPL-2.1-or-later
+
+"""
+Sysfs
+-----
+
+The ``drgn.helpers.linux.sysfs`` module provides helpers for working with
+the sysfs interface built on top of kernfs.
+"""
+
+from typing import List, Optional
+
+from drgn import NULL, Object, Program, cast, container_of
+from drgn.helpers.common.prog import takes_program_or_default
+from drgn.helpers.linux.kernfs import (
+    _kernfs_node_type,
+    kernfs_children,
+    kernfs_parent,
+    kernfs_walk,
+)
+
+__all__ = (
+    "sysfs_lookup_node",
+    "sysfs_lookup_kobject",
+    "sysfs_lookup",
+    "sysfs_listdir",
+)
+
+
+@takes_program_or_default
+def sysfs_lookup_node(prog: Program, path: str) -> Object:
+    """
+    Look up a ``struct kernfs_node *`` for a given sysfs path.
+
+    The path may be provided either relative to ``/sys`` (e.g.,
+    ``devices/system/cpu``) or as an absolute path beginning with
+    ``/sys``.
+
+    If the path is empty or refers to ``/sys`` itself, this returns
+    the sysfs root node.
+
+    :param path: Sysfs path (absolute or relative to ``/sys``)
+    :return: ``struct kernfs_node *`` or ``NULL``
+    """
+    root_kn = prog["sysfs_root_kn"]
+
+    if path.startswith("/"):
+        path = path.lstrip("/")
+        if path == "sys":
+            path = ""
+        elif path.startswith("sys/"):
+            path = path[4:].lstrip("/")
+        else:
+            return NULL(prog, "struct kernfs_node *")
+
+    return kernfs_walk(root_kn, path)
+
+
+@takes_program_or_default
+def sysfs_lookup_kobject(prog: Program, path: str) -> Optional[Object]:
+    """
+    Look up the ``struct kobject *`` corresponding to a sysfs path.
+
+    If the path refers to a directory, this returns ``kn->priv``.
+    If the path refers to a file (attribute), this returns the
+    parent kobject (i.e., the parent node’s ``priv``)
+
+    :param path: Sysfs path (absolute or relative to ``/sys``)
+    :return: ``struct kobject *`` or ``NULL``
+    """
+    kn = sysfs_lookup_node(prog, path)
+    if not kn:
+        return NULL(prog, "struct kobject *")
+
+    if _kernfs_node_type(kn, "KERNFS_DIR"):
+        return cast("struct kobject *", kn.priv)
+
+    parent = kernfs_parent(kn)
+    if not parent:
+        return NULL(prog, "struct kobject *")
+
+    return cast("struct kobject *", parent.priv)
+
+
+@takes_program_or_default
+def sysfs_lookup(prog: Program, path: str) -> Optional[Object]:
+    """
+    Look up the object represented by a sysfs path.
+
+    If the resolved kobject corresponds to a ``struct device``,
+    return the containing ``struct device *``. Otherwise, return
+    the ``struct kobject *``.
+
+    Note: This may return more specific types for other cases in the future.
+
+    :param path: Sysfs path (absolute or relative to ``/sys``)
+    :return: ``struct device *`` or ``struct kobject *`` or ``NULL``
+    """
+    kobj = sysfs_lookup_kobject(prog, path)
+    if not kobj:
+        return NULL(prog, "struct kobject *")
+
+    try:
+        device_ktype = prog["device_ktype"].address_of_()
+    except KeyError:
+        pass
+    else:
+        if kobj.ktype == device_ktype:
+            return container_of(kobj, "struct device", "kobj")
+
+    return kobj
+
+
+@takes_program_or_default
+def sysfs_listdir(prog: Program, path: str) -> List[bytes]:
+    """
+    List the children of a sysfs directory.
+
+    :param path: Sysfs directory path (absolute or relative to ``/sys``)
+    :return: List of child entry names
+    :raises ValueError: If path is not found or not a directory
+    """
+    kn = sysfs_lookup_node(prog, path)
+    if not kn:
+        raise ValueError(f"{path}: not found")
+
+    return [child.name.string_() for child in kernfs_children(kn)]
--- a/drgn/helpers/linux/sysfs.py
+++ b/drgn/helpers/linux/sysfs.py
@@ -88,19 +88,32 @@ def sysfs_lookup(prog: Program, path: str) -> Optional[Object]:
     """
     Look up the object represented by a sysfs path.
 
-    If the resolved kobject corresponds to a ``struct device``,
-    return the containing ``struct device *``. Otherwise, return
-    the ``struct kobject *``.
+    Resolve the ``struct kobject`` corresponding to a sysfs entry and,
+    when possible, return the kernel structure which contains that
+    kobject.
 
     Note: This may return more specific types for other cases in the future.
 
+    Depending on the kobject type, this helper may return:
+
+    - ``struct device *``
+    - ``struct class *``
+    - ``struct bus_type *``
+    - ``struct device_driver *``
+    - ``struct module_kobject *``
+
+    If the kobject type does not correspond to a known container
+    structure, the function returns the ``struct kobject *``.
+
     :param path: Sysfs path (absolute or relative to ``/sys``)
-    :return: ``struct device *`` or ``struct kobject *`` or ``NULL``
+    :return: Corresponding container structure pointer or
+        ``struct kobject *`` or ``NULL``
     """
     kobj = sysfs_lookup_kobject(prog, path)
     if not kobj:
         return NULL(prog, "struct kobject *")
 
+    # device
     try:
         device_ktype = prog["device_ktype"].address_of_()
     except KeyError:
@@ -108,6 +121,47 @@ def sysfs_lookup(prog: Program, path: str) -> Optional[Object]:
     else:
         if kobj.ktype == device_ktype:
             return container_of(kobj, "struct device", "kobj")
+
+    # module
+    try:
+        module_ktype = prog["module_ktype"].address_of_()
+    except KeyError:
+        pass
+    else:
+        if kobj.ktype == module_ktype:
+            return container_of(kobj, "struct module_kobject", "kobj")
+
+    # driver
+    try:
+        driver_ktype = prog["driver_ktype"].address_of_()
+    except KeyError:
+        pass
+    else:
+        if kobj.ktype == driver_ktype:
+            drv_priv = container_of(kobj, "struct driver_private", "kobj")
+            return drv_priv.driver
+
+    # class
+    try:
+        class_ktype = prog["class_ktype"].address_of_()
+    except KeyError:
+        pass
+    else:
+        if kobj.ktype == class_ktype:
+            kset = container_of(kobj, "struct kset", "kobj")
+            subsys = container_of(kset, "struct subsys_private", "subsys")
+            return getattr(subsys, "class")
+
+    # bus
+    try:
+        bus_ktype = prog["bus_ktype"].address_of_()
+    except KeyError:
+        pass
+    else:
+        if kobj.ktype == bus_ktype:
+            kset = container_of(kobj, "struct kset", "kobj")
+            subsys = container_of(kset, "struct subsys_private", "subsys")
+            return subsys.bus
 
     return kobj
 
--- a/tests/linux_kernel/helpers/test_sysfs.py
+++ b/tests/linux_kernel/helpers/test_sysfs.py
@@ -16,14 +16,15 @@ from tests.linux_kernel import LinuxKernelTestCase
 class TestSysfs(LinuxKernelTestCase):
     @classmethod
     def kernfs_node_from_fd(cls, fd):
+        import os
+
+        from drgn import cast
         from drgn.helpers.linux.fs import fget
         from drgn.helpers.linux.pid import find_task
-        from drgn import cast
-        import os
 
         file = fget(find_task(cls.prog, os.getpid()), fd)
         return cast("struct kernfs_node *", file.f_inode.i_private)
-    
+
     def test_sysfs_lookup_node(self):
         fds = []
         try:
@@ -165,6 +166,60 @@ class TestSysfs(LinuxKernelTestCase):
                         dev = sysfs_lookup(self.prog, entry.path[5:])
                         self.assertTrue(dev)
                         self.assertEqual(dev.type_.type_name(), "struct device *")
+
+        # Module case
+        if "module_ktype" in self.prog:
+            path = "/sys/module"
+            if os.path.exists(path):
+                with os.scandir(path) as entries:
+                    entry = next((e for e in entries if e.is_dir()), None)
+                    if entry:
+                        mod = sysfs_lookup(self.prog, entry.path[5:])
+                        self.assertTrue(mod)
+                        self.assertEqual(
+                            mod.type_.type_name(), "struct module_kobject *"
+                        )
+
+        # Driver case
+        if "driver_ktype" in self.prog:
+            path = "/sys/bus"
+            if os.path.exists(path):
+                with os.scandir(path) as buses:
+                    bus = next((b for b in buses if b.is_dir()), None)
+                    if bus:
+                        drv_dir = os.path.join(bus.path, "drivers")
+                        if os.path.exists(drv_dir):
+                            with os.scandir(drv_dir) as drivers:
+                                drv = next((d for d in drivers if d.is_dir()), None)
+                                if drv:
+                                    driver = sysfs_lookup(self.prog, drv.path[5:])
+                                    self.assertTrue(driver)
+                                    self.assertEqual(
+                                        driver.type_.type_name(),
+                                        "struct device_driver *",
+                                    )
+
+        # Class case
+        if "class_ktype" in self.prog:
+            path = "/sys/class"
+            if os.path.exists(path):
+                with os.scandir(path) as entries:
+                    entry = next((e for e in entries if e.is_dir()), None)
+                    if entry:
+                        cls = sysfs_lookup(self.prog, entry.path[5:])
+                        self.assertTrue(cls)
+                        self.assertIn("struct class *", cls.type_.type_name())
+
+        # Bus case
+        if "bus_ktype" in self.prog:
+            path = "/sys/bus"
+            if os.path.exists(path):
+                with os.scandir(path) as entries:
+                    entry = next((e for e in entries if e.is_dir()), None)
+                    if entry:
+                        bus = sysfs_lookup(self.prog, entry.path[5:])
+                        self.assertTrue(bus)
+                        self.assertIn("struct bus_type *", bus.type_.type_name())
 
     def test_sysfs_listdir(self):
         expected = os.listdir(b"/sys/kernel")
--- a/tests/linux_kernel/helpers/test_sysfs.py
+++ b/tests/linux_kernel/helpers/test_sysfs.py
@@ -1,0 +1,183 @@
+# (C) Copyright IBM Corp. 2026
+# SPDX-License-Identifier: LGPL-2.1-or-later
+
+import os
+
+from drgn import NULL
+from drgn.helpers.linux.sysfs import (
+    sysfs_listdir,
+    sysfs_lookup,
+    sysfs_lookup_kobject,
+    sysfs_lookup_node,
+)
+from tests.linux_kernel import LinuxKernelTestCase
+
+
+class TestSysfs(LinuxKernelTestCase):
+    @classmethod
+    def kernfs_node_from_fd(cls, fd):
+        from drgn.helpers.linux.fs import fget
+        from drgn.helpers.linux.pid import find_task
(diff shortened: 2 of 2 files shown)


muskangoyal0606:
Hi, just a quick follow-up on my PRs. Looking forward to your feedback, thanks!

drgn/helpers/linux/kernfs.py
```diff
@@ -150,3 +154,119 @@ def kernfs_children(kn: Object) -> Optional[Iterator[Object]]:
     return rbtree_inorder_for_each_entry(
         "struct kernfs_node", kn.dir.children.address_of_(), "rb"
     )
+
+
+def sysfs_lookup_node(prog: Program, path: str) -> Optional[Object]:
```
osandov, line 159: Use `@takes_program_or_default` (from `drgn.helpers.common.prog`) so that it is optional to pass `prog`.

```suggestion
@takes_program_or_default
def sysfs_lookup_node(prog: Program, path: str) -> Optional[Object]:
```

drgn/helpers/linux/kernfs.py
```diff
@@ -150,3 +154,119 @@ def kernfs_children(kn: Object) -> Optional[Iterator[Object]]:
...
+    Look up a ``struct kernfs_node *`` for a given sysfs path.
+
+    The path may be provided either relative to ``/sys`` (e.g.,
+    ``devices/system/cpu``) or as an absolute path beginning with
+    ``/sys``. Leading slashes are ignored.
+
+    If the path is empty or refers to ``/sys`` itself, this returns
+    the sysfs root node.
+
+    :param prog: ``struct drgn_program *``
```
osandov, line 170: Once you add `@takes_program_or_default`, documentation for `prog` will be added automatically.

```suggestion
```

drgn/helpers/linux/kernfs.py
```diff
@@ -150,3 +154,119 @@ def kernfs_children(kn: Object) -> Optional[Iterator[Object]]:
...
     )
+
+
+def sysfs_lookup_node(prog: Program, path: str) -> Optional[Object]:
+    """
+    Look up a ``struct kernfs_node *`` for a given sysfs path.
+
+    The path may be provided either relative to ``/sys`` (e.g.,
+    ``devices/system/cpu``) or as an absolute path beginning with
+    ``/sys``. Leading slashes are ignored.
```
osandov, line 165: I like the idea of accepting relative and absolute paths. I'm not sure about ignoring leading slashes entirely, though. Could we make this a bit stricter so that it only accepts either a relative path or a path beginning with "/sys"? Maybe something like:

```python3
if path.startswith("/"):
    path = path.lstrip("/")
    if path == "sys":
        path = ""
    elif path.startswith("sys/"):
        path = path[4:].lstrip("/")
    else:
        return NULL(prog, "struct kernfs_node *")
```

drgn/helpers/linux/kernfs.py
```diff
@@ -150,3 +154,119 @@ def kernfs_children(kn: Object) -> Optional[Iterator[Object]]:
...
+
+    :param prog: ``struct drgn_program *``
+    :param path: Sysfs path (absolute or relative to ``/sys``)
+    :return: ``struct kernfs_node *`` or ``NULL``
+    :raises LookupError: If ``sysfs_root_kn`` symbol is not present
+    """
+    try:
+        root_kn = prog["sysfs_root_kn"]
+    except KeyError as e:
+        raise LookupError("sysfs_root_kn symbol not found") from e
```
osandov, line 178: There's no need for special handling here, just let the lookup fail.
```suggestion
    """
    root_kn = prog["sysfs_root_kn"]
```

drgn/helpers/linux/kernfs.py
```diff
@@ -150,3 +154,119 @@ def kernfs_children(kn: Object) -> Optional[Iterator[Object]]:
...
+
+    path = path.strip()
```
osandov, line 180: Although sysfs doesn't typically have whitespace, whitespace is generally meaningful in paths so I wouldn't strip it.
```suggestion
```

drgn/helpers/linux/kernfs.py
```diff
@@ -150,3 +154,119 @@ def kernfs_children(kn: Object) -> Optional[Iterator[Object]]:
...
+
+    if not path:
+        return root_kn
```
osandov, line 183: `kernfs_lookup()` already handles the empty string, so no need for a special case.
```suggestion
```

drgn/helpers/linux/kernfs.py
```diff
@@ -150,3 +154,119 @@ def kernfs_children(kn: Object) -> Optional[Iterator[Object]]:
...
+    path = path.lstrip("/")
+
+    if path == "sys":
+        return root_kn
+
+    if path.startswith("sys/"):
+        path = path[4:]
+
+    if not path:
+        return root_kn
```
osandov, line 194: Same here.

```suggestion
```

drgn/helpers/linux/kernfs.py
```diff
@@ -150,3 +154,119 @@ def kernfs_children(kn: Object) -> Optional[Iterator[Object]]:
...
+
+    return kernfs_walk(root_kn, path)
+
+
+def sysfs_lookup_kobject(prog: Program, path: str) -> Optional[Object]:
```
osandov, line 199: `@takes_program_or_default` and no explicit `:param prog:` here, too.

drgn/helpers/linux/kernfs.py
```diff
@@ -150,3 +154,119 @@ def kernfs_children(kn: Object) -> Optional[Iterator[Object]]:
...
+    """
+    Look up the ``struct kobject *`` corresponding to a sysfs path.
+
+    If the path refers to a directory, this returns ``kn->priv``.
+    If the path refers to a file (attribute), this returns the
+    parent kobject (i.e., ``kn->__parent->priv``).
+
+    :param prog: ``struct drgn_program *``
+    :param path: Sysfs path relative to ``/sys``
```
osandov, line 208: Should this be `Sysfs path (absolute or relative to ``/sys``)` like for `sysfs_lookup_node()`?

drgn/helpers/linux/kernfs.py
```diff
@@ -150,3 +154,119 @@ def kernfs_children(kn: Object) -> Optional[Iterator[Object]]:
...
+    :return: ``struct kobject *`` or ``NULL``
+    """
+    kn = sysfs_lookup_node(prog, path)
+    if not kn:
+        return None
```
osandov, line 213: It is more idiomatic in drgn to return a `NULL` object for failed lookups. I.e.,
```suggestion
        return NULL(prog, "struct kobject *")
```

drgn/helpers/linux/kernfs.py
```diff
@@ -150,3 +154,119 @@ def kernfs_children(kn: Object) -> Optional[Iterator[Object]]:
...
+        return cast("struct kobject *", kn.priv)
+
+    parent = kn.__parent
+    if not parent:
+        return None
+
+    return cast("struct kobject *", parent.priv)
+
+
+def sysfs_lookup(prog: Program, path: str) -> Optional[Object]:
```
osandov, line 225: Same comments about `takes_program_or_default`, no `:param prog`, `:param path` can be absolute.

drgn/helpers/linux/kernfs.py
```diff
@@ -150,3 +154,119 @@ def kernfs_children(kn: Object) -> Optional[Iterator[Object]]:
...
+    """
+    Look up the object represented by a sysfs path.
+
+    If the resolved kobject corresponds to a ``struct device``,
+    return the containing ``struct device *``. Otherwise, return
+    the ``struct kobject *``.
```
osandov, line 231: Could you please add a note that this may return more specific types for other cases in the future?

drgn/helpers/linux/kernfs.py
```diff
@@ -150,3 +154,119 @@ def kernfs_children(kn: Object) -> Optional[Iterator[Object]]:
...
+    except KeyError:
+        return kobj
+
+    if kobj.ktype == device_ktype:
+        return container_of(kobj, "struct device", "kobj")
+
+    return kobj
+
+
+def sysfs_listdir(prog: Program, path: str) -> List[str]:
```
osandov, line 252: Same comments about `takes_program_or_default`, no `:param prog`, `:param path` can be absolute.

drgn/helpers/linux/kernfs.py
```diff
@@ -150,3 +154,119 @@ def kernfs_children(kn: Object) -> Optional[Iterator[Object]]:
...
+    """
+    List the children of a sysfs directory.
+
+    :param prog: ``struct drgn_program *``
+    :param path: Sysfs directory path relative to ``/sys``
+    :return: ``List[str]`` of child entry names
```
osandov, line 258: The signature already says `List[str]`, so this can just be:
```suggestion
    :return: List of child entry names
```

drgn/helpers/linux/kernfs.py
```diff
@@ -150,3 +154,119 @@ def kernfs_children(kn: Object) -> Optional[Iterator[Object]]:
...
+        raise ValueError(f"{path}: not found")
+
+    if not _kernfs_node_type(kn, "KERNFS_DIR"):
+        raise ValueError(f"{path}: not a directory")
+
+    children_iter = kernfs_children(kn)
+    if children_iter is None:
+        return []
+
+    return [child.name.string_().decode() for child in children_iter]
```
osandov, line 272: A few things here:

* `kernfs_children()` already checks that the path is a directory.
* This made me realize that `kernfs_children()`'s type annotation was wrong. I just fixed it: https://github.com/osandov/drgn/commit/40c59f5b2edaa3ead58b9d3d3d42c3219cf2d3a8. You'll have to rebase on that.
* We don't typically decode filenames to strings in case the data is corrupted.

So this can all be simplified to

```suggestion
    return [child.name.string_() for child in kernfs_children(kn)]
```

drgn/helpers/linux/kernfs.py
```diff
@@ -226,25 +226,73 @@ def sysfs_lookup(prog: Program, path: str) -> Optional[Object]:
...
+        pass
 
-    if kobj.ktype == device_ktype:
-        return container_of(kobj, "struct device", "kobj")
+    try:
+        module_ktype = prog["module_ktype"].address_of_()
+        if kobj.ktype == module_ktype:
+            return container_of(kobj, "struct module_kobject", "kobj")
+    except KeyError:
+        pass
```
osandov, line 269: For all of these checks, put the check in an `else:` and keep only the lookup in the `try:` so that it doesn't accidentally catch an unrelated exception.
```suggestion
    except KeyError:
        pass
    else:
            if kobj.ktype == module_ktype:
                return container_of(kobj, "struct module_kobject", "kobj")
```

osandov (changes requested):
This is awesome, thank you.

Please move all of the new sysfs helpers to their own `drgn/helpers/linux/sysfs.py` module. I've also left some comments throughout, but there's nothing major.

muskangoyal0606:
@osandov I’ve incorporated the suggested changes and pushed the updates. Most things look good, but I’m seeing an issue with `sysfs_lookup_node()`.

Based on your earlier suggestion to make the absolute path handling stricter, I updated the logic accordingly. With the current handling, paths like `"/kernel"` or `"/devices/..."` are getting rejected and returning `NULL`. But after stripping the leading `/`, these are actually valid inputs for `kernfs_walk()` (e.g., `"kernel"`), so we end up rejecting valid cases a bit too early.

Since `sysfs_lookup_node()` is used by the other helpers, this is also causing some cascading failures.

I tried handling it like this:

```
if path.startswith("/"):
    path = path.lstrip("/")

if path == "sys":
    path = ""
elif path.startswith("sys/"):
    path = path[4:]
```

This still handles `/sys/...` correctly while allowing cases like `"/kernel"` to work.

Does this approach look reasonable? If yes, I can update the PR accordingly.

osandov:
My intention was that only `/sys/kernel/...` and `kernel/...` would be valid. If you really want `/kernel/...` to be valid, I wouldn't be strongly opposed, but it seems a little cleaner semantically to only accept absolute paths starting with `/sys` or paths relative to `/sys`.

muskangoyal0606:
@osandov I’ve addressed the previous feedback and updated the helpers along with the corresponding tests. Would appreciate your review.

drgn/helpers/linux/sysfs.py
```diff
@@ -0,0 +1,132 @@
...
+__all__ = (
+    "sysfs_lookup_node",
+    "sysfs_lookup_kobject",
+    "sysfs_lookup",
+    "sysfs_listdir",
+)
+
+
+@takes_program_or_default
+def sysfs_lookup_node(prog: Program, path: str) -> Optional[Object]:
```
osandov: As far as I can tell, this will never return `None`, right?
```suggestion
def sysfs_lookup_node(prog: Program, path: str) -> Object:
```

drgn/helpers/linux/sysfs.py
```diff
@@ -0,0 +1,132 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
osandov: I didn't write this, so this should copyright should be for your employer 😄

tests/linux_kernel/helpers/test_kernfs.py
```diff
@@ -142,3 +148,189 @@ def test_kernfs_children(self):
...
+                os.close(fd)
+
+    def test_sysfs_lookup_kobject(self):
+        fds = []
+        try:
+            for path in ("/sys", "/sys/kernel"):
+                fds.append(os.open(path, os.O_RDONLY))
+
+            kns = [self.kernfs_node_from_fd(fd) for fd in fds]
+            kobj_kernel = cast("struct kobject *", kns[1].priv)
```
osandov: This is the same cast as inside of `sysfs_lookup_kobject()`, so this test is a little circular. How about:
```suggestion
            kernel_kobj = self.prog["kernel_kobj"]
```
And then you can get rid of all of the fd stuff in this test case.

tests/linux_kernel/helpers/test_kernfs.py
```diff
@@ -142,3 +148,189 @@ def test_kernfs_children(self):
...
+                os.close(fd)
+
+    def test_sysfs_lookup(self):
+        fds = []
+        try:
+            for path in ("/sys/kernel",):
+                fds.append(os.open(path, os.O_RDONLY))
+
+            kns = [self.kernfs_node_from_fd(fd) for fd in fds]
+            kobj_kernel = cast("struct kobject *", kns[0].priv)
```
osandov: Similar thing to `test_sysfs_lookup_kobject` here, use the global `kernel_kobj` instead.

tests/linux_kernel/helpers/test_kernfs.py
```diff
@@ -142,3 +148,189 @@ def test_kernfs_children(self):
...
+                    with os.scandir(path) as entries:
+                        # Pick a directory under /sys/block (e.g., "sda", "loop0"),
+                        # which corresponds to a block device, and verify that
+                        # sysfs_lookup() returns a non-NULL object for it.
+                        entry = next((e for e in entries if e.is_dir()), None)
+                        if entry:
+                            dev = sysfs_lookup(self.prog, entry.path[5:])
+                            self.assertNotEqual(
+                                dev, NULL(self.prog, "struct kobject *")
+                            )
```
osandov: How about testing that it returns a non-NULL `struct device *`?
```suggestion
                            self.assertEqual(dev.type_.type_name(), "struct device *")
                            self.assertTrue(dev)
```

drgn/helpers/linux/sysfs.py
```diff
@@ -88,19 +88,30 @@ def sysfs_lookup(prog: Program, path: str) -> Optional[Object]:
...
     Look up the object represented by a sysfs path.
 
-    If the resolved kobject corresponds to a ``struct device``,
-    return the containing ``struct device *``. Otherwise, return
-    the ``struct kobject *``.
+    Resolve the ``struct kobject`` corresponding to a sysfs entry and,
+    when possible, return the kernel structure which contains that
+    kobject.
 
-    Note: This may return more specific types for other cases in the future.
```
osandov: Keep the note about more types in the future, that's still relevant.

tests/linux_kernel/helpers/test_kernfs.py
```diff
@@ -308,6 +311,68 @@ def test_sysfs_lookup(self):
...
+                        bus = next((b for b in buses if b.is_dir()), None)
+                        if bus:
+                            drv_dir = os.path.join(bus.path, "drivers")
+                            if os.path.exists(drv_dir):
+                                with os.scandir(drv_dir) as drivers:
+                                    drv = next((d for d in drivers if d.is_dir()), None)
+                                    if drv:
+                                        driver = sysfs_lookup(self.prog, drv.path[5:])
+                                        self.assertNotEqual(
+                                            driver, NULL(self.prog, "struct kobject *")
```
osandov: You can do similar type checks to the one I mentioned for `struct device *` for all of these.

tests/linux_kernel/helpers/test_kernfs.py
```diff
@@ -199,9 +199,17 @@ def test_sysfs_lookup_kobject(self):
...
 
-            self.assertEqual(sysfs_lookup_kobject(self.prog, ""), NULL(self.prog, "struct kobject *"))
-            self.assertEqual(sysfs_lookup_kobject(self.prog, "/sys"), NULL(self.prog, "struct kobject *"))
-            self.assertEqual(sysfs_lookup_kobject(self.prog, "/sys/"), NULL(self.prog, "struct kobject *"))
+            self.assertEqual(
+                sysfs_lookup_kobject(self.prog, ""), NULL(self.prog, "struct kobject *")
+            )
+            self.assertEqual(
+                sysfs_lookup_kobject(self.prog, "/sys"),
+                NULL(self.prog, "struct kobject *"),
```
osandov: The commit that introduced these should have them formatted correctly so that you don't need to fix it in a later commit.

tests/linux_kernel/helpers/test_kernfs.py
```diff
@@ -272,11 +280,6 @@ def test_sysfs_lookup(self):
                 sysfs_lookup(self.prog, "/sys/kernel/vmcoreinfo"), kobj_kernel
             )
 
-            self.assertEqual(
-                sysfs_lookup(self.prog, "kernel"),
-                sysfs_lookup(self.prog, "kernel/vmcoreinfo"),
-            )
```
osandov: Why was this removed?

drgn/helpers/linux/sysfs.py
```diff
@@ -0,0 +1,184 @@
...
+    """
+    kn = sysfs_lookup_node(prog, path)
+    if not kn:
+        raise ValueError(f"{path}: not found")
+
+    children = kernfs_children(kn)
+    if not children:
+        return []
+
+    return [child.name.string_() for child in children]
```
osandov: Rebase on `main` to pick up https://github.com/osandov/drgn/commit/40c59f5b2edaa3ead58b9d3d3d42c3219cf2d3a8, then this can be simplified to:
```suggestion
    return [child.name.string_() for child in kernfs_children(kn)]
```

tests/linux_kernel/helpers/test_kernfs.py
```diff
@@ -145,3 +148,254 @@ def test_kernfs_children(self):
...
+                            )
+
+        finally:
+            for fd in fds:
+                os.close(fd)
+
+    def test_sysfs_listdir(self):
+        fd = os.open("/sys/kernel", os.O_RDONLY)
+        try:
+            expected = [name.encode() for name in os.listdir("/sys/kernel")]
```
osandov: I think if you pass in `bytes` you get `bytes` out.
```suggestion
            expected = os.listdir(b"/sys/kernel")
```

tests/linux_kernel/helpers/test_kernfs.py
```diff
@@ -145,3 +148,254 @@ def test_kernfs_children(self):
...
+
+            self.assertCountEqual(sysfs_listdir(self.prog, "kernel"), expected)
+            self.assertCountEqual(sysfs_listdir(self.prog, "/sys/kernel"), expected)
+
+        finally:
+            os.close(fd)
+
+        with self.assertRaises(ValueError) as context:
+            sysfs_listdir(self.prog, "sys/kernel")
+        self.assertEqual(str(context.exception), "sys/kernel: not found")
```
osandov: You can use `assertRaisesRegex` to simplify these.
```suggestion
        self.assertRaisesRegex(ValueError, "not found", sysfs_listdir, self.prog, "sys/kernel")
```

osandov (changes requested):
Thanks, just minor things now.

muskangoyal0606:
@osandov Addressed the previous feedback and updated the helpers and tests accordingly. Could you please take another look

osandov (commented):
Looks great, thanks. One last thing I missed, the new test cases should also be in their own test_sysfs.py file. I can fix that up before merging unless you would like to do it.

osandov (approved):
Thank you. I'll merge it as soon as tests pass.

osandov:
Test failure is unrelated.

PR #611, opened 2026-03-11
tests: fix kernfs_root test by using sysfs_root_kn

muskangoyal0606:
test_kernfs_root() compares the result of kernfs_root() with the kernel symbol sysfs_root. However, sysfs_root does not reliably resolve to a struct kernfs_root in all kernels.

In many kernels, the symbol sysfs_root refers to a struct device used by the driver core rather than the kernfs root backing the sysfs filesystem. When drgn resolves this symbol, it may therefore return a struct device *, which leads to a type mismatch when compared against the struct kernfs_root * returned by kernfs_root().

This results in failures such as:

    AssertionError: Object(prog, 'struct kernfs_root *', ...)
    != Object(prog, 'struct device *', ...)

The sysfs filesystem is implemented on top of kernfs, and its root node is exposed through the symbol sysfs_root_kn, which is a struct kernfs_node *. The corresponding kernfs_root can be obtained unambiguously via:

    kernfs_root(sysfs_root_kn)

Fix the test by resolving the expected root from sysfs_root_kn instead of directly comparing against sysfs_root. This avoids the ambiguity and ensures that both sides of the comparison refer to the same struct kernfs_root.

### DIFF
--- a/tests/linux_kernel/helpers/test_kernfs.py
+++ b/tests/linux_kernel/helpers/test_kernfs.py
@@ -36,13 +36,16 @@ class TestKernfs(LinuxKernelTestCase):
             )
 
     def test_kernfs_root(self):
+        # Since Linux kernel commit 6138f3451516, "sysfs_root" is ambiguous.
+        # Ensure we resolve the symbol from sysfs (mount.c).
+        expected_root = self.prog.object("sysfs_root", filename="mount.c")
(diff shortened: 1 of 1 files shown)


muskangoyal0606:
Thanks for the suggestion! Updated as suggested and added a comment.

PR #612, opened 2026-03-18
linux/helpers: add sysfs_kobject_path() helper

muskangoyal0606:
Add a helper to reconstruct the sysfs path for a given struct kobject.

sysfs_kobject_path() walks the underlying kernfs_node hierarchy starting
from kobj.sd and builds the full /sys/... path by traversing parent nodes
up to the sysfs root.
The helper supports both struct kobject * and raw pointer inputs, and
handles edge cases such as NULL pointers, invalid addresses, and objects
not present in sysfs.

Also add tests in tests/linux_kernel to validate common sysfs paths,
address-based inputs, and error handling.
Tested on a running kernel and across the sysfs tree to ensure correctness.

### DIFF
--- a/drgn/helpers/linux/sysfs.py
+++ b/drgn/helpers/linux/sysfs.py
@@ -17,6 +17,7 @@ from drgn.helpers.linux.kernfs import (
     _kernfs_node_type,
     kernfs_children,
     kernfs_parent,
+    kernfs_path,
     kernfs_walk,
 )
 
@@ -25,6 +26,7 @@ __all__ = (
     "sysfs_lookup_kobject",
     "sysfs_lookup",
     "sysfs_listdir",
+    "sysfs_kobject_path",
 )
 
 
@@ -180,3 +182,19 @@ def sysfs_listdir(prog: Program, path: str) -> List[bytes]:
         raise ValueError(f"{path}: not found")
 
     return [child.name.string_() for child in kernfs_children(kn)]
+
+
+def sysfs_kobject_path(kobj: Object) -> bytes:
+    """
+    Return the sysfs path corresponding to a ``struct kobject *``.
+
+    :param kobj: ``struct kobject *``
+    :return: Full sysfs path as bytes
+    """
+    kn = kobj.sd
+    path = kernfs_path(kn)
+
+    if path == b"/":
+        return b"/sys"
+
+    return b"/sys" + path
--- a/tests/linux_kernel/helpers/test_sysfs.py
+++ b/tests/linux_kernel/helpers/test_sysfs.py
@@ -5,6 +5,7 @@ import os
 
 from drgn import NULL
 from drgn.helpers.linux.sysfs import (
+    sysfs_kobject_path,
     sysfs_listdir,
     sysfs_lookup,
     sysfs_lookup_kobject,
@@ -235,4 +236,12 @@ class TestSysfs(LinuxKernelTestCase):
         )
         self.assertRaisesRegex(
             ValueError, "not a directory", sysfs_listdir, self.prog, "kernel/vmcoreinfo"
+        )
+
+    def test_sysfs_kobject_path(self):
+        kernel_kobj = self.prog["kernel_kobj"]
+
+        self.assertEqual(
+            sysfs_kobject_path(kernel_kobj),
+            b"/sys/kernel",
         )


drgn/helpers/linux/kernfs.py
```diff
@@ -150,3 +151,67 @@ def kernfs_children(kn: Object) -> Optional[Iterator[Object]]:
...
+    try:
+        kn = kobj.sd
+        if not kn:
+            raise ValueError("kobject not present in sysfs")
+    except FaultError:
+        raise ValueError("invalid kobject pointer")
+
+    path_components = []
+
+    try:
```
osandov, line 203: Is there a reason that `kernfs_path()` can't be used to implement this?

drgn/helpers/linux/sysfs.py
```diff
@@ -180,3 +182,32 @@ def sysfs_listdir(prog: Program, path: str) -> List[bytes]:
         raise ValueError(f"{path}: not found")
 
     return [child.name.string_() for child in kernfs_children(kn)]
+
+
+@takes_program_or_default
+def sysfs_kobject_path(prog: Program, kobj: Object) -> str:
```
osandov: Since this one takes an object, it doesn't need to take `prog`; the helper can get it via `kobj.prog_` if needed (although it doesn't seem like it's needed in this case).

```suggestion
def sysfs_kobject_path(kobj: Object) -> str:
```

drgn/helpers/linux/sysfs.py
```diff
@@ -180,3 +182,32 @@ def sysfs_listdir(prog: Program, path: str) -> List[bytes]:
...
+    to reconstruct the full ``/sys/...`` path.
+
+    :param kobj: ``struct kobject *``
+    :return: Full sysfs path as a string
+    :raises ValueError: If the kobject is NULL or not present in sysfs
+    """
+
+    if not kobj:
+        raise ValueError("NULL kobject")
+
```
osandov: No need for this check, it will fail naturally with a `FaultError` (and don't worry about testing that).
```suggestion
```

drgn/helpers/linux/sysfs.py
```diff
@@ -180,3 +182,32 @@ def sysfs_listdir(prog: Program, path: str) -> List[bytes]:
...
+    if not kn:
+        raise ValueError("kobject not present in sysfs")
+
+    path = kernfs_path(kn)
+
+    decoded = path.decode()
+    if decoded == "/":
+        return "/sys"
+
+    return "/sys" + decoded
```
osandov: Our helpers generally return paths as bytes, so this should use bytes and not decode.

drgn/helpers/linux/sysfs.py
```diff
@@ -180,3 +182,32 @@ def sysfs_listdir(prog: Program, path: str) -> List[bytes]:
...
+    :return: Full sysfs path as a string
+    :raises ValueError: If the kobject is NULL or not present in sysfs
+    """
+
+    if not kobj:
+        raise ValueError("NULL kobject")
+
+    kn = kobj.sd
+    if not kn:
+        raise ValueError("kobject not present in sysfs")
```
osandov: Out of curiosity, what cases does this happen in?

tests/linux_kernel/helpers/test_sysfs.py
```diff
@@ -236,3 +237,32 @@ def test_sysfs_listdir(self):
...
+            "/sys/kernel",
+        )
+
+        self.assertEqual(
+            sysfs_kobject_path(self.prog, sysfs_lookup_kobject(self.prog, "kernel")),
+            sysfs_kobject_path(
+                self.prog, sysfs_lookup_kobject(self.prog, "kernel/vmcoreinfo")
+            ),
+        )
+
```
osandov: These aren't adding any more coverage since they're calling `sysfs_lookup_kobject()` on the same `kernel_kobj` that we just tested, so I'd just remove them.

```suggestion
```

tests/linux_kernel/helpers/test_sysfs.py
```diff
@@ -236,3 +237,32 @@ def test_sysfs_listdir(self):
...
+        self.assertRaisesRegex(
+            ValueError,
+            "NULL kobject",
+            sysfs_kobject_path,
+            self.prog,
+            NULL(self.prog, "struct kobject *"),
+        )
```
osandov: As noted above, just remove this along with the explicit check for null.
```suggestion
```

osandov (changes requested):

drgn/helpers/linux/sysfs.py
muskangoyal0606: This would only occur for invalid or uninitialized kobjects where sd is NULL. Since helpers assume valid inputs, I’ll remove this and let it fail naturally.

drgn/helpers/linux/sysfs.py
```diff
@@ -180,3 +182,22 @@ def sysfs_listdir(prog: Program, path: str) -> List[bytes]:
...
 
     return [child.name.string_() for child in kernfs_children(kn)]
+
+
+def sysfs_kobject_path(kobj: Object) -> bytes:
+    """
+    Return the sysfs path corresponding to a ``struct kobject *``.
+
+    This uses the underlying kernfs node associated with the kobject
+    and calls ``kernfs_path()`` to construct the full ``/sys/...`` path.
```
osandov: This sentence is just providing implementation details that the caller doesn't really need to know, so it can be deleted. Other than that, this looks ready to land.

osandov (approved):
Looks great, thank you!

osandov:
Heads up, I realized after landing this that I'd prefer the name `kobject_path()`, so I changed it: https://github.com/osandov/drgn/commit/d3105bbee292a3217cf897f4ba5797ac3d176428.

PR #615, opened 2026-04-07
contrib: add qemu_mem.py for faster memory access

brenns10:
The qemu support commits nerd sniped me! I was playing targeting some vmtest VMs and thought drgn could probably read memory a bit faster if it could reliably read from `/proc/PID/mem` of QEMU. This obviously won't work in all situations, and the current assumptions are quite naive, but I figured it's a start. I wrote it as a plugin (but it can also be run via `execscript()`). That way you can use it pretty transparently -- for example, see below where I can use it on top of another contrib script. The goal is that it will fail quickly and in a reasonably orderly way so that you can still use the built-in QMP reading mechanism if this approach does not work.

Without:
```
$ /usr/bin/time python -m drgn -q --qemu build/vmtest/vm.sock --kernel-dir build/vmtest/x86_64/kernel-7.0.0-rc7-vmtest40.1default/ contrib/ps.py >/dev/null
1.74user 0.50system 0:20.19elapsed 11%CPU (0avgtext+0avgdata 229816maxresident)k
0inputs+0outputs (0major+23353minor)pagefaults 0swaps
```

With:
```
$ DRGN_PLUGINS=qemu=contrib/qemu_mem.py /usr/bin/time python -m drgn -q --qemu build/vmtest/vm.sock --kernel-dir build/vmtest/x86_64/kernel-7.0.0-rc7-vmtest40.1default/ contrib/ps.py >/dev/null
0.86user 0.10system 0:00.27elapsed 348%CPU (0avgtext+0avgdata 229788maxresident)k
0inputs+0outputs (0major+23356minor)pagefaults 0swaps
```

### DIFF
--- a/contrib/qemu_mem.py
+++ b/contrib/qemu_mem.py
@@ -1,0 +1,162 @@
+# Copyright (c) 2026, Oracle and/or its affiliates.
+# SPDX-License-Identifier: LGPL-2.1-or-later
+"""
+qemu_mem.py - speed up memory access for QEMU programs via /proc/pid/mem
+
+Parsing the QEMU HMP "xp" command is much slower than native reads for multiple
+reasons. While that's great to boostrap a debugging session, it may be possible
+to speed up the process. If QEMU is creating a single, anonymous memory region
+for guest memory, we may be able to create a faster reader via /proc/pid/mem.
+While this is not a common configuration for real-world virtualization, it's a
+good starting point and could probably be extended to more practical use cases.
+
+Use this in one of two ways:
+
+1. Once drgn has started: execscript("path/to/qemu_mem.py")
+2. As a drgn plugin: DRGN_PLUGINS=qemu:path/to/qemu_mem.py drgn ...
+"""
+import logging
+import os
+import socket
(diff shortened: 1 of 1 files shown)


osandov (changes requested):
This is missing the copyright notice and SPDX identifier. Other than that, I like this as a proof of concept despite the hacks and wouldn't be opposed to integrating it into libdrgn if you think it could be done reliably.

brenns10:
Sorry for the delay, I was out on my own short break. I added the copyright & license, thanks for the reminder.

Regarding integrating this kind of approach into libdrgn... I think that would be really exciting! And probably really useful for real-world use cases where kernel developers are debugging a local VM via something like virtme-ng. For production VMs though I think there's a lot of complexity surrounding memory (numa, hotplug, ballooning, hugetlb backing, confidential compute, probably others). I'd want to support what we could and fail gracefully for the rest. Sadly I think parsing QEMU arguments would be the best way to detect the memory configuration reliably. I'll think a bit more about it.

osandov (approved):

PR #621, opened 2026-05-23
s390x: Correct kdump address translation loop

### LINKED ISSUE #566: Address translation loops when doing s390x dump analysis
When using drgn with s390x kernel dumps like this:

```
drgn -c dump.0 -s vmlinux
```

the following warning appears:

```
warning: can't find kernel modules: couldn't read modules list: recursive address translation; page table may be missing from core dump: 0x39ad4bb0000
```

and then very few commands work. The reason is that the dump contains only physical memory, but `linux_kernel_pgtable_iterator_next_s390x()` starts with `swapper_pg_dir`, which is virtual. This works on live systems, because `/proc/kcore` contains virtual memory.

Looking at what crash does, for many kinds of addresses it uses heuristics (`s390x_vr_VTOP()`) and this way it sidesteps the necessity to do a page table walk for reading from `swapper_pg_dir`.

Also, lowcore contains the kernel ASCE at all times in its `kernel_asce` member, so this can be used as well instead of `swapper_pg_dir`.

theotherjimmy:
The kernel uses a virtual memory address for the `swapper_pg_dir` on s390x, and this is placed in an offset defined by the `kaslr_offset` and `kaslr_offset_phys`. Since we don't have virtal memory enabled when we're translating the virtual memory address, we have to use this offset instead of translation.

Resolves #566

Note that `DRGN_PREFER_PGTABLE_READER=1` is required to use this logic.

### DIFF
--- a/libdrgn/arch_s390x.c
+++ b/libdrgn/arch_s390x.c
@@ -395,6 +395,19 @@ linux_kernel_pgtable_iterator_next_s390x(struct drgn_program *prog,
 	const uint64_t va = it->it.virt_addr;
 	uint64_t table = _it->pgtable & ~UINT64_C(0xfff);
 	bool table_physical = false;
+	if (table == prog->vmcoreinfo.swapper_pg_dir) {
+		// Since Linux kernel commits 378e32aa8197 ("s390/vmcoreinfo:
+		// Store virtual memory layout") and c98d2ecae08f ("s390/mm:
+		// Uncouple physical vs virtual address spaces") (in v6.10), we
+		// can translate the kernel page table address to a physical
+		// address using VMCOREINFO. Before that, kernel virtual and
+		// physical addresses are equal.
+		if (prog->vmcoreinfo.have_kaslr_offset_phys) {
+			table -= prog->vmcoreinfo.kaslr_offset;
+			table += prog->vmcoreinfo.kaslr_offset_phys;
+		}
+		table_physical = true;
+	}
 	int level, length = 2048, offset = 0;
 	uint64_t entry;
 
@@ -404,7 +417,7 @@ linux_kernel_pgtable_iterator_next_s390x(struct drgn_program *prog,
 	 * the linux kernel does: read the first level entry, and deduct the
 	 * number of levels from the TT bits.
 	 */
-	struct drgn_error *err = drgn_program_read_u64(prog, table, false,
+	struct drgn_error *err = drgn_program_read_u64(prog, table, table_physical,
 						       &entry);
 	if (err)
 		return err;
--- a/libdrgn/program.h
+++ b/libdrgn/program.h
@@ -196,6 +196,11 @@ struct drgn_program {
 				 * layout randomization (KASLR) is enabled.
 				 */
 				uint64_t kaslr_offset;
+				/**
+				 * The offset from physical memory address 0 of
+				 * the kernel image on s390x.
+				 */
+				uint64_t kaslr_offset_phys;
 				/** Kernel page table. */
 				uint64_t swapper_pg_dir;
 				/**
@@ -233,6 +238,8 @@ struct drgn_program {
 				bool have_crashtime;
 				/** Whether `phys_base` was in the VMCOREINFO. */
 				bool have_phys_base;
+				/** Whether `kaslr_offset_phys` was in the VMCOREINFO. */
+				bool have_kaslr_offset_phys;
 				/** Length of build ID. */
 				unsigned int build_id_len;
 				/**
--- a/libdrgn/drgn_program_parse_vmcoreinfo.inc.strswitch
+++ b/libdrgn/drgn_program_parse_vmcoreinfo.inc.strswitch
@@ -105,6 +105,13 @@ struct drgn_error *drgn_program_parse_vmcoreinfo(struct drgn_program *prog,
 			if (err)
 				return err;
 			break;
+		@case "KERNELOFFPHYS"@
+			err = parse_vmcoreinfo_u64(value, newline, 16,
+						   &prog->vmcoreinfo.kaslr_offset_phys);
+			if (err)
+				return err;
+			prog->vmcoreinfo.have_kaslr_offset_phys = true;
+			break;
 		@case "SYMBOL(swapper_pg_dir)"@
 			err = parse_vmcoreinfo_u64(value, newline, 16,
 						   &prog->vmcoreinfo.swapper_pg_dir);


theotherjimmy:
I added the DCO. Sorry I missed that the first time.

libdrgn/arch_s390x.c
```diff
@@ -395,6 +395,13 @@ linux_kernel_pgtable_iterator_next_s390x(struct drgn_program *prog,
 	const uint64_t va = it->it.virt_addr;
 	uint64_t table = _it->pgtable & ~UINT64_C(0xfff);
 	bool table_physical = false;
+	if (table == prog->vmcoreinfo.swapper_pg_dir
+	    && prog->vmcoreinfo.kaslr_offset
+	    && prog->vmcoreinfo.kaslr_offset_phys) {
```
osandov: Is there a reason that this wouldn't be valid if kaslr_offset or kaslr_offset_phys happen to be zero, or if either is absent and defaulted to 0? Do we explicitly need to record whether the values were present in the vmcoreinfo (like we do with `have_phys_base`, e.g.)?

libdrgn/arch_s390x.c
```diff
@@ -395,6 +395,13 @@ linux_kernel_pgtable_iterator_next_s390x(struct drgn_program *prog,
...
+	    table -= prog->vmcoreinfo.kaslr_offset;
+	    table += prog->vmcoreinfo.kaslr_offset_phys;
+	    table_physical = true;
```
osandov: Style nit, this needs to be indented with 2 tabs rather than a tab and 4 spaces.

osandov (changes requested):
Awesome, thanks for this. A couple of minor comments below.

libdrgn/arch_s390x.c
theotherjimmy: On s390x, both offsets are part of lowcore, so they should both be present in a kdump. I'll remove the checks, we can always add them back if we hit a corner case where we need them.

libdrgn/arch_s390x.c
theotherjimmy: Go it in the next version.

theotherjimmy:
@osandov Thanks for the review. I corrected the PR. I think it's ready.

theotherjimmy:
I have another patch that correct `%crash ps -u -a`. Should I add that here or submit that as a separate PR?

osandov:
> I have another patch that correct `%crash ps -u -a`. Should I add that here or submit that as a separate PR?

New PR for that one, please.

theotherjimmy:
I'll create a new PR for that then. Is there something that this PR still needs to get a second round of review?

libdrgn/arch_s390x.c
```diff
@@ -395,6 +395,11 @@ linux_kernel_pgtable_iterator_next_s390x(struct drgn_program *prog,
 	const uint64_t va = it->it.virt_addr;
 	uint64_t table = _it->pgtable & ~UINT64_C(0xfff);
 	bool table_physical = false;
+	if (table == prog->vmcoreinfo.swapper_pg_dir) {
+		table -= prog->vmcoreinfo.kaslr_offset;
+		table += prog->vmcoreinfo.kaslr_offset_phys;
+		table_physical = true;
+	}
```
osandov, line 410: The case I'm not sure about is kernels before the virtual/physical split. In that case, won't this break if the KASLR offset is non-zero? Maybe we want something like this?

```suggestion
	if (table == prog->vmcoreinfo.swapper_pg_dir) {
		// Since Linux kernel commits 378e32aa8197 ("s390/vmcoreinfo:
		// Store virtual memory layout") and c98d2ecae08f ("s390/mm:
		// Uncouple physical vs virtual address spaces") (in v6.10), we
		// can translate the kernel page table address to a physical
		// address using VMCOREINFO. Before that, kernel virtual and
		// physical addresses are equal.
		if (prog->vmcoreinfo.have_kaslr_offset_phys) {
			table -= prog->vmcoreinfo.kaslr_offset;
			table += prog->vmcoreinfo.kaslr_offset_phys;
		}
		table_physical = true;
	}
```

libdrgn/program.h
```diff
@@ -196,6 +196,14 @@ struct drgn_program {
 				 * layout randomization (KASLR) is enabled.
 				 */
 				uint64_t kaslr_offset;
+				/**
+				 * The offset from physical memory address 0 of
+				 * the kernel image.
+				 *
+				 * This is non-zero if kernel address space
+				 * layout randomization (KASLR) is enabled.
```
osandov: Since this only exists for s390x at the moment, let's clarify:
```suggestion
				 * the kernel image on s390x.
```

libdrgn/arch_s390x.c
theotherjimmy: This, with the updates to `program.h` and the vmcoreinfo seems like a nice improvement. I'm adding it now.

theotherjimmy:
I tested locally. This works just fine with my specific kdump.

osandov (approved):
Looks good. I'll run some tests locally and land once those pass.

theotherjimmy:
What tests do you run locally (So I can do them too)?

theotherjimmy:
eln-aarch64 build failure looks spurious.

osandov:
The CI only runs tests on x86-64, so for other architectures I run them locally. The setup is a bit involved:

If you have qemu-s390x-static set up in binfmt_misc as well as qemu-system-s390x installed, then this should all work out of the box.

Create chroots (the x86-64 chroot is needed because it provides a compatible cross-compiler). This will probably take awhile.

```sh
python3 -m vmtest.rootfsbuild -a x86_64
python3 -m vmtest.rootfsbuild -a s390x --build-drgn
```

Create a GitHub authentication token. This is needed to download test kernels from GitHub. Without it, things may work, but you’ll probably be rate limited aggressively.

```sh
gh auth login
```

Then, you can use the main `vmtest` script to run tests. E.g., running the entire test suite on the latest kernel:

```sh
GITHUB_TOKEN=$(gh auth token) python3 -m vmtest -a s390x -k '*'
```

On a specific kernel:

```sh
GITHUB_TOKEN=$(gh auth token) python3 -m vmtest -a s390x -k '6.6.*'
```

All supported kernels, parallelized (takes a long time even on a big machine):

```sh
GITHUB_TOKEN=$(gh auth token) python3 -m vmtest -a s390x -k all -j
```

theotherjimmy:
> All supported kernels, parallelized (takes a long time even on a big machine):

I'll "cheat" and use s390x hardware for this step.

osandov:
> > All supported kernels, parallelized (takes a long time even on a big machine):
> 
> I'll "cheat" and use s390x hardware for this step.

I've never actually tried running those scripts on a non-x86-64 machine. In theory it should still work, but I wouldn't be surprised if something breaks. The first thing I can think of is that the test kernels are cross-compiled on x86-64, so the build tools they contain are x86-64 binaries.

osandov:
CI and my local tests passed, and enough of the packit builds have succeeded, so I'll go ahead and land this. Thank you!

PR #623, opened 2026-06-05
Translate pgd addresses other than swapper_pg_dir

theotherjimmy:
This corrects the `%crash ps -u -a` output, allowing reading of memory in non-`swapper_pg_dir` virtual memory spaces, such as user processes.

### DIFF
--- a/libdrgn/linux_kernel_helpers.c
+++ b/libdrgn/linux_kernel_helpers.c
@@ -4,6 +4,7 @@
 #include <inttypes.h>
 #include <stdio.h>
 
+#include "array.h"
 #include "drgn_internal.h"
 #include "error.h"
 #include "helpers.h"
@@ -14,22 +15,27 @@
 
 static void end_virtual_address_translation(struct drgn_program *prog)
 {
-	prog->in_address_translation = false;
+	prog->address_translation_depth--;
 }
 
 static struct drgn_error *
 begin_virtual_address_translation(struct drgn_program *prog, uint64_t pgtable,
-				  uint64_t virt_addr)
+				  uint64_t virt_addr, struct pgtable_iterator **ret)
 {
 	struct drgn_error *err;
-
-	if (prog->in_address_translation) {
-		return drgn_error_create_fault("recursive address translation; "
-					       "page table may be missing from core dump",
+	for (int i = 0; i < prog->address_translation_depth; i++) {
+		if (prog->pgtable_its[i]->pgtable == pgtable) {
+			return drgn_error_create_fault("recursive address translation; "
+						       "page table may be missing from core dump",
+						       virt_addr);
+		}
+	}
+	if (prog->address_translation_depth >= array_size(prog->pgtable_its)) {
+		return drgn_error_create_fault("address translation recursion limit hit",
 					       virt_addr);
 	}
-	prog->in_address_translation = true;
-	if (!prog->pgtable_it) {
+
+	if (!prog->pgtable_its[prog->address_translation_depth]) {
 		if (!(prog->flags & DRGN_PROGRAM_IS_LINUX_KERNEL)) {
 			err = drgn_error_create(DRGN_ERROR_UNSUPPORTED_OPERATION,
 						"virtual address translation is only available for the Linux kernel");
@@ -46,16 +52,21 @@ begin_virtual_address_translation(struct drgn_program *prog, uint64_t pgtable,
 						prog->platform.arch->name);
 			goto err;
 		}
-		err = prog->platform.arch->linux_kernel_pgtable_iterator_create(prog,
-										&prog->pgtable_it);
+		err = prog->platform.arch->linux_kernel_pgtable_iterator_create(
+			prog, &prog->pgtable_its[prog->address_translation_depth]
+		);
 		if (err) {
-			prog->pgtable_it = NULL;
+			prog->pgtable_its[prog->address_translation_depth] = NULL;
 			goto err;
 		}
 	}
-	prog->pgtable_it->pgtable = pgtable;
-	prog->pgtable_it->virt_addr = virt_addr;
-	prog->platform.arch->linux_kernel_pgtable_iterator_init(prog, prog->pgtable_it);
+	struct pgtable_iterator *it =
+		prog->pgtable_its[prog->address_translation_depth];
+	it->pgtable = pgtable;
+	it->virt_addr = virt_addr;
+	prog->platform.arch->linux_kernel_pgtable_iterator_init(prog, it);
+	*ret = it;
+	prog->address_translation_depth++;
 	return NULL;
 
 err:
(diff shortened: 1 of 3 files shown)


theotherjimmy:
@osandov This is the follow up I promised. I would like some feedback on how this one is done, as it's a bit more involved.

osandov:
Do you get a "recursive address translation" error today? If so, I think this same bug exists for other architectures, and we should solve it more generally. We could try supporting one level of recursion in `begin_virtual_address_translation()` as long as the pgtable address is different.

theotherjimmy:
After my prior fix, yes. I get the "recursive translation error" with `%crash ps -a -u` on current main.

theotherjimmy:
@osandov Is there something else needed before review?

osandov (changes requested):
This makes it so that pgtable is always passed to the page table iterator as a physical address unless it's `swapper_pg_dir`? Mixing the address spaces this way is a bit confusing and fragile. In fact, it breaks all of the other architectures that are still treating the address as virtual. However, the test suite didn't catch this because of a bug that I just fixed: https://github.com/osandov/drgn/commit/beca2ab04bddd6440e37c30ca0da05408399a8a6. Please rebase on the `main` branch and squash your commits.

It's also not always necessary to do this automatic translation (e.g., on live kernels, kernel virtual memory can be read directly).

I think the most flexible way to do this would be to cache an array of page table iterators that can be used for recursive calls. I.e., replacing `struct drgn_program::pgtable_it` and `struct drgn_program::in_address_translation` with an array of two iterators and a recursion counter:

```c
struct pgtable_iterator *pgtable_its[2];
size_t address_translation_depth;
```

Then do something like this:

```c
// begin_virtual_address_translation():
if (address_translation_depth >= ARRAY_SIZE(pgtable_its))
	return fault error about address translation depth being exceeded

if (any active pgtable_it matches the passed pgtable)
	return fault error about recursive address translation

if (!pgtable_its[address_translation_depth])
	create new cached iterator

pgtable_it = pgtable_its[address_translation_depth++];

use pgtable_it

// end_virtual_address_translation():
address_translation_depth--;
```

theotherjimmy:
Thanks for the thoughtful feedback. I've reworked the PR as you suggested and rebased onto current main.

libdrgn/linux_kernel_helpers.c
```diff
@@ -12,24 +12,32 @@
 #include "program.h"
 #include "util.h"
 
+#define ARRAY_SIZE(x) (sizeof(x) / sizeof((x)[0]))
```
osandov: Oops, I forgot that it's named `array_size()` in libdrgn. Just include `"array.h"` and use the lowercase version instead.

libdrgn/linux_kernel_helpers.c
```diff
@@ -12,24 +12,32 @@
...
 static void end_virtual_address_translation(struct drgn_program *prog)
 {
-	prog->in_address_translation = false;
+	prog->address_translation_depth--;
 }
 
 static struct drgn_error *
 begin_virtual_address_translation(struct drgn_program *prog, uint64_t pgtable,
-				  uint64_t virt_addr)
+				  uint64_t virt_addr, struct pgtable_iterator** it)
```
osandov: Style nit: asterisks on the other side, and name it `ret` to indicate that it's an out-parameter:

```suggestion
				  uint64_t virt_addr, struct pgtable_iterator **ret)
```

libdrgn/linux_kernel_helpers.c
```diff
@@ -12,24 +12,32 @@
...
-
-	if (prog->in_address_translation) {
-		return drgn_error_create_fault("recursive address translation; "
-					       "page table may be missing from core dump",
+	for (int i = 0; i < prog->address_translation_depth; i++) {
+		if (prog->pgtable_its[i]->pgtable == pgtable) {
+			return drgn_error_create_fault("recursive address translation; "
+						       "atempted to translate page directory address using itself; "
+						       "information on swapper_pg_dir may be missing from core dump",
+						       virt_addr);
```
osandov, line 30: I'd prefer to keep the original, broader error message since this condition can arise in other pathological cases.

libdrgn/linux_kernel_helpers.c
```diff
@@ -46,16 +54,21 @@ begin_virtual_address_translation(struct drgn_program *prog, uint64_t pgtable,
...
 	}
-	prog->pgtable_it->pgtable = pgtable;
-	prog->pgtable_it->virt_addr = virt_addr;
-	prog->platform.arch->linux_kernel_pgtable_iterator_init(prog, prog->pgtable_it);
+	prog->pgtable_its[prog->address_translation_depth]->pgtable = pgtable;
+	prog->pgtable_its[prog->address_translation_depth]->virt_addr = virt_addr;
+	prog->platform.arch->linux_kernel_pgtable_iterator_init(
+		prog, prog->pgtable_its[prog->address_translation_depth]
+	);
+	*it = prog->pgtable_its[prog->address_translation_depth];
```
osandov: Formatting-wise, I think this would look a little nicer:

```suggestion
	struct pgtable_iterator *it =
		prog->pgtable_its[prog->address_translation_depth];
	it->pgtable = pgtable;
	it->virt_addr = virt_addr;
	prog->platform.arch->linux_kernel_pgtable_iterator_init(prog, it);
	*ret = it;
```

libdrgn/program.c
```diff
@@ -36,6 +36,8 @@
 #include "util.h"
 #include "vector.h"
 
+#define ARRAY_SIZE(x) (sizeof(x) / sizeof((x)[0]))
```
osandov: Same thing re: "array.h"/`array_size()` here.

libdrgn/program.c
```diff
@@ -162,8 +164,12 @@ void drgn_program_deinit(struct drgn_program *prog)
 		drgn_thread_destroy(prog->crashed_thread);
 		drgn_thread_destroy(prog->main_thread);
 	}
-	if (prog->pgtable_it)
-		prog->platform.arch->linux_kernel_pgtable_iterator_destroy(prog->pgtable_it);
+	for (int i = 0; i < ARRAY_SIZE(prog->pgtable_its); i++) {
+		struct pgtable_iterator *it = prog->pgtable_its[i];
+		if (it) {
+			prog->platform.arch->linux_kernel_pgtable_iterator_destroy(it);
+		}
```
osandov: Style nit:
```suggestion
		if (it)
			prog->platform.arch->linux_kernel_pgtable_iterator_destroy(it);
```

osandov (changes requested):
Thanks, this looks great. Just a few final nitpicks and this should be ready to go.

theotherjimmy:
That should do it.

osandov (approved):
Thanks again!

PR #625, opened 2026-06-11
Add BTF type & object finder

brenns10:
I think the BTF code is getting close to ready, so I was hoping to request an API review if possible, to ensure that it's going in a direction you're happy with.

The API I have here is surprisingly small: I just add a `Module.load_btf()` method. This method can optionally take a buffer, but if the buffer is not provided, it will search for a `.BTF` section in the loaded or debug file and use that. It will read and index the types, so that the already-registered BTF type/object finder can use them.

For the DWARFless case, the current status of BTF in the kernel is still missing all global variables (except percpu). So I've included some helpers that allow you to provide a list of variable declarations to help drgn bootstrap its module API and find the BTF associated with each module. This is definitely not ideal, but it's a starting place. It's actually surprising how much you can do with just a few hard-coded global variable declarations. Ideally, I'll get global variables into the kernel BTF somehow upstream, and then this will be usable without any of the variable declaration shenanigans.

Since the API allows finding `.BTF` sections from an ELF file, one interesting result is that you can actually use this with userspace programs that were compiled with `gcc -gbtf`. The variable lookup works a bit differently than the Linux kernel due to a quirk in the BTF DATASEC generation. But type lookup is identical. I figured this would be a good way to help test the BTF support.

What I'd love a sanity check on is:
- The API on the Module class
- Whether the helpers are worthwhile to include
- Should the object finder be separated out into another PR?

I know there's still a lot of work needed in certain areas:
- I'll need to create a unit test suite similar to the DWARF suite. I hadn't started since I didn't want to build on a shifting API.
- Same difference on testing the kernel BTF support in the vm tests: i'll need to add a test suite once the API is set.
- I definitely need to do a second pass on code organization and structure for `btf_info.c`.

If you wanted to try it out, I've been doing this on my Fedora development machine:
1. DWARFless: test out loading types or percpu variables
```
$ python -m drgn -qk --no-default-symbols
drgn 0.2.0+19.g1d09fd07.dirty (using Python 3.14.5, elfutils 0.195, with debuginfod (dlopen), with json-c, with libkdumpfile, with lzma, with pcre2, with libbpf)
For help, type help(drgn).
>>> import drgn
>>> from drgn import FaultError, NULL, Object, alignof, cast, container_of, execscript, implicit_convert, offsetof, reinterpret, sizeof, search_memory, search_memory_regex, search_memory_u16, search_memory_u32, search_memory_u64, search_memory_word, source_location, stack_trace
>>> from drgn.helpers.common import *
>>> load_builtin_btf(prog)
>>> %py prog.type("struct task_struct") |head
struct task_struct {
	struct thread_info thread_info;
	unsigned int __state;
	unsigned int saved_state;
	void *stack;
	refcount_t usage;
	unsigned int flags;
	unsigned int ptrace;
	int on_cpu;
	struct __call_single_node wake_entry;
```
2. Userspace test (note that rodata is not included in the core dump, similar results happen with DWARF).
```
$ cat test.c
#include <stdio.h>
#include <stdlib.h>
struct my_struct { int foo; char *bar; };
struct my_struct value = {12, "blah"};
int foobar = 42;
const char ch = 'c';
const char *string = "string data";
int main(int argc, char **argv)
{
	abort();
	return 0;
}
$ gcc -gbtf -o test test.c
$ ./test
Aborted                    (core dumped) ./test
$ coredumpctl dump > test.core
$ python -m drgn -c test.core -s test
drgn 0.2.0+19.g1d09fd07.dirty (using Python 3.14.5, elfutils 0.195, with debuginfod (dlopen), with json-c, with libkdumpfile, with lzma, with pcre2, with libbpf)
warning: missing debugging symbols for /home/stephen/repos/drgn/test
critical: missing some debugging symbols; see https://drgn.readthedocs.io/en/latest/getting_debugging_symbols.html
For help, type help(drgn).
>>> import drgn
>>> from drgn import FaultError, NULL, Object, alignof, cast, container_of, execscript, implicit_convert, offsetof, reinterpret, sizeof, search_memory, search_memory_regex, search_memory_u16, search_memory_u32, search_memory_u64, search_memory_word, source_location, stack_trace
>>> from drgn.helpers.common import *
>>> prog.main_module().load_btf()
>>> prog["value"]
(struct my_struct){
	.foo = (int)12,
	.bar = (char *)__dso_handle+0x8 = 0x401180,
}
>>> prog["ch"]
warning: can't print value: memory not saved in core dump: 0x401185
Object(prog, 'const char', address=0x401185)
>>> prog["foobar"]
(int)42
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -56,7 +56,7 @@ jobs:
       - name: Install dependencies
         run: |
           sudo apt-get update -y
-          sudo apt-get install -y btrfs-progs check dwarves libelf-dev libdw-dev libjson-c-dev libkdumpfile-dev libpcre2-dev qemu-kvm zstd ${{ matrix.cc == 'clang' && 'libomp-$(clang --version | sed -rn "s/.*clang version ([0-9]+).*/\\1/p")-dev' || '' }}
+          sudo apt-get install -y btrfs-progs check dwarves libbpf-dev libelf-dev libdw-dev libjson-c-dev libkdumpfile-dev libpcre2-dev qemu-kvm zstd ${{ matrix.cc == 'clang' && 'libomp-$(clang --version | sed -rn "s/.*clang version ([0-9]+).*/\\1/p")-dev' || '' }}
           # pyroute2 0.9.1 dropped support for Python < 3.9.
           if [[ "${{ matrix.python-version }}" =~ ^3\.[678]$ ]]; then
               pyroute2_version="<0.9.1"
--- a/libdrgn/configure.ac
+++ b/libdrgn/configure.ac
@@ -158,6 +158,19 @@ AS_CASE(["x$with_json_c"],
 AM_CONDITIONAL([WITH_JSON_C], [test "x$with_json_c" = xyes])
 AM_COND_IF([WITH_JSON_C], [AC_DEFINE(WITH_JSON_C)])
 
+AC_ARG_WITH([libbpf],
+	    [AS_HELP_STRING([--with-libbpf],
+			    [build with support for loading BTF type information with libbpf
+			     @<:@default=auto@:>@])],
+			     [], [with_libbpf=auto])
+AS_CASE(["x$with_libbpf"],
+	[xyes], [PKG_CHECK_MODULES(bpf, [libbpf >= 1.0.0])],
+	[xauto], [PKG_CHECK_MODULES(bpf, [libbpf >= 1.0.0],
+				    [with_libbpf=yes],
+				    [with_libbpf=no])])
+AM_CONDITIONAL([WITH_LIBBPF], [test "x$with_libbpf" = xyes])
+AM_COND_IF([WITH_LIBBPF], [AC_DEFINE(WITH_LIBBPF)])
+
 dnl We need check for running tests, but we don't want to fail the build over
 dnl it. Instead, if it's not found, set variables so that only `make check`
 dnl fails.
--- a/libdrgn/python/module.c
+++ b/libdrgn/python/module.c
@@ -181,6 +181,22 @@ static PyObject *Module_try_file(Module *self, PyObject *args, PyObject *kwds)
 	Py_RETURN_NONE;
 }
 
+static PyObject *Module_load_btf(Module *self, PyObject *args, PyObject *kwds)
+{
+	static char *keywords[] = { "data", "main_module_base", NULL };
+	const char *btf_data = NULL;
+	size_t btf_size = 0;
+	int main_module_base = DRGN_TRISTATE_DEFAULT;
+	if (!PyArg_ParseTupleAndKeywords(args, kwds, "|$z#p:load_btf", keywords,
+					 &btf_data, &btf_size, &main_module_base))
+		return NULL;
+	struct drgn_error *err = drgn_module_load_btf(self->module, btf_data,
+						      btf_size, main_module_base);
+	if (err)
+		return set_drgn_error(err);
+	Py_RETURN_NONE;
+}
+
 static Program *Module_get_prog(Module *self, void *arg)
 {
 	Program *prog = Module_prog(self);
@@ -497,6 +513,8 @@ static PyMethodDef Module_methods[] = {
 	 drgn_Module_wanted_supplementary_debug_file_DOC},
 	{"try_file", (PyCFunction)Module_try_file,
 	 METH_VARARGS | METH_KEYWORDS, drgn_Module_try_file_DOC},
+	{"load_btf", (PyCFunction)Module_load_btf,
+	 METH_VARARGS | METH_KEYWORDS, drgn_Module_load_btf_DOC},
 	{},
 };
 
--- a/README.rst
+++ b/README.rst
@@ -162,13 +162,13 @@ First, install dependencies:
 
   .. code-block:: console
 
-      $ sudo dnf install autoconf automake check-devel elfutils-debuginfod-client-devel elfutils-devel gcc git json-c-devel libkdumpfile-devel libtool make pcre2-devel pkgconf python3 python3-devel python3-pip python3-setuptools xz-devel
+      $ sudo dnf install autoconf automake check-devel elfutils-debuginfod-client-devel elfutils-devel gcc git json-c-devel libbpf-devel libkdumpfile-devel libtool make pcre2-devel pkgconf python3 python3-devel python3-pip python3-setuptools xz-devel
 
 * RHEL/CentOS < 9, Oracle Linux
 
   .. code-block:: console
 
-      $ sudo dnf install autoconf automake check-devel elfutils-devel gcc git json-c-devel libtool make pcre2-devel pkgconf python3 python3-devel python3-pip python3-setuptools xz-devel
+      $ sudo dnf install autoconf automake check-devel elfutils-devel gcc git json-c-devel libbpf-devel libtool make pcre2-devel pkgconf python3 python3-devel python3-pip python3-setuptools xz-devel
 
   Optionally, install ``libkdumpfile-devel`` from EPEL on RHEL/CentOS >= 8 or
   install `libkdumpfile <https://github.com/ptesarik/libkdumpfile>`_ from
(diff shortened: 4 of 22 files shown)


libdrgn/python/drgnpy.h
```diff
@@ -549,5 +547,7 @@ PyObject *drgnpy_linux_helper_load_proc_kallsyms(PyObject *self, PyObject *args,
 						 PyObject *kwds);
 PyObject *drgnpy_linux_helper_load_builtin_kallsyms(PyObject *self, PyObject *args,
 						    PyObject *kwds);
+PyObject *drgnpy_linux_helper_load_btf(PyObject *self, PyObject *args,
+				       PyObject *kwds);
```
osandov: Seems like  this is stale?
```suggestion
```

.github/workflows/ci.yml
```diff
@@ -56,7 +56,7 @@ jobs:
       - name: Install dependencies
         run: |
           sudo apt-get update -y
-          sudo apt-get install -y btrfs-progs check dwarves libelf-dev libdw-dev libjson-c-dev libpcre2-dev qemu-kvm zstd ${{ matrix.cc == 'clang' && 'libomp-$(clang --version | sed -rn "s/.*clang version ([0-9]+).*/\\1/p")-dev' || '' }}
+          sudo apt-get install -y btrfs-progs check dwarves libelf-dev libdw-dev libjson-c-dev libpcre2-dev libbpf-dev qemu-kvm zstd ${{ matrix.cc == 'clang' && 'libomp-$(clang --version | sed -rn "s/.*clang version ([0-9]+).*/\\1/p")-dev' || '' }}
```
osandov: Nit: keep this in alphabetical order.

```suggestion
          sudo apt-get install -y btrfs-progs check dwarves libbpf-dev libelf-dev libdw-dev libjson-c-dev libpcre2-dev qemu-kvm zstd ${{ matrix.cc == 'clang' && 'libomp-$(clang --version | sed -rn "s/.*clang version ([0-9]+).*/\\1/p")-dev' || '' }}
```

libdrgn/python/module.c
```diff
@@ -181,6 +181,24 @@ static PyObject *Module_try_file(Module *self, PyObject *args, PyObject *kwds)
 	Py_RETURN_NONE;
 }
 
+static PyObject *Module_load_btf(Module *self, PyObject *args, PyObject *kwds)
+{
+	static char *keywords[] = {"data", "main_module_base"};
+	const char *btf_data = NULL;
+	size_t btf_size = 0;
+	int main_module_base = 0;
+	if (!PyArg_ParseTupleAndKeywords(args, kwds, "|$y#p:load_btf", keywords,
```
osandov: `y#` is not accurate for `Optional[bytes]` (it allows `str` and doesn't allow an explicit `None`).

libdrgn/python/module.c
```diff
@@ -181,6 +181,24 @@ static PyObject *Module_try_file(Module *self, PyObject *args, PyObject *kwds)
...
+					 &btf_data, &btf_size, &main_module_base))
+		return NULL;
+	struct drgn_error *err = drgn_module_load_btf(self->module, btf_data,
+						      btf_size, main_module_base);
+	if (err) {
+		set_drgn_error(err);
+		return NULL;
+	}
```
osandov: Style nit:
```suggestion
	if (err)
		return set_drgn_error(err);
```

libdrgn/configure.ac
```diff
@@ -158,6 +158,19 @@ AS_CASE(["x$with_json_c"],
...
+			    [build with support for loading BTF type information with libbpf
+			     @<:@default=auto@:>@])],
+			     [], [with_bpf=auto])
+AS_CASE(["x$with_bpf"],
+	[xyes], [PKG_CHECK_MODULES(bpf, [libbpf])],
+	[xauto], [PKG_CHECK_MODULES(bpf, [libbpf],
+				    [with_bpf=yes],
+				    [with_bpf=no])])
+AM_CONDITIONAL([WITH_BPF], [test "x$with_bpf" = xyes])
+AM_COND_IF([WITH_BPF], [AC_DEFINE(WITH_BPF)])
```
osandov: This is pretty inconsequential, but this PR is inconsistent with naming this `with_bpf` or `with_libbpf` throughout the various layers. Let's unify on one.

With the exception of libkdumpfile, which was the first one, I've been preferring leaving off the "lib", but "with bpf" is a big vague, so maybe "with libbpf" is better?

osandov (commented):
I think the main design choices are going to revolve around how automatic this should be and how it should interact with DWARF, which we'll discuss in the debugging tools meeting shortly.

Just dropping a few trivial nitpicks that I spotted as I was reading this in preparation.

libdrgn/configure.ac
brenns10: Yeah, sorry about that. I agree it should at least be internally consistent for all the "with libbpf" options/variables. And I also agree that "with libbpf" is clearer, since "with bpf" may imply some special-sauce BPF programs or something.

brenns10:
Notes from our discussion at the meeting yesterday:
- We're in agreement that there shouldn't be any automatic use of BTF, not even a CLI flag, yet. This is mainly just so we don't need to commit to maintaining any behavior, as the upstream BTF global variable situation will evolve.
- The BTF type finder should be registered after the DWARF finder -- it's registered all the time so it's best to keep it secondary.
- The Python-based helpers should probably move to `drgn.helpers.experimental.btf` for now, for similar maintenance reasons.
- The API structure (one `load_btf()` method with a few optional arguments) seemed good. You noted that the `main_module_base` argument should default to whatever is expected for the kernel since that's the common case.
- From my side - I know this needs a heck of a lot more testing, and there's some dead code and cleanup that needs to be handled still.
- I won't make any major changes here until you've finished your current look.

brenns10:
Hey @osandov, do you mind if I start cleaning up and incorporating some of the feedback into this PR this week?

osandov:
Yup, took another skim and I think it's all good structurally. I'll just clarify one point:

> You noted that the `main_module_base` argument should default to whatever is expected for the kernel since that's the common case.

If the common case differs between the kernel and userspace, then it's fine to have the default depend on `prog->flags & DRGN_PROGRAM_IS_LINUX_KERNEL`. In one of my dev branches, I even have tristate for expressing this in libdrgn, so you could even add that if you feel it's appropriate:

```c
enum drgn_tristate {
	DRGN_TRISTATE_DEFAULT = -1,
	DRGN_TRISTATE_FALSE,
	DRGN_TRISTATE_TRUE,
};
```

brenns10:
Ok, I've updated this and believe it is ready for review!

1. I've added tests - for BTF type & object finders, as well as Linux kernel tests that load kernel & module BTF and verify the same type & object finders.
2. vmtest is passing locally on all architectures & flavors, and [all Python versions in Github CI](https://github.com/brenns10/drgn/actions/runs/29933970471).
3. I've made all the changes we discussed in the last meeting, and I set a minimum version of 1.0.0 for libbpf.

One notable difference is that this new iteration has two different BTF object finders: `btf` and `btf_symbol` (and then there is the "C declaration object finder" from the Python helpers).

- The `btf` finder does things the way that BTF is designed: a `BTF_KIND_VAR` gives you the name and type of a variable. A `BTF_KIND_DATASEC` contains an entry that indicates the ELF section containing the variable, its size, and its offset from the start of the ELF section. So this finder returns only variables for which it has a VAR, a DATASEC, and for which it can identify the base address of the containing section.
  - It can be used for kernel percpu variables.
  - It can be used for userspace applications whose BTF is converted from DWARF via `pahole`.
- The `btf_symbol` finder does not rely on `BTF_KIND_DATASEC`. For any `BTF_KIND_VAR`, we look up the symbol of the same name, and then create an object using the symbol address. This is hacky, but it's quite effective: the CTF finder works exactly like this and has been fantastic for us.
  - It can be used for userspace applications whose BTF was directly generated via GCC. (GCC emits `btf_var_secinfo` entries with offset=0 for any non-BPF backend).
  - It could also be used as a backup implementation in case DATASEC base addresses cannot be determined (see below).
- The C declaration object finder does not even rely on a VAR or a DATASEC. You provide it a set of C declarations in string format. It looks up the variable name as a symbol, and the type name (using whatever the configured symbol/type finders are), and then squishes them together to create an object.
  - This is what can be used for all non-percpu kernel variables, until we can get the kernel to enable the pahole `global_var` feature to include all global variables.
  - It's not even necessarily BTF-specific; I think this finder may be a useful generic tool to have around.

By making the `btf` and `btf_symbol` finders separate, this makes it possible to be very flexible and support the BTF which GCC emits as of today, while also supporting BTF generated by pahole with proper DATASEC offsets.

Of course, since BTF variables rely on ELF section base addresses, we need a way to determine those. Thankfully this is already easily available for kernel modules via `section_addresses`. For vmlinux, for now we only care about percpu variables, but I have a heuristic which should work for many sections (I've tested it on kernels built with pahole's global_var feature). It's an open question about whether we want some way to provide section addresses for vmlinux. For userspace, we can just rely on the ELF sections on the loaded/debug files.

libdrgn/python/module.c
brenns10: I changed this to `z#` and changed the type annotation to `Union[None, bytes, str]`. While I don't want to allow `str`... there's no `Optional[bytes]` option and I'm not sure whether it's worth implementing a converter for it.

PR #627, opened 2026-06-24
RFC: Direct VMCOREINFO Retrieval for QMP Targets

muskangoyal0606:
This RFC explores an alternative approach for retrieving VMCOREINFO from
QMP targets and validates the approach on ppc64.

The series contains:

1. Direct VMCOREINFO retrieval through QMP memory accesses using
   vmcoreinfo_note and vmcoreinfo_size.

2. Additional ppc64 changes required to initialize drgn and perform
   address translation when using the RFC VMCOREINFO implementation.

The goal is to evaluate whether direct VMCOREINFO retrieval is a viable
alternative to the current dump-guest-memory() based approach for QMP
targets.

Tested on:
- ppc64 QEMU guest

Feedback is appreciated.

### DIFF
--- a/drgn/cli.py
+++ b/drgn/cli.py
@@ -445,6 +445,10 @@ def _load_debugging_symbols(prog: drgn.Program, args: argparse.Namespace) -> Non
             critical = False
         logger.log(logging.CRITICAL if critical else logging.WARNING, "%s", e)
 
+
+def _load_only_extra_debug_symbols(
+    prog: drgn.Program, args: argparse.Namespace
+) -> None:
     if args.extra_symbols:
         for extra_symbol_path in args.extra_symbols:
             extra_symbol_path = os.path.abspath(extra_symbol_path)
@@ -676,6 +680,8 @@ def _main() -> None:
             vmcoreinfo = f.read()
 
     prog = drgn.Program(platform=platform, vmcoreinfo=vmcoreinfo)
+    _load_only_extra_debug_symbols(prog, args)
+
     try:
         if args.core is not None:
             prog.set_core_dump(args.core)
--- a/libdrgn/arch_ppc64.c
+++ b/libdrgn/arch_ppc64.c
@@ -360,7 +360,7 @@ linux_kernel_pgtable_iterator_create_ppc64(struct drgn_program * prog,
 	drgn_error_destroy(err);
 
 	// Identify the MMU type.
-	err = drgn_program_find_object(prog, "cur_cpu_spec", NULL,
+	/*err = drgn_program_find_object(prog, "cur_cpu_spec", NULL,
 				       DRGN_FIND_OBJECT_ANY, &tmp);
 	if (err)
 		return err;
@@ -375,7 +375,7 @@ linux_kernel_pgtable_iterator_create_ppc64(struct drgn_program * prog,
 	if (!(mmu_features & 0x40)) {
 		return drgn_error_create(DRGN_ERROR_NOT_IMPLEMENTED,
 					 "virtual address translation is only supported for Radix MMU");
-	}
+	}*/
 
 	*ret = &no_cleanup_ptr(it)->it;
 	return NULL;
@@ -420,8 +420,14 @@ linux_kernel_pgtable_iterator_next_ppc64(struct drgn_program *prog,
 		uint64_t table;
 		bool table_physical;
 		if (level == levels) {
-			table = it->it.pgtable;
-			table_physical = false;
+			if (it->it.pgtable == prog->vmcoreinfo.swapper_pg_dir) {
+						table = it->it.pgtable &
+								~UINT64_C(0xc000000000000000);
+						table_physical = true;
+				} else {
+						table = it->it.pgtable;
+						table_physical = false;
+				}
 		} else {
 			// PAGE_PTE bit represents huge page.
 			if (!(entry & PAGE_PRESENT) || (entry & PAGE_PTE) || level == 0) {
--- a/libdrgn/qemu_machine_protocol.c
+++ b/libdrgn/qemu_machine_protocol.c
@@ -768,6 +768,171 @@ static struct drgn_error *qmp_read_vmcoreinfo(struct drgn_program *prog,
 	return parse_vmcoreinfo_from_dump(prog, fd);
 }
 
+static struct drgn_error *drgn_qmp_read_vmcoreinfo(struct drgn_program *prog)
+{
+	static const uint64_t addr_mask = ~UINT64_C(0xc000000000000000);
+	/*
+	 * vmcoreinfo_size holds only the descriptor payload size. The full ELF
+	 * note in memory also includes Elf64_Nhdr (12 bytes) + "VMCOREINFO\0"
+	 * padded to 4-byte alignment (12 bytes) = 24 bytes of overhead.
+	 */
+	static const size_t note_overhead =
+		sizeof(Elf64_Nhdr) + ((sizeof("VMCOREINFO") + 3) & ~3U);
+	struct drgn_error *err;
+
+	struct drgn_symbol *note_sym = NULL;
+	struct drgn_symbol *size_sym = NULL;
+
+	err = drgn_program_find_symbol_by_name(prog, "vmcoreinfo_note",
+					       &note_sym);
+	if (err) {
+		drgn_log_error(prog,
+			       "failed to find vmcoreinfo_note symbol: %s",
+			       err->message);
+		return err;
+	}
+
+	err = drgn_program_find_symbol_by_name(prog, "vmcoreinfo_size",
+					       &size_sym);
+	if (err) {
+		drgn_log_error(prog,
+			       "failed to find vmcoreinfo_size symbol: %s",
+			       err->message);
+		drgn_symbol_destroy(note_sym);
+		return err;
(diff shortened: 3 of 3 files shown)


libdrgn/qemu_machine_protocol.c
```diff
@@ -768,6 +768,171 @@ static struct drgn_error *qmp_read_vmcoreinfo(struct drgn_program *prog,
 	return parse_vmcoreinfo_from_dump(prog, fd);
 }
 
+static struct drgn_error *drgn_qmp_read_vmcoreinfo(struct drgn_program *prog)
+{
+	static const uint64_t addr_mask = ~UINT64_C(0xc000000000000000);
```
theotherjimmy, line 773: What is this magic number about?

libdrgn/qemu_machine_protocol.c
```diff
@@ -768,6 +768,171 @@ static struct drgn_error *qmp_read_vmcoreinfo(struct drgn_program *prog,
...
+	 * vmcoreinfo_size holds only the descriptor payload size. The full ELF
+	 * note in memory also includes Elf64_Nhdr (12 bytes) + "VMCOREINFO\0"
+	 * padded to 4-byte alignment (12 bytes) = 24 bytes of overhead.
+	 */
+	static const size_t note_overhead =
+		sizeof(Elf64_Nhdr) + ((sizeof("VMCOREINFO") + 3) & ~3U);
+	struct drgn_error *err;
+
+	struct drgn_symbol *note_sym = NULL;
+	struct drgn_symbol *size_sym = NULL;
```
theotherjimmy, line 784: These should use a cleanup attribute. See here https://github.com/osandov/drgn/blob/main/CONTRIBUTING.rst#c

libdrgn/qemu_machine_protocol.c
```diff
@@ -768,6 +768,171 @@ static struct drgn_error *qmp_read_vmcoreinfo(struct drgn_program *prog,
...
+
+	if (vmcoreinfo_size == 0) {
+		drgn_log_error(prog, "vmcoreinfo_size is zero");
+		return drgn_error_create(DRGN_ERROR_OTHER,
+					 "vmcoreinfo_size is zero");
+	}
+
+	size_t note_size = note_overhead + vmcoreinfo_size;
+
+	void *buf = malloc(note_size);
```
theotherjimmy, line 842: Cleanup attribute here too.

drgn/cli.py
```diff
@@ -445,6 +445,10 @@ def directories_option(arg_name: str, option_name: Optional[str] = None) -> None
             critical = False
         logger.log(logging.CRITICAL if critical else logging.WARNING, "%s", e)
 
+
+def _load_only_extra_debug_symbols(
+    prog: drgn.Program, args: argparse.Namespace
+) -> None:
```
theotherjimmy, line 451: This delta seems unrelated to the commit that it's in. Could it be moved out to another commit or removed?

libdrgn/arch_ppc64.c
```diff
@@ -420,8 +420,14 @@ linux_kernel_pgtable_iterator_next_ppc64(struct drgn_program *prog,
...
-			table = it->it.pgtable;
-			table_physical = false;
+			if (it->it.pgtable == prog->vmcoreinfo.swapper_pg_dir) {
+						table = it->it.pgtable &
+								~UINT64_C(0xc000000000000000);
+						table_physical = true;
+				} else {
+						table = it->it.pgtable;
+						table_physical = false;
+				}
```
theotherjimmy, line 430: Could you break the address translation fixes into another commit?

theotherjimmy (changes requested):
Looks pretty good overall. I left some minor nits to help get this ready as soon as possible.

-------

What's needed to move this out of RFC status in your opinion?

osandov:
Missing VMCOREINFO is a more general problem not limited to QMP, so it'd be nice to solve this more generally. I need to try some stuff out to see where would be the best place to do that, so I'll get back to you on that part.

In the meantime, please do break the address translation improvement into its own PR. That should be straightforward to merge.

osandov:
This doesn't work with KASLR, right?

I have a prototype that hooks into debug info loading: if you provide a vmlinux file (e.g., with `-s`), we try to bootstrap from that at debug info load time. It requires architecture-specific, hardcoded address translation logic like you pointed out. On x86-64, it doesn't work with KASLR, and I think that's also true for ppc64. Arm32 requires trying several common memory layouts. It can't work on Arm64 at all. It also can't work on s390x, but we should be using lowcore there. I'll share it once I get it cleaned up.

I also ran into a bug in the page table iterator recursion detection that overlaps a bit with this. I'll get that fixed, too.

theotherjimmy:
I'm happy to fix the bug with page table iterator recursion detection if that would help, though it might be faster for you since you already have a reproducer.

osandov:
Please test my [`vmcoreinfo-from-vmlinux`](https://github.com/osandov/drgn/tree/vmcoreinfo-from-vmlinux) branch (which also includes the page table iterator recursion [fix](https://github.com/osandov/drgn/commit/4655ff3335422b47f65d2a7ce1263bd7c3a8f57a)). With that, running drgn with `-s <vmlinux>` _should_ be sufficient to read VMCOREINFO as long as KASLR is disabled (e.g., booted with `nokaslr`).

```
$ python3 -m vmtest.vm -k ppc64:7.2.0-rc2-vmtest41.1default --qemu-options '-qmp unix:qemu.sock,server=on,wait=off' --kernel-cmdline nokaslr
$ python3 -m drgn --qemu ./qemu.sock -s ./build/vmtest/ppc64/kernel-7.2.0-rc2-vmtest41.1default/build/vmlinux -e 'print(prog["VMCOREINFO"].string_().decode())'
warning: can't find kernel modules: couldn't read modules list: recursive address translation; page table may be missing from core dump: 0xc0000000009f5d78
OSRELEASE=7.2.0-rc2-vmtest41.1default
BUILD-ID=e1afcbc95610cc9d8785fe93d941c92fc783d74c
...
```

I still get that recursive translation warning, so your address translation fix will be needed regardless.

theotherjimmy:
I've reproduced the failure using revision "7.2.0-rc6-vmtest42.1default", since it's been long enough that rc2 is not available anymore. I'll look into what's causing this specific address translation failure.

theotherjimmy:
Right, so the problem is that creating a virtual addr translation iterator requires reading some kernel structs:
https://github.com/osandov/drgn/blob/main/libdrgn/arch_ppc64.c#L359-L375

```C
        // Identify the MMU type.
        err = drgn_program_find_object(prog, "cur_cpu_spec", NULL,
                                       DRGN_FIND_OBJECT_ANY, &tmp);
        if (err)
                return err;
        err = drgn_object_member_dereference(&tmp, &tmp, "mmu_features");
        if (err)
                return err;
        uint64_t mmu_features;
        err = drgn_object_read_unsigned(&tmp, &mmu_features);

        if (err)
                return err;
        if (!(mmu_features & 0x40)) {
                return drgn_error_create(DRGN_ERROR_NOT_IMPLEMENTED,
                                         "virtual address translation is only supported for Radix MMU");
        }
```

theotherjimmy:
So the required flag is in VMCOREINFO...

```
>>> print(prog["VMCOREINFO"].string_().decode().split()[-2:-1])
['NUMBER(RADIX_MMU)=1']
```

theotherjimmy:
Adding that into the C vmcoreinfo parsing fixes that up:
```bash
python3.14 -m drgn --qemu ./qemu.sock -s ./build/vmtest/ppc64/kernel-7.2.0-rc6-vmtest42.1default/build/vmlinux -e 'print(prog["VMCOREINFO"].string_().decode())'
warning: missing debugging symbols for binfmt_misc
warning: missing debugging symbols for virtio_rng
warning: missing debugging symbols for rng_core
warning: missing some debugging symbols; see https://drgn.readthedocs.io/en/latest/getting_debugging_symbols.html
OSRELEASE=7.2.0-rc6-vmtest42.1default
BUILD-ID=6a53533c4f072eb200d4c1fb058172eb9d4c12f2
```

@osandov Where should I push my changes that correct this error?

libdrgn/qemu_machine_protocol.c
theotherjimmy: Notes for later: This is `PAGE_PRESENT | PAGE_PTE`

theotherjimmy: This actually happens to match `_stext` which is used as the start of the identity map.

osandov:
If the change is just using `NUMBER(RADIX_MMU)` from vmcoreinfo, I'd be happy to take it as a standalone PR.

One complication is that `NUMBER(RADIX_MMU)` was introduced in torvalds/linux@36e826b568e412f61d68fedc02a67b4d8b7583cc in Linux 6.7. For older versions, we could either keep the existing code as a fallback (which still only works if the core dump has virtual address information), simply assume the radix MMU for now, or try to come up with another heuristic.

theotherjimmy:
It's `NUMBER(RADIX_MMU)` with the current implementation as a fallback and `SYMBOL(_stext)` since that's what PPC64 uses for kernel physical memory offset. Further, I only tested this on the `vmcoreinfo-from-vmlinux` branch.

PR #628, opened 2026-06-26
libdrgn: add basic LoongArch64 architecture support

georgejguo:
Add tier-1 support for the LoongArch64 architecture: architecture recognition
  plus reading memory, variables, and types. This mirrors the minimal,
  recognition-only RISC-V support and follows the porting guide in
  `libdrgn/platform.h`.

  ## Changes

  - **`libdrgn/arch_loongarch64.c`** (new): define `arch_info_loongarch64` with
    `name`, `arch`, `default_flags` (64-bit, little-endian), `scalar_alignment`,
    and `register_by_name = drgn_register_by_name_unknown`.
  - **`libdrgn/drgn.h`**: add `DRGN_ARCH_LOONGARCH64` to `enum drgn_architecture`
    (this auto-generates the `Architecture.LOONGARCH64` Python constant).
  - **`libdrgn/platform.h`**: declare `arch_info_loongarch64`.
  - **`libdrgn/platform.c`**: wire up the dispatch points — host platform
    detection (`__loongarch__` / `__loongarch_grlen == 64`),
    `drgn_platform_create()`, and `drgn_platform_from_elf()` (`EM_LOONGARCH`).
    Also add an `#ifndef EM_LOONGARCH` fallback (258), since it was only added in
    glibc 2.34 / binutils 2.36.
  - **`libdrgn/kdump.c`**: recognize `KDUMP_ARCH_LOONGARCH64` (guarded by
    `#ifdef`, since older libkdumpfile lacks the constant).
  - **`libdrgn/qemu_machine_protocol.c`**: recognize the `"loongarch64"` machine.
  - **`libdrgn/Makefile.am`**: add `arch_loongarch64.c` to
    `libdrgn_common_la_SOURCES`.
  - **`_drgn.pyi`**, **`docs/support_matrix.rst`**: document the new architecture.
  - **`tests/elfwriter.py`**: add an optional `e_machine` parameter to
    `create_elf_file()` (backward compatible).
  - **`tests/test_program.py`**: add `TestCoreDump.test_loongarch64`, asserting an
    `EM_LOONGARCH` core dump is detected as `Architecture.LOONGARCH64`.

### DIFF
--- a/libdrgn/kdump.c
+++ b/libdrgn/kdump.c
@@ -46,6 +46,7 @@ static struct drgn_error *drgn_platform_from_kdump(kdump_ctx_t *ctx,
 	else if (strcmp(str, KDUMP_ARCH_RISCV32) == 0)
 		arch = &arch_info_riscv32;
 #endif
+	// libkdumpfile doesn't support LoongArch as of version 0.5.6.
 	else
 		arch = &arch_info_unknown;
 
@@ -95,6 +96,7 @@ static struct drgn_error *drgn_platform_to_kdump(kdump_ctx_t *ctx,
 	else if (platform->arch == &arch_info_riscv32)
 		arch_str = KDUMP_ARCH_RISCV32;
 #endif
+	// libkdumpfile doesn't support LoongArch as of version 0.5.6.
 
 	if (arch_str) {
 		ks = kdump_set_string_attr(ctx, KDUMP_ATTR_ARCH_NAME, arch_str);
--- a/libdrgn/platform.c
+++ b/libdrgn/platform.c
@@ -44,6 +44,8 @@ LIBDRGN_PUBLIC const struct drgn_platform drgn_host_platform = {
 // __s390__ is also defined for s390x, so the order is important.
 #elif __s390__
 	.arch = &arch_info_s390,
+#elif defined(__loongarch__) && __loongarch_grlen == 64
+	.arch = &arch_info_loongarch64,
 #else
 	.arch = &arch_info_unknown,
 #endif
@@ -88,6 +90,9 @@ drgn_platform_create(enum drgn_architecture arch,
 		break;
 	case DRGN_ARCH_S390:
 		arch_info = &arch_info_s390;
+		break;
+	case DRGN_ARCH_LOONGARCH64:
+		arch_info = &arch_info_loongarch64;
 		break;
 	default:
 		return drgn_error_create(DRGN_ERROR_INVALID_ARGUMENT,
@@ -179,6 +184,12 @@ void drgn_platform_from_elf(GElf_Ehdr *ehdr, struct drgn_platform *ret)
 			arch = &arch_info_s390x;
 		else
 			arch = &arch_info_s390;
+		break;
+	case EM_LOONGARCH:
+		if (ehdr->e_ident[EI_CLASS] == ELFCLASS64)
+			arch = &arch_info_loongarch64;
+		else
+			arch = &arch_info_unknown;
 		break;
 	default:
 		arch = &arch_info_unknown;
--- a/_drgn.pyi
+++ b/_drgn.pyi
@@ -2350,6 +2350,9 @@ class Architecture(enum.Enum):
     S390 = ...
     """The 32-bit s390 architecture, a.k.a. System/390."""
 
+    LOONGARCH64 = ...
+    """The 64-bit LoongArch architecture."""
+
     UNKNOWN = ...
     """
     An architecture which is not known to drgn. Certain features are not
--- a/docs/support_matrix.rst
+++ b/docs/support_matrix.rst
@@ -44,6 +44,10 @@ of this support is:
       - ✓
       -
       -
+    * - LoongArch64
+      -
+      -
+      -
 
 .. rubric:: Key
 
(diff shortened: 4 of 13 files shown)


libdrgn/platform.c
```diff
@@ -9,6 +9,12 @@
 #include "error.h"
 #include "platform.h"
 
+// EM_LOONGARCH was added in glibc 2.34 / binutils 2.36. Define it as a fallback
+// for older toolchains.
+#ifndef EM_LOONGARCH
+#define EM_LOONGARCH 258
+#endif
+
```
osandov: We have a centralized place to put these fallback definitions. You can replace this with

```diff
diff --git a/libdrgn/include/elf.h b/libdrgn/include/elf.h
index 44d27374..ed8f241d 100644
--- a/libdrgn/include/elf.h
+++ b/libdrgn/include/elf.h
@@ -160,5 +160,8 @@
 #ifndef NT_ARM_PAC_MASK
 #define NT_ARM_PAC_MASK 0x406
 #endif
+#ifndef EM_LOONGARCH
+#define EM_LOONGARCH 258
+#endif
 
 #endif /* DRGN_ELF_H */
diff --git a/scripts/gen_elf_compat.py b/scripts/gen_elf_compat.py
index 00443942..ade3491e 100755
--- a/scripts/gen_elf_compat.py
+++ b/scripts/gen_elf_compat.py
@@ -65,6 +65,8 @@ MACROS = (
     "R_RISCV_32_PCREL",
     # Added in glibc 2.30.
     "NT_ARM_PAC_MASK",
+    # Added in glibc 2.36.
+    "EM_LOONGARCH",
 )
 
 
```

libdrgn/kdump.c
```diff
@@ -45,6 +45,10 @@ static struct drgn_error *drgn_platform_from_kdump(kdump_ctx_t *ctx,
 		arch = &arch_info_riscv64;
 	else if (strcmp(str, KDUMP_ARCH_RISCV32) == 0)
 		arch = &arch_info_riscv32;
+#endif
+#ifdef KDUMP_ARCH_LOONGARCH64
+	else if (strcmp(str, KDUMP_ARCH_LOONGARCH64) == 0)
+		arch = &arch_info_loongarch64;
```
osandov: It doesn't seem like this exists in any version of libkdumpfile, and I don't see any PRs adding it. For now, let's just add a comment here and in `drgn_platform_to_kdump()` saying that libkdumpfile doesn't support LoongArch as of version 0.5.6.

osandov (changes requested):
Thanks, a couple of minor comments below. Would it also make sense to add the same basic support for LoongArch32 while we're here?

georgejguo:
> Thanks, a couple of minor comments below. Would it also make sense to add the same basic support for LoongArch32 while we're here?

Good point. I looked into LA32 while working on this, but I'd suggest we skip it for now:
- No widely available hardware to test on (mostly FPGA soft cores and embedded SoCs like the Loongson-1 series).
- Loongson's upstream effort is heavily skewed toward LA64; LA32 kernel support is still fairly basic.
- drgn's users are mostly debugging servers and desktops, where LA64 is the only relevant target.

It should be straightforward to add later as a separate change if someone has an LA32 use case and hardware to validate on. Happy to revisit then.

osandov (approved):
Looks good, thanks! I'll merge when tests pass.

Are you planning on adding any more of the architecture support for LoongArch64? At the very least, relocation support is necessary for module support.

georgejguo:
> Looks good, thanks! I'll merge when tests pass.
> 
> Are you planning on adding any more of the architecture support for LoongArch64? At the very least, relocation support is necessary for module support.

Yes — module relocation support is next. I already have it implemented locally: apply_elf_reloc for the data relocations that appear in module debug info (R_LARCH_32/64, R_LARCH_ADD/SUB{8,16,32,64}, and R_LARCH_ADD6/SUB6).  Instruction-field relocations (B16/B21/B26, *_HI20/LO12, PCALA_*, …) only occur in executable sections, which drgn doesn't relocate, so those aren't needed. 
R_LARCH_ADD_ULEB128/SUB_ULEB128 I've left as a TODO for now.

I'll send it as a follow-up PR once this one lands.  After that I'd like to look at stack unwinding and virtual address translation.

georgejguo:
Validated on real LoongArch64 hardware (self-compiled mainline kernel,
`CONFIG_DEBUG_INFO=y`):

```
# Live kernel architecture recognition (reads /proc/kcore, no debuginfo)
$ sudo python3 -c "
import drgn
prog = drgn.Program()
prog.set_kernel()
print(prog.platform)
assert prog.platform.arch == drgn.Architecture.LOONGARCH64
"
Platform(<Architecture.LOONGARCH64: 10>, <PlatformFlags.IS_64_BIT|IS_LITTLE_ENDIAN: 3>)

# Reading a variable + resolving its type (with vmlinux DWARF)
$ sudo python3 -m drgn -s /path/to/vmlinux --main-symbols \
    -e 'print(prog["init_task"].comm)'
(char [16])"swapper/0"
```

The new unit test and the existing test suite pass:

```
$ python3 -m unittest tests.test_program tests.test_platform -v
... 101 tests OK
```

PR #629, opened 2026-06-30
crash: implement dis command

### LINKED ISSUE #331: Add disassembler
We can probably just find a good library to use. The MVP would be an API for disassembling functions and/or arbitrary memory. In the future, it'd be really cool to intermingle the disassembly with an interpretation of the DWARF location descriptions for arguments/local variables.

theotherjimmy:
The bulk of the disassembly is handled by capstone, and matches most of the arguments from crash

## TODO:
 - [x] Implement Tests
 - [x] Find a way to have capstone installed so that the test can pass
 - [x] Correct mypy errors
 
 resolves https://github.com/osandov/drgn/issues/331

### DIFF
--- a/drgn/commands/_crash/_dis.py
+++ b/drgn/commands/_crash/_dis.py
@@ -1,0 +1,91 @@
+# Copyright (c) IBM Corp. 2026
+# SPDX-License-Identifier: LGPL-2.1-or-later
+
+"""Disassmebly crash command (dis)."""
+
+import argparse
+from typing import Any
+
+from drgn import Program
+from drgn.commands import argument, drgn_argument, mutually_exclusive_group
+from drgn.commands._crash.common import CrashDrgnCodeBuilder, crash_command
+from drgn.helpers.common.disasm import disasm
+
+
+@crash_command(
+    description="disassemble",
+    long_description="Disassemble the memory at a specific address.",
+    arguments=(
+        argument(
+            "address",
+            type=str,
+            help="start address of the dissas, specified in addr2line format",
+        ),
+        argument(
+            "count",
+            type="decimal_or_hexadecimal",
+            nargs="?",
+            help="number of instructions to dissassemble",
+        ),
+        mutually_exclusive_group(
+            argument(
+                "-d",
+                dest="offset_base",
+                action="store_const",
+                const=10,
+                help="override default output format with decimal format.",
+            ),
+            argument(
+                "-x",
+                dest="offset_base",
+                action="store_const",
+                const=16,
+                help="override default output format with hexadecimal format.",
+            ),
+        ),
+        mutually_exclusive_group(
+            argument(
+                "-f",
+                dest="forward",
+                action="store_true",
+                help="dissassemble from addr to addr+size.",
+            ),
+            argument(
+                "-r",
+                dest="reverse",
+                action="store_true",
+                help="disassemble from addr-size to addr.",
+            ),
+        ),
+        drgn_argument,
+    ),
+)
+def _crash_cmd_dis(
+    prog: Program, name: str, args: argparse.Namespace, **kwargs: Any
+) -> None:
+    offset_base = args.offset_base or prog.config.get("crash_radix", 10)
+    if args.drgn:
+        code = CrashDrgnCodeBuilder(prog)
+        code.add_from_import("drgn.helpers.common.disasm", "disasm")
+        code.append(
+            f"""\
+disasm(
+    "{args.address}",
+"""
+        )
+        if args.count:
+            code.append(f"    {args.count},\n")
+        if args.reverse:
+            code.append(f"    reverse={args.reverse}\n")
+        if args.offset_base or prog.config.get("crash_radix"):
+            code.append(f"    offset_base={offset_base},\n")
+        code.append(")")
+        code.print()
+    else:
+        disasm(
+            prog,
+            args.address,
+            args.count,
+            reverse=args.reverse,
+            offset_base=args.offset_base or 10,
+        )
--- a/drgn/helpers/common/disasm.py
+++ b/drgn/helpers/common/disasm.py
@@ -1,0 +1,247 @@
+# (C) Copyright IBM Corp. 2026
+# SPDX-License-Identifier: LGPL-2.1-or-later
+
+"""
+Disassembly
+-----------
+The ``drgn.helpers.common.disasm`` module provides helpers for disassembling
+machine code.
+"""
+
+from typing import Optional, Tuple, Union
+
+from _drgn import _parse_addr2line
+from drgn import Architecture, IntegerLike, Program
+from drgn.helpers.common.prog import takes_program_or_default
+
+__all__ = ("disasm",)
+
+try:
+    import capstone  # type: ignore # no type hints available
+
+    _HAVE_CAPSTONE = True
+except ImportError:
+    _HAVE_CAPSTONE = False
+
+
+def _resolve_addr(prog: Program, address: str) -> int:
+    # Resolve an addr2line string into an address
+    # Note that we drop the symbol name here, as
+    # it's valid to compute an address using one symbol
+    # that falls into the region of another symbol.
+    # For example: resloving `__schedule+3022` could
+    # be the same as `yeild_to+622`, and we should
+    # accept the prior
+    sym_name, offset = _parse_addr2line(address)
+    if sym_name:
+        sym = prog.symbol(sym_name)
+        return sym.address + offset
+    else:
+        return offset
+
+
+def _make_capstone_disassembler(arch: Architecture) -> "capstone.Cs":
+    try:
+        capstone_args = {
+            Architecture.X86_64: (capstone.CS_ARCH_X86, capstone.CS_MODE_64),
+            Architecture.I386: (capstone.CS_ARCH_X86, capstone.CS_MODE_32),
+            Architecture.AARCH64: (capstone.CS_ARCH_ARM64, capstone.CS_MODE_ARM),
+            Architecture.ARM: (capstone.CS_ARCH_ARM, capstone.CS_MODE_ARM),
+            Architecture.PPC64: (capstone.CS_ARCH_PPC, capstone.CS_MODE_64),
+            Architecture.RISCV32: (
+                capstone.CS_ARCH_RISCV,
+                capstone.CS_MODE_RISCV32,
+            ),
+            Architecture.RISCV64: (
+                capstone.CS_ARCH_RISCV,
+                capstone.CS_MODE_RISCV64,
+            ),
+            Architecture.S390X: (capstone.CS_ARCH_SYSZ, 0),
+            Architecture.S390: (capstone.CS_ARCH_SYSZ, 0),
+        }[arch]
+    except KeyError:
+        raise NotImplementedError(f"disassably not supported on {arch}")
+    disassembler = capstone.Cs(*capstone_args)
+    try:
+        disassembler.syntax = capstone.CS_OPT_SYNTAX_ATT
+    except capstone.CsError as cse:
+        # An errant backport in capstone 5.0.1 caused mismatched constants
+        # between the C library and Python binding. See:
+        # https://github.com/capstone-engine/capstone/issues/2145
+        # and https://github.com/capstone-engine/capstone/issues/2240
+        # Since this is the version RHEL 10 ships, do a bit of monkey patching
+        if cse.errno == capstone.CS_ERR_OPTION and capstone.CS_OPT_SYNTAX_ATT == 8:
+            disassembler.syntax = 2
+        else:
+            raise
+    # The skipdata mode adds `.byte 0xYY` for instructions that it does not understand
+    # This allows for complete disassembly, even when the instruction stream contains
+    # invalid instructions or instruction that are newer that the version of Capstone
+    # used
+    disassembler.skipdata = True
+    # The detail mode allows examining the operands
+    disassembler.detail = True
+    return disassembler
+
+
+def _get_disasassembler(prog: Program) -> "capstone.Cs":
+    # Get a disassembler for this specific program. This may be a cached instance
+    # stored in the program's cache attribute.
+    if not _HAVE_CAPSTONE:
+        raise NotImplementedError("disassembly support requires Capstone")
+    if not prog.platform:
+        raise ValueError("cannot disassemble without platform")
+    try:
+        disassembler = prog.cache["disassembler"]
+    except KeyError:
+        disassembler = _make_capstone_disassembler(prog.platform.arch)
+        prog.cache["disassembler"] = disassembler
+    return disassembler
+
+
+def _symbolize(prog: Program, addr: int, offset_base: int) -> Optional[str]:
+    try:
+        sym = prog.symbol(addr)
+    except LookupError:
+        return None
+    sym_offset = addr - sym.address
+    if sym_offset == 0:
+        return f"\t<{sym.name}>"
+    num_string = f"0x{sym_offset:x}" if offset_base == 16 else f"{sym_offset:d}"
+    return f"\t<{sym.name}+{num_string}>"
+
+
+def _symbolize_dest(
+    prog: Program, i: "capstone.CsInsn", offset_base: int
+) -> Optional[str]:
+    if i.id == 0 or len(i.operands) == 0:
+        return None
+    op = i.operands[-1]
+    if op.type == capstone.CS_OP_IMM:
+        # Capstone returns signed integers so use & with a mask of the address
+        # size to 'cast to unsigned'. This makes it possible for symbolize to
+        # correctly compute offsets and lookup symbols by address.
+        op_addr = op.imm & ((1 << 8 * prog.address_size()) - 1)
+        symbolized = _symbolize(prog, op_addr, offset_base)
+        if symbolized:
+            return symbolized
+    return None
+
+
+def _find_mc_forward(prog: Program, start: int, count: int) -> Tuple[int, int]:
+    disassembler = _get_disasassembler(prog)
+    current_iaddr = start
+    for _ in range(count):
+        current_ilen = 0
+        machine_code = bytearray()
+        while True:
+            machine_code += prog.read(current_iaddr + current_ilen, 1)
+            current_ilen += 1
+            try:
+                (addr, size, mnem, _) = next(
+                    disassembler.disasm_lite(machine_code, current_iaddr, count=1)
+                )
+            except StopIteration:
+                continue
+            if mnem == disassembler.skipdata_mnem and size == current_ilen:
+                continue
+            current_iaddr = addr + size
+            break
+    total_len = current_iaddr - start
+    return (start, total_len)
+
+
+def _find_mc_reverse(prog: Program, start: int, count: int) -> Tuple[int, int]:
+    # We always want to ensure that the instruction listing includes the
+    # instruction starting at `addr`. This implies that we should look
+    # forward by 1 instruction from `addr`.
+    (_, first_instruction_len) = _find_mc_forward(prog, start, 1)
+    count -= 1
+
+    # After finding the first instruction, we search backwards for the
+    # remaining instructions nibbling a byte at a time
+    disassembler = _get_disasassembler(prog)
+    current_iaddr = start
+    end = start + first_instruction_len
+    for _ in range(count):
+        current_ilen = 0
+        machine_code = bytes()
+        while True:
+            machine_code = prog.read(current_iaddr - current_ilen, 1) + machine_code
+            current_ilen += 1
+            try:
+                (addr, size, mnem, _) = next(
+                    disassembler.disasm_lite(
+                        machine_code, current_iaddr - current_ilen, count=1
+                    )
+                )
+            except StopIteration:
+                continue
+            if size == current_ilen:
+                continue
+            elif size != current_ilen:
+                # We went a nibble too far, and the prior instruction is what we want
+                # even if it was a skipdata.
+                current_iaddr = addr - 1
+                break
+            else:
+                current_iaddr = addr
+                break
+    total_len = end - current_iaddr
+    return (current_iaddr, total_len)
+
+
+@takes_program_or_default
+def disasm(
+    prog: Program,
+    address: Union[IntegerLike, str],
+    count: Optional[int] = None,
+    *,
+    reverse: Optional[bool] = None,
+    offset_base: int = 10,
+) -> None:
+    """
+    Print a disassembly.
+
+    :param address: the start address.
+    :param count: an optional count, in instructions, to disassemble.
+        When not present and the address lies within a symbol, the rest of the symbol is disassembled.
+        When not present and the address is not within a symbol, one instruction is disassembled.
+    :param reverse: The disassembly will end at the specified address instead of
+        starting there.
+    :param offset_base: print all symbol offsets using this base.
+    """
+    disassembler = _get_disasassembler(prog)
+
+    if isinstance(address, str):
+        addr = _resolve_addr(prog, address)
+    else:
+        addr = int(address)
+    try:
+        symbol = prog.symbol(addr)
+    except LookupError:
+        symbol = None
+    if symbol and not count:
+        # When no count is specified, we want to disassemble the remainder of
+        # the symbol. The upper bound of the number of instructions in a function
+        # is certainly <= the number of bytes in a function.
+        count = symbol.size
+    elif not count:
+        count = 1
+    if symbol:
+        (start, nbytes) = (symbol.address, symbol.size)
+    elif not reverse:
+        (start, nbytes) = _find_mc_forward(prog, addr, count)
+    else:
+        (start, nbytes) = _find_mc_reverse(prog, addr, count)
+    insns = disassembler.disasm(prog.read(start, nbytes), start)
+    if reverse:
+        insns = [i for i in insns if i.address <= addr]
+        insns = insns[-count:]
+    else:
+        insns = [i for i in insns if i.address >= addr]
+        insns = insns[:count]
+    for i in insns:
+        addr_str = _symbolize(prog, i.address, offset_base) or ""
+        dest_str = _symbolize_dest(prog, i, offset_base) or ""
+        print(f"0x{i.address:x}{addr_str}:\t{i.mnemonic}\t{i.op_str}{dest_str}")
--- a/drgn/helpers/common/disasm.py
+++ b/drgn/helpers/common/disasm.py
@@ -1,0 +1,261 @@
+# (C) Copyright IBM Corp. 2026
+# SPDX-License-Identifier: LGPL-2.1-or-later
+
+"""
+Disassembly
+-----------
+The ``drgn.helpers.common.disasm`` module provides helpers for disassembling
+machine code.
+"""
+
+from typing import Dict, Optional, Tuple, Union
+
+from _drgn import _parse_addr2line
+from drgn import Architecture, IntegerLike, PlatformFlags, Program
+from drgn.helpers.common.prog import takes_program_or_default
+
+__all__ = ("disasm",)
+
+try:
+    import capstone  # type: ignore # no type hints available
+
+    _HAVE_CAPSTONE = True
+except ImportError:
+    _HAVE_CAPSTONE = False
+
+
+def _resolve_addr(prog: Program, address: str) -> int:
+    # Resolve an addr2line string into an address
+    # Note that we drop the symbol name here, as
+    # it's valid to compute an address using one symbol
+    # that falls into the region of another symbol.
+    # For example: resloving `__schedule+3022` could
+    # be the same as `yeild_to+622`, and we should
+    # accept the prior
+    sym_name, offset = _parse_addr2line(address)
+    if sym_name:
+        sym = prog.symbol(sym_name)
+        return sym.address + offset
+    else:
+        return offset
+
+
+def _make_capstone_disassembler(arch: Architecture) -> "capstone.Cs":
+    try:
+        capstone_args = {
+            Architecture.X86_64: (capstone.CS_ARCH_X86, capstone.CS_MODE_64),
+            Architecture.I386: (capstone.CS_ARCH_X86, capstone.CS_MODE_32),
+            Architecture.AARCH64: (capstone.CS_ARCH_ARM64, capstone.CS_MODE_ARM),
+            Architecture.ARM: (capstone.CS_ARCH_ARM, capstone.CS_MODE_ARM),
+            Architecture.PPC64: (capstone.CS_ARCH_PPC, capstone.CS_MODE_64),
+            Architecture.RISCV32: (
+                capstone.CS_ARCH_RISCV,
+                capstone.CS_MODE_RISCV32,
+            ),
+            Architecture.RISCV64: (
+                capstone.CS_ARCH_RISCV,
+                capstone.CS_MODE_RISCV64,
+            ),
+            Architecture.S390X: (capstone.CS_ARCH_SYSZ, 0),
+            Architecture.S390: (capstone.CS_ARCH_SYSZ, 0),
+        }[arch]
+    except KeyError:
+        raise NotImplementedError(f"disassably not supported on {arch}")
+    disassembler = capstone.Cs(*capstone_args)
+    try:
+        disassembler.syntax = capstone.CS_OPT_SYNTAX_ATT
+    except capstone.CsError as cse:
+        # An errant backport in capstone 5.0.1 caused mismatched constants
+        # between the C library and Python binding. See:
+        # https://github.com/capstone-engine/capstone/issues/2145
+        # and https://github.com/capstone-engine/capstone/issues/2240
+        # Since this is the version RHEL 10 ships, do a bit of monkey patching
+        if cse.errno == capstone.CS_ERR_OPTION and capstone.CS_OPT_SYNTAX_ATT == 8:
+            disassembler.syntax = 2
+        else:
+            raise
+    # The skipdata mode adds `.byte 0xYY` for instructions that it does not understand
+    # This allows for complete disassembly, even when the instruction stream contains
+    # invalid instructions or instruction that are newer that the version of Capstone
+    # used
+    disassembler.skipdata = True
+    # The detail mode allows examining the operands
+    disassembler.detail = True
+    return disassembler
+
+
+def _get_diasassembler(prog: Program) -> "capstone.Cs":
+    # Get a disasembler for this specific program. This may be a cached instance
+    # stored in the program's cache attribute.
+    if not _HAVE_CAPSTONE:
+        raise NotImplementedError("disassembly support requires Capstone")
+    if not prog.platform:
+        raise ValueError("cannot disassemble without platform")
+    try:
+        disassembler = prog.cache["disassembler"]
+    except KeyError:
+        disassembler = _make_capstone_disassembler(prog.platform.arch)
+        prog.cache["disassembler"] = disassembler
+    return disassembler
+
+
+def _symbolize(prog: Program, addr: int, offset_base: int) -> Optional[str]:
+    try:
+        sym = prog.symbol(addr)
+    except LookupError:
+        return None
+    sym_offset = addr - sym.address
+    num_string = f"0x{sym_offset:x}" if offset_base == 16 else f"{sym_offset:d}"
+    return f"\t<{sym.name}+{num_string}>"
+
+
+def _symbolize_dest(
+    prog: Program, i: "capstone.CsInsn", offset_base: int
+) -> Optional[str]:
+    if i.id == 0 or prog.platform is None or len(i.operands) == 0:
+        return None
+    op = i.operands[-1]
+    if op.type == capstone.CS_OP_IMM:
+        is_64_bit = PlatformFlags.IS_64_BIT in prog.platform.flags
+        addr_mask = 0xFFFFFFFFFFFFFFFF if is_64_bit else 0xFFFFFFFF
+        op_addr = op.imm & addr_mask
+        symbolized = _symbolize(prog, op_addr, offset_base)
+        if symbolized:
+            return symbolized
+    return None
+
+
+_INSTRUCTION_ALIGNMENT: Dict[Architecture, int] = {
+    Architecture.X86_64: 1,
+    Architecture.I386: 1,
+    Architecture.AARCH64: 4,
+    Architecture.ARM: 2,
+    Architecture.PPC64: 2,
+    Architecture.RISCV32: 2,
+    Architecture.RISCV64: 2,
+    Architecture.S390X: 2,
+    Architecture.S390: 2,
+}
+
+
+def _find_mc_forward(prog: Program, start: int, count: int) -> Tuple[int, int]:
+    if not prog.platform:
+        raise ValueError("cannot disassemble without platform")
+    alignment = _INSTRUCTION_ALIGNMENT[prog.platform.arch]
+    disassembler = _get_diasassembler(prog)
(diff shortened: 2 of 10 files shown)


drgn/commands/_crash/_dis.py
```diff
@@ -0,0 +1,155 @@
...
+    drgn_argument,
+    mutually_exclusive_group,
+)
+from drgn.commands._crash.common import (
+    _resolve_addr_or_sym,
+    crash_command,
+)
+
+try:
+    from capstone import *
```
osandov: I'd prefer not using a wildcard import for this.

drgn/commands/_crash/_dis.py
```diff
@@ -0,0 +1,155 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# Copyright (c) 2025, Kylin Software, Inc. and affiliates.
+# SPDX-License-Identifier: LGPL-2.1-or-later
+
+"""Memory management-related crash commands (other than kmem)."""
```
osandov: The copyright notice and docstring need to be updated.

drgn/commands/_crash/_dis.py
```diff
@@ -0,0 +1,155 @@
...
+        if args.hexadecimal:
+            print("        print(f\"0x{i.address:x} <{symbol.namer:s}+0x{offset:x}>:\\t{i.mnemonic}\\t{i.op_str}\")")
+        else:
+            print("        print(f\"0x{i.address:x} <{symbol.namer:s}+{offset:d}>:\\t{i.mnemonic}\\t{i.op_str}\")")
+        print("    else:")
+        print("        print(f\"0x{i.address:x}:\\t{i.mnemonic}\\t{i.op_str}\")")
+
+        return None
+    if not HAVE_CAPSTONE:
+        print("The crash dis command reqiires the python capstone library bindings")
```
osandov: Typo in "reqiires"

osandov (commented):
Awesome, I've wanted this for awhile! I didn't look at the command implementation super closely yet, but I noticed a few minor things below.

A couple of additional high-level comments:

- Capstone seems to use Intel x86-64 disassembly syntax, whereas Crash defaults to AT&T syntax. Is there a way to get Capstone to use AT&T syntax?
- You can rebase on my [`capstone`](https://github.com/osandov/drgn/tree/capstone) branch to get the dependency properly set up (including in CI).

theotherjimmy:
There is a switch for AT&T syntax, which I have updated the PR to include. This switch applies to all archs unconditionally, and seems to have no affect when it's not relevant.

----

I've also cherry-picked the relevant commit from the `capstone` branch, and this does not seem to affect my ability to test this locally.

theotherjimmy:
I think I figured out why things were failing for me: My python version was mismatched with the inner vm's version, which caused setup.py to pass in the wrong executable name to the VM, which was an installed version of python that did not have capstone. I'm working through the bugs in the test and generated `--drgn` output now.

theotherjimmy:
And tests now pass!

theotherjimmy:
@osandov I fixed up the tests, added more tests, and resolved the review comments. It's ready for another round of review when you are.

theotherjimmy:
I missed a single mypy error. Sorry for the noise.

--------

mypy may still report errors though; capstone does not have any type info and that's showing up as an error on my mypy:
```
drgn/commands/_crash/_dis.py:22: error: Skipping analyzing "capstone": module is installed, but missing library stubs or py.typed marker  [import-untyped]
```

drgn/commands/_crash/_dis.py
```diff
@@ -0,0 +1,169 @@
...
+
+import argparse
+from typing import Any
+
+from drgn import Architecture, Program
+from drgn.commands import argument, drgn_argument, mutually_exclusive_group
+from drgn.commands._crash.common import _resolve_addr_or_sym, crash_command
+
+try:
+    from capstone import (
```
osandov: Sorry, I should've been more clear. I'd prefer `import capstone` and using the fully-qualified name everywhere instead of importing all of these names.

You can work around the mypy failure with:

```python3
    import capstone  # type: ignore  # No type hints available.
```

drgn/commands/_crash/_dis.py
```diff
@@ -0,0 +1,169 @@
...
+        ),
+        drgn_argument,
+    ),
+)
+def _crash_cmd_dis(
+    prog: Program, name: str, args: argparse.Namespace, **kwargs: Any
+) -> None:
+    if prog.platform is None:
+        return
+    if args.drgn:
```
osandov, line 67: The complexity of the `--drgn` path makes me think that we should factor out a helper to do this that the normal path can use and `--drgn` can recommend. Maybe a new `drgn.helpers.common.disasm` module with something like:

```python3
@takes_program_or_default
def disasm(
    prog: Program,
    address: Union[IntegerLike, str],
    size: Optional[int] = None,
    *,
    offset_base: int = 16,
) -> None:
    ... print disassembly ...
```

Where `address` is an integer address or string in the same formats that `drgn.Program.source_location()` accepts and `offset_base` is 10 to print symbol offsets in decimal or 16 to print them in hexadecimal. If capstone isn't available, it can just raise `NotImplementedError()`.

What do you think?

osandov (changes requested):

drgn/commands/_crash/_dis.py
theotherjimmy: My bad, that makes more sense. I'll get that going now.

drgn/commands/_crash/_dis.py
theotherjimmy: I think that's reasonable. I'll implement that and simplify the commandline implementation.

theotherjimmy:
Finally got it to pass CI. I broke pre-commit, and this just goes to show that i really need to fix it.

drgn/helpers/common/disasm.py
```diff
@@ -0,0 +1,92 @@
...
+            Architecture.ARM: (capstone.CS_ARCH_ARM, capstone.CS_MODE_ARM),
+            Architecture.PPC64: (capstone.CS_ARCH_PPC, capstone.CS_MODE_64),
+            Architecture.S390X: (capstone.CS_ARCH_SYSZ, 0),
+            Architecture.S390: (capstone.CS_ARCH_SYSZ, 0),
+        }[prog.platform.arch]
+    except LookupError:
+        raise NotImplementedError()
+
+    disassembler = capstone.Cs(*capstone_args)
+    disassembler.syntax = capstone.CS_OPT_SYNTAX_ATT
```
brenns10: An interesting note, and to be clear this is nothing wrong with your code, is that this doesn't work on capstone 5.0.1. See capstone-engine/capstone#2145 and capstone-engine/capstone#2240. When I try to run it, I get:
```
>>> %crash dis lookup_fast
drgn: error: Invalid option (CS_ERR_OPTION)
```
Because the `CS_OPT_SYNTAX_ATT` python value (8) is not the same as the enumeration in the capstone header (2).

Normally I wouldn't care, as this is fixed in subsequent v5.0.x versions, but frustratingly RHEL decided to stick with 5.0.1 for their capstone version in RHEL 10. I'd like to see if I can't get them to include the patch from that PR or rebase to 5.0.2 in Centos stream. But separately, I got this working with:

```python
    try:
        disassembler.syntax = capstone.CS_OPT_SYNTAX_ATT
    except capstone.CsError as cse:
        # An errant backport in capstone 5.0.1 caused mismatched constants
        # between the C library and Python binding. See:
        # https://github.com/capstone-engine/capstone/issues/2145
        # and https://github.com/capstone-engine/capstone/issues/2240
        # Since this is the version RHEL 10 ships, do a bit of monkey patching
        # until it's fixed.
        if cse.errno == capstone.CS_ERR_OPTION and capstone.CS_OPT_SYNTAX_ATT == 8:
            disassembler.syntax = 2
        else:
            raise

```
It's kinda an ugly hack and I don't know if I actually want that code to be part of this, but I wanted to at least raise the concern.

drgn/helpers/common/disasm.py
```diff
@@ -0,0 +1,92 @@
...
+
+    :param address: the start address.
+    :param size: an optional size, in bytes, to disasseble. When not present,
+        the address is looked up as a symbol and the symbol size is used.
+    :param reverse: use the address as the end address, and compute the disassebly
+        starting at address - size
+    :param offset_base: print the symbol offset using this base.
+    """
+    if not HAVE_CAPSTONE:
+        raise NotImplementedError()
```
osandov: We should add an explanation to all of the `NotImplementedError`s.
```suggestion
        raise NotImplementedError("disassembly support requires capstone")
```

drgn/helpers/common/disasm.py
```diff
@@ -0,0 +1,92 @@
...
+
+    if not prog.platform:
+        raise NotImplementedError()
```
osandov: This case is similar to stack tracing: https://github.com/osandov/drgn/blob/9c1a2e52bf9d453cd66f0bdf2d853f421f2fecf7/libdrgn/stack_trace.c#L1291-L1292

So it'd probably be more appropriate to use `ValueError`:

```suggestion
        raise ValueError("cannot disassemble without platform")
```

drgn/helpers/common/disasm.py
```diff
@@ -0,0 +1,92 @@
...
+        capstone_args = {
+            Architecture.X86_64: (capstone.CS_ARCH_X86, capstone.CS_MODE_64),
+            Architecture.I386: (capstone.CS_ARCH_X86, capstone.CS_MODE_32),
+            Architecture.AARCH64: (capstone.CS_ARCH_ARM64, capstone.CS_MODE_ARM),
+            Architecture.ARM: (capstone.CS_ARCH_ARM, capstone.CS_MODE_ARM),
+            Architecture.PPC64: (capstone.CS_ARCH_PPC, capstone.CS_MODE_64),
+            Architecture.S390X: (capstone.CS_ARCH_SYSZ, 0),
+            Architecture.S390: (capstone.CS_ARCH_SYSZ, 0),
+        }[prog.platform.arch]
+    except LookupError:
```
osandov: We can be more specific with the exception:
```suggestion
    except KeyError:
```

drgn/helpers/common/disasm.py
```diff
@@ -0,0 +1,92 @@
...
+        raise NotImplementedError()
```
osandov:
```suggestion
        raise NotImplementedError(f"disassembly not supported on {prog.platform.name} architecture")
```

drgn/helpers/common/disasm.py
```diff
@@ -0,0 +1,92 @@
...
+    :param offset_base: print the symbol offset using this base.
+    """
+    if not HAVE_CAPSTONE:
+        raise NotImplementedError()
+
+    if not prog.platform:
+        raise NotImplementedError()
+
+    try:
+        capstone_args = {
```
osandov: What about RISC-V?

drgn/helpers/common/disasm.py
```diff
@@ -0,0 +1,92 @@
...
+            Architecture.AARCH64: (capstone.CS_ARCH_ARM64, capstone.CS_MODE_ARM),
+            Architecture.ARM: (capstone.CS_ARCH_ARM, capstone.CS_MODE_ARM),
+            Architecture.PPC64: (capstone.CS_ARCH_PPC, capstone.CS_MODE_64),
+            Architecture.S390X: (capstone.CS_ARCH_SYSZ, 0),
+            Architecture.S390: (capstone.CS_ARCH_SYSZ, 0),
+        }[prog.platform.arch]
+    except LookupError:
+        raise NotImplementedError()
+
+    disassembler = capstone.Cs(*capstone_args)
```
osandov: Do you have a sense of how expensive it is to create this? Should we create it once and cache it (in `prog.cache`)?

drgn/helpers/common/disasm.py
```diff
@@ -0,0 +1,92 @@
...
+    disassembler.syntax = capstone.CS_OPT_SYNTAX_ATT
+    addr = _resolve_addr(prog, address)
```
osandov: This doesn't handle `symbol+offset` strings like `source_location()`. I just added `_drgn._parse_addr2line()` in my `capstone` branch that you can use for that purpose. Something roughly along these lines:

```python3
from _drgn import _parse_addr2line

if isinstance(address, str):
    sym_name, offset = _parse_addr2line(address)
    if sym_name:
        sym = prog.symbol(sym_name)
        address = sym.address + offset
    else:
        address = offset
```

drgn/commands/_crash/_dis.py
```diff
@@ -0,0 +1,92 @@
...
+            "address",
+            type="addr_or_sym",
+            help="hexadecimal start virtual address or symbol to disassemble",
+        ),
+        argument(
+            "size",
+            type="hexadecimal",
+            nargs="?",
+            help="size of the region to disassemble in bytes",
+        ),
```
brenns10, line 29: The `size` argument in crash actually refers to the number of instructions, not bytes. While this is annoying, I do think it matters. A number of bytes is somewhat meaningless to me (I have no intuition for instruction sizes), whereas the number of instructions is something that I may actually care to specify in a disassembly.

Any thoughts on whether this should be in-scope for strict compatibility @osandov?

Separately from that issue: the `count` flag in crash is interpreted as a decimal number by default, or hexadecimal otherwise. So use `type="decimal_or_hexadecimal"` here regardless.

drgn/helpers/common/disasm.py
```diff
@@ -0,0 +1,92 @@
...
+from drgn.helpers.common.prog import takes_program_or_default
+
+__all__ = ("disasm",)
+
+try:
+    import capstone  # type: ignore # no type hints available
+
+    HAVE_CAPSTONE = True
+except ImportError:
+    HAVE_CAPSTONE = False
```
osandov: Since this is private, let's give it the underscore prefix.

```suggestion
    _HAVE_CAPSTONE = True
except ImportError:
    _HAVE_CAPSTONE = False
```

drgn/helpers/common/disasm.py
```diff
@@ -0,0 +1,92 @@
...
+    disassembler.syntax = capstone.CS_OPT_SYNTAX_ATT
+    addr = _resolve_addr(prog, address)
+    try:
+        symbol = prog.symbol(address)
+    except LookupError:
+        symbol = None
+    if size is None and symbol:
+        if reverse:
+            size = addr - symbol.address
+            addr = symbol.address
```
brenns10: As a user of crash's dis command it's quite important to me that `-r` shows the instructions up to _and including_ the instruction at the requested address. (This is true regardless of whether we end up using units of bytes or instructions...).

drgn/commands/_crash/_dis.py
```diff
@@ -0,0 +1,92 @@
...
+        code = CrashDrgnCodeBuilder(prog)
+        code.add_from_import("drgn.helpers.common.disasm", "disasm")
+        if args.address[0] == "sym":
+            addr_args_str = f'"{args.address[1]}"'
+        else:
+            addr_args_str = f"{args.address[1]}"
+        code.append(
+            f"""\
+disasm(
+    prog,
```
osandov: For `--drgn` output we typically use the implied default program.
```suggestion
```

drgn/commands/_crash/_dis.py
```diff
@@ -0,0 +1,92 @@
...
+from drgn.helpers.common.disasm import disasm
+
+
+@crash_command(
+    description="disassemble",
+    long_description="Disassemble the memory at a specific address.",
+    arguments=(
+        argument(
+            "address",
+            type="addr_or_sym",
```
osandov: Once we switch `disasm()` to use `_parse_addr2line()`, then I believe this can just be a `str`.

drgn/commands/_crash/_dis.py
```diff
@@ -0,0 +1,92 @@
...
+            addr_args_str = f'"{args.address[1]}"'
+        else:
+            addr_args_str = f"{args.address[1]}"
+        code.append(
+            f"""\
+disasm(
+    prog,
+    {addr_args_str},
+    {args.size},
+    reverse={args.reverse},
```
osandov: Can we make adding the `reverse` argument conditional on it actually being `True` to not clutter the default `False` case?

drgn/commands/_crash/_dis.py
```diff
@@ -0,0 +1,92 @@
...
+    """
+        )
+        code.print()
+    else:
+        disasm(
+            prog,
+            args.address[1],
+            args.size,
+            reverse=args.reverse,
+            offset_base=args.offset_base or 10,
```
osandov, line 90: Does `dis` honor the `set radix` option? If so, this should be:

```suggestion
            offset_base=args.offset_base or prog.config.get("crash_radix", 10),
```

Same with the `--drgn` case, so maybe the logic can be deduplicated.

drgn/helpers/common/disasm.py
```diff
@@ -0,0 +1,92 @@
+from typing import Optional, Union
```
osandov: File needs a copyright header and a top-level docstring like:


```suggestion
"""
Disassembly
-----------

The ``drgn.helpers.common.disasm`` module provides helpers for disassembling
machine code.
"""

from typing import Optional, Union
```

tests/linux_kernel/crash_commands/test_dis.py
```diff
@@ -0,0 +1,48 @@
+from tests.linux_kernel.crash_commands import CrashCommandTestCase
```
osandov, line 6: File needs a copyright header.

osandov (changes requested):
Thanks, I like this structure much more. Would you mind splitting the helper into its own preparatory patch? It's probably fine that it just gets tested via the `dis` command for now. Other than that, just some minor fixes below.

drgn/helpers/common/disasm.py
```diff
@@ -0,0 +1,92 @@
...
+    machine_code = prog.read(addr, size)
+    for i in disassembler.disasm(machine_code, addr):
+        if symbol:
+            offset = i.address - symbol.address
+            num_string = f"0x{offset:x}" if offset_base == 16 else f"{offset:d}"
+            print(
+                f"0x{i.address:x} <{symbol.name:s}+{num_string:s}>:\t{i.mnemonic}\t{i.op_str}"
+            )
+        else:
+            print(f"0x{i.address:x}:\t{i.mnemonic}\t{i.op_str}")
```
brenns10: This may be a wishlist item, because it looks to be architecture specific.

One thing that I find really useful in crash/GDB disassembly is its ability to tell give you branch/jump target addresses as a `symbol+offset`. For example:

```
0xffffffff91d0f374 <lookup_fast+20>:    je      0xffffffff91d0f3e8
vs:
0xffffffff91d0f374 <lookup_fast+20>:    je     0xffffffff91d0f3e8 <lookup_fast+136>
```

It seems like Capstone has a "detail=True" mode which can give you details about instruction operands. For immediate operands, or for memory operands, it can be possible to determine the address (depending on addressing mode) and then convert it to a symbol and add that to the instruction.

brenns10: (But like I said, this is more of note/wishlist... not a request that you implement this for each architecture!)

drgn/helpers/common/disasm.py
theotherjimmy: It seems like there is minimal cost here, but it's not a lot of work to put in in `prog.cache["disassebler"]`

drgn/helpers/common/disasm.py
theotherjimmy: I'm including it. Your explanation makes it easy enough to delete when the issue is resolved across all versions we care about. Or we can leave it in, it's not much code and is unlikely to see bit rot since the check is so specific.

drgn/commands/_crash/_dis.py
theotherjimmy: Ah, honestly this is way harder than it looks like at first, since x86 supports variable instruction widths from 1 to something like 24 bytes. I'll have to dig deeper into the capstone API and see what I can come up with.

drgn/helpers/common/disasm.py
theotherjimmy: This I can do with bytes by reading some extra bytes and aborting once we're past the official end of the instructions to disasseble. 

You're looking for something like:
```
>>> disasm(0xffffffff9df67a7f, reverse=True)
0xffffffff9df67a70 <__schedule+0>:	pushq	%r14
0xffffffff9df67a72 <__schedule+2>:	pushq	%r13
0xffffffff9df67a74 <__schedule+4>:	pushq	%r12
0xffffffff9df67a76 <__schedule+6>:	pushq	%rbp
0xffffffff9df67a77 <__schedule+7>:	pushq	%rbx
0xffffffff9df67a78 <__schedule+8>:	subq	$0x20, %rsp
0xffffffff9df67a7c <__schedule+12>:	cmpl	$1, %edi
0xffffffff9df67a7f <__schedule+15>:	movq	%gs:0x2216581(%rip), %r12
```

drgn/helpers/common/disasm.py
theotherjimmy: I'll try to get this in too.

theotherjimmy: Ah, turns out this is way harder than I initially thought. Could I do this as a follow on PR?

drgn/commands/_crash/_dis.py
theotherjimmy: Capstone does have a `count` field that could be used to help with the forward case, but the reverse case is particularly difficult.

drgn/helpers/common/disasm.py
theotherjimmy: Not as bad as I thought:
```
>>> disasm(0xffffffff9df68081)
0xffffffff9df68081 <__schedule+1553>:	callq	0xffffffff9d6563b0	<__ubsan_handle_out_of_bounds+0>
0xffffffff9df68086 <__schedule+1558>:	jmp	0xffffffff9df67ace	<__schedule+94>
0xffffffff9df6808b <__schedule+1563>:	callq	0xffffffff9df60b00	<__stack_chk_fail+0>
0xffffffff9df68090 <__schedule+1568>:	movq	$18446744072074020928, %rdi
0xffffffff9df68097 <__schedule+1575>:	callq	0xffffffff9ca0f560	<panic+0>
0xffffffff9df6809c <__schedule+1580>:	nopl	(%rax)
```

drgn/helpers/common/disasm.py
theotherjimmy: I'm not seeing that in the capstone branch. Have you pushed it?

osandov: Oops, just pushed it.

theotherjimmy: Thanks. I cherry-picked it into my branch where it belongs in the commit history.

theotherjimmy:
I just pushed a big update that should resolve every review comment I've marked resolved.

Short list:
 * addr2line format is expected by drgn `disasm` and crash `dis`
 * jump and call instructions print destination `<symbol+offset>`
 * reverse includes the instruction at the address specified
 * capstone disassembler cached in the `prog.cache`
 * crash `dis` heeds `set radix` both with and without `--drgn`
 * enabled capstone's "skipdata" mode that emits `.bytes 0x..` for instructions it does not recognize.

theotherjimmy:
Sorry for the push noise. I'm having to tip toe around using capstone types in function signatures, as it'll cause failures when capstone is not installed, which we have to allow since it's an optional dependency.

brenns10:
> I'm having to tip toe around using capstone types in function signatures, as it'll cause failures when capstone is not installed

You should be able to annotate the types using a [string literal](https://typing.python.org/en/latest/spec/annotations.html#string-annotations) rather than the actual type, e.g.:

```python
def foo(arg: "capstone.Cs") -> None:
    ...
```

This is originally intended to be used to resolve circular import issues, or to reference the class name from a method annotation. But it works equally well here - all environments w/ type checking should have the optional dependency.

theotherjimmy:
I'm working on a much refactored variant of the helpers to unify the addr2line printing (for both source instruction address, and jump/call destination address). This actually caught a bug: the destination addr2line strings did not heed offset_base.

Further, it turns out that a better heuristic for figuring out if something should have a destination addr2lrine string is just to check if the last operand is an immediate. For example `BRASL`, a s390x call-like instruction (branch relative and save long) does not have a jump or call group. It does, however, have an immediate that can be computed and looked up as a symbol.

I'll be pushing this new version soon.

drgn/helpers/common/disasm.py
```diff
@@ -0,0 +1,163 @@
...
+        the address is looked up as a symbol and the symbol size is used.
+    :param reverse: use the address as the end address, and compute the disassebly
+        starting at address - size
+    :param offset_base: print the symbol offset using this base.
+    """
+    if not _HAVE_CAPSTONE:
+        raise NotImplementedError("disassebly support requires Capstone")
+
+    if not prog.platform:
+        raise ValueError("cannot disasseble without platform")
```
osandov: Typos:

```suggestion
        raise NotImplementedError("disassembly support requires Capstone")

    if not prog.platform:
        raise ValueError("cannot disassemble without platform")
```

drgn/helpers/common/disasm.py
```diff
@@ -0,0 +1,163 @@
...
+def _dest_addr2line(
+    prog: Program, i: "capstone.CsInsn", offset_base: int
+) -> Optional[str]:
+    if i.id == 0 or prog.platform is None or len(i.operands) == 0:
+        return None
+    op = i.operands[-1]
+    if op.type == capstone.CS_OP_IMM:
+        is_64_bit = PlatformFlags.IS_64_BIT in prog.platform.flags
+        addr_mask = 0xFFFFFFFFFFFFFFFF if is_64_bit else 0xFFFFFFFF
+        op_addr = op.imm & addr_mask
```
osandov: Why truncate to the address size?

drgn/helpers/common/disasm.py
```diff
@@ -0,0 +1,163 @@
...
+
+
+def _addr_to_addr2line(prog: Program, addr: int, offset_base: int) -> Optional[str]:
+    try:
+        sym = prog.symbol(addr)
+        sym_offset = addr - sym.address
+        num_string = f"0x{sym_offset:x}" if offset_base == 16 else f"{sym_offset:d}"
+        return f"\t<{sym.name}+{num_string}>"
+    except LookupError:
+        return None
```
osandov, line 106: Minimize the try block:

```suggestion
    try:
        sym = prog.symbol(addr)
    except LookupError:
        return None
    sym_offset = addr - sym.address
    num_string = f"0x{sym_offset:x}" if offset_base == 16 else f"{sym_offset:d}"
    return f"\t<{sym.name}+{num_string}>"
```

drgn/helpers/common/disasm.py
```diff
@@ -0,0 +1,163 @@
...
+    # This allows for complete disassebly, even when the instruction stream contains
+    # invalid instructions or instruction that are newer that the version of Capstone
+    # used
+    disassembler.skipdata = True
+    # The detail mode allows examining the operands
+    disassembler.detail = True
+    return disassembler
+
+
+def _addr_to_addr2line(prog: Program, addr: int, offset_base: int) -> Optional[str]:
```
osandov: The drgn codebase calls this operation "symbolize":

```suggestion
def _symbolize(prog: Program, addr: int, offset_base: int) -> Optional[str]:
```

drgn/helpers/common/disasm.py
```diff
@@ -0,0 +1,163 @@
...
+    try:
+        sym = prog.symbol(addr)
+        sym_offset = addr - sym.address
+        num_string = f"0x{sym_offset:x}" if offset_base == 16 else f"{sym_offset:d}"
+        return f"\t<{sym.name}+{num_string}>"
+    except LookupError:
+        return None
+
+
+def _dest_addr2line(
```
osandov:
```suggestion
def _symbolize_dest(
```

drgn/helpers/common/disasm.py
```diff
@@ -0,0 +1,163 @@
...
+        addr2line = _addr_to_addr2line(prog, op_addr, offset_base)
+        if addr2line:
+            return addr2line
+    return None
+
+
+@takes_program_or_default
+def disasm(
+    prog: Program,
+    addr2line: str,
```
osandov: This should also accept bare integers:

```suggestion
    address: Union[IntegerLike, str],
```

drgn/helpers/common/disasm.py
```diff
@@ -0,0 +1,163 @@
...
+        symbol = None
+    if size is None and symbol:
+        if reverse:
+            size = addr - symbol.address
+            addr = symbol.address
+        else:
+            size = symbol.size - (addr - symbol.address)
+    size = size or 20
+    end = addr + size
+    machine_code = prog.read(addr, size + 16)
```
osandov: What is this + 16 for?

osandov (changes requested):
I'm glad that @brenns10 pointed out that in crash, it's `count`, not `size`, because this seems like a big enough difference that I'd rather not differ from crash. If you want to land the basic functionality, then we can drop support for `size`/`count` entirely for now and come back to it (which means that we'll either need to error out when we can't find the symbol size or just try to decode one instruction).

Other than that, some details in the code.

drgn/helpers/common/disasm.py
theotherjimmy: This is doing signed to unsigned conversion. I noticed that large addresses, like those used in the kernel sometimes, would show up from capstone as negative numbers. this mask 'converts' them to unsigned.

drgn/helpers/common/disasm.py
theotherjimmy: There were another 3 missing 'm's. I wonder if my keyboard is doing okay.

I fixed them all.

drgn/helpers/common/disasm.py
theotherjimmy: That was to handle the case when disasm is passed both an address and reverse=True. In that instance, you want to include the instruction at the address, which actually extends past the end of the buffer you would normally want to read into. I'll admit that 16 is a bit of a hack, as it's probably the maximum number of bytes needed for an instruction across all instruction sets supported by drgn:
 * x86-64: 16 bytes is _probably_ the longest legal instruction
 * Arm64, RiskV, PPC, Loongarch: 4 bytes is the longest
 * s390x: 6 bytes
 
 So I took the max. I'll think on how better to do this. The solution might honestly be to disassemble in a loop until the address is passed, or the count is met? But I'm still not sure how to correctly disassemble 'backwards' without symbol information on variable instruction width ISAs.

theotherjimmy: This is removed by doing everything by instruction counts.

theotherjimmy:
Right, I got everything working by count. From a dump without module info:
```
>>> disasm(next(for_each_module()).syms.value.value_(), count=8, reverse=True)
0x80000800001ee:	balr	%r8, %r0
0x80000800001f0:	.byte	0x00, 0x00
0x80000800001f2:	.byte	0x00, 0x00
0x80000800001f4:	ssm	0x1f0
0x80000800001f8:	.byte	0x00, 0x00
0x80000800001fa:	.byte	0x00, 0x00
0x80000800001fc:	ssm	0x1f0
0x8000080000200:	.byte	0x00, 0x00
>>> disasm(next(for_each_module()).syms.value.value_(), count=8)
0x8000080000200:	.byte	0x00, 0x00
0x8000080000202:	.byte	0x00, 0x00
0x8000080000204:	.byte	0x00, 0x00
0x8000080000206:	.byte	0x00, 0x00
0x8000080000208:	.byte	0x00, 0x00
0x800008000020a:	.byte	0x00, 0x00
0x800008000020c:	.byte	0x00, 0x00
0x800008000020e:	.byte	0x00, 0x00
```

it's trying, and that's what counts (pun intended)

theotherjimmy:
I'm doing the relevant history rewriting before pushing this up.

theotherjimmy:
Failure looks unrelated to implementing or testing dis; one of the VMs paniced

drgn/helpers/common/disasm.py
```diff
@@ -0,0 +1,261 @@
...
+    # invalid instructions or instruction that are newer that the version of Capstone
+    # used
+    disassembler.skipdata = True
+    # The detail mode allows examining the operands
+    disassembler.detail = True
+    return disassembler
+
+
+def _get_diasassembler(prog: Program) -> "capstone.Cs":
+    # Get a disasembler for this specific program. This may be a cached instance
```
osandov: A couple of typos:
```suggestion
def _get_disassembler(prog: Program) -> "capstone.Cs":
    # Get a disassembler for this specific program. This may be a cached instance
```

drgn/helpers/common/disasm.py
```diff
@@ -0,0 +1,261 @@
...
+def _symbolize_dest(
+    prog: Program, i: "capstone.CsInsn", offset_base: int
+) -> Optional[str]:
+    if i.id == 0 or prog.platform is None or len(i.operands) == 0:
+        return None
+    op = i.operands[-1]
+    if op.type == capstone.CS_OP_IMM:
+        is_64_bit = PlatformFlags.IS_64_BIT in prog.platform.flags
+        addr_mask = 0xFFFFFFFFFFFFFFFF if is_64_bit else 0xFFFFFFFF
+        op_addr = op.imm & addr_mask
```
osandov: Let's add a comment explaining the signed -> unsigned conversion. I also have a slight preference for doing it like this:

```suggestion
    if i.id == 0 is None or len(i.operands) == 0:
        return None
    op = i.operands[-1]
    if op.type == capstone.CS_OP_IMM:
        op_addr = op.imm & ((1 << 8 * prog.address_size()) - 1)
```

drgn/helpers/common/disasm.py
```diff
@@ -0,0 +1,261 @@
...
+    Architecture.RISCV64: 2,
+    Architecture.S390X: 2,
+    Architecture.S390: 2,
+}
+
+
+def _find_mc_forward(prog: Program, start: int, count: int) -> Tuple[int, int]:
+    if not prog.platform:
+        raise ValueError("cannot disassemble without platform")
+    alignment = _INSTRUCTION_ALIGNMENT[prog.platform.arch]
```
osandov: I'd rather shut up the type checker than introduce redundant runtime checks:

```suggestion
    alignment = _INSTRUCTION_ALIGNMENT[prog.platform.arch]  # type: ignore[union-attr]  # platform can't be None.
```

drgn/helpers/common/disasm.py
```diff
@@ -0,0 +1,261 @@
...
+        is_64_bit = PlatformFlags.IS_64_BIT in prog.platform.flags
+        addr_mask = 0xFFFFFFFFFFFFFFFF if is_64_bit else 0xFFFFFFFF
+        op_addr = op.imm & addr_mask
+        symbolized = _symbolize(prog, op_addr, offset_base)
+        if symbolized:
+            return symbolized
+    return None
+
+
+_INSTRUCTION_ALIGNMENT: Dict[Architecture, int] = {
```
osandov: Is this required for correctness or is it an optimization that allows us to read multiple bytes at a time for some architectures?

drgn/helpers/common/disasm.py
```diff
@@ -0,0 +1,261 @@
...
+    if not prog.platform:
+        raise ValueError("cannot disassemble without platform")
+    alignment = _INSTRUCTION_ALIGNMENT[prog.platform.arch]
+    disassembler = _get_diasassembler(prog)
+    current_iaddr = start
+    for _ in range(count):
+        current_ilen = 0
+        while True:
+            current_ilen += alignment
+            machine_code = prog.read(current_iaddr, current_ilen)
```
osandov: This re-reads byte quadratically in the length of the instruction. Instead, can you create a bytearray() that you continually extend so that you never have to re-read a byte?

drgn/helpers/common/disasm.py
```diff
@@ -0,0 +1,261 @@
...
+            try:
+                (addr, size, mnem, _) = next(
+                    disassembler.disasm_lite(machine_code, current_iaddr, count=1)
+                )
+            except StopIteration:
+                continue
+            if mnem == disassembler.skipdata_mnem and size == current_ilen:
+                continue
```
osandov, line 147: The `size == current_ilen` check causes this to accept a partial instruction on x86-64. E.g., if we're disassembling `48 89 e5` (`movq %rsp, %rbp`), this does:

```pycon
>>> list(d.disasm_lite(b"\x48", 0xf000, count=1))
[(61440, 1, '.byte', '0x48')]
>>> list(d.disasm_lite(b"\x48\x89", 0xf000, count=1))
[(61440, 1, '.byte', '0x48')]
```

And never makes it to:

```pycon
>>> list(d.disasm_lite(b"\x48\x89\xe5", 0xf000, count=1))
[(61440, 3, 'movq', '%rsp, %rbp')]
```

We might need to do this probing step with `skipdata = False` to avoid this.

drgn/helpers/common/disasm.py
```diff
@@ -0,0 +1,261 @@
...
+
+
+def _symbolize(prog: Program, addr: int, offset_base: int) -> Optional[str]:
+    try:
+        sym = prog.symbol(addr)
+    except LookupError:
+        return None
+    sym_offset = addr - sym.address
+    num_string = f"0x{sym_offset:x}" if offset_base == 16 else f"{sym_offset:d}"
+    return f"\t<{sym.name}+{num_string}>"
```
osandov, line 111: Omit the offset if it is 0.

osandov (commented):
Thanks for working on the `count` change. A few things to fix there.

drgn/helpers/common/disasm.py
theotherjimmy: Ah, I really wish my spell check would actually have the word disassembler in it so that I could use it on this program.

drgn/helpers/common/disasm.py
theotherjimmy: It's mostly an optimization, though I don't know what would happen if I tried to disassemble aarch64 with 3 bytes of data. Probably nothing good, but might also be fine. I'll give it a go.

drgn/helpers/common/disasm.py
theotherjimmy: That's why there is a check for `mnem == disassembler.skipdata_mnem` in the if.

theotherjimmy: If we probe with skipdata == false, then capstone ends disassembly when it encounters non-instruction data. We need to keep it on so that capstone tries to find the instruction after the data, and we don't have to re-implement its find-instruction-after-data algo

theotherjimmy: I missed that the second disassembly (of `b"\x48\x89"`) is a single byte result implying that `size != current_ilen`. I'm not sure what I can do about this, as turning off skipdata means that this will disassemble to `[]` but if you actually have a true piece of data between instructions, the loop does not end, as the next instruction is always data. This is a bit of a problem, and I don't know that it can easily be resolved.

My counterexample uses aarch64, 
```
>>> list(d.disasm_lite(b"\x00\x00\x80\x08\x00\x00\x00\x00", 0xf000, count=1))
[]
>>> d.skipdata=True
>>> list(d.disasm_lite(b"\x00\x00\x80\x08\x00\x00\x00\x00", 0xf000, count=1))
[(61440, 4, '.byte', '0x00, 0x00, 0x80, 0x08')]
>>> list(d.disasm_lite(b"\x00\x00\x80\x08\x00\x00\x00\x00", 0xf000, count=2))
[(61440, 4, '.byte', '0x00, 0x00, 0x80, 0x08'), (61444, 4, 'andeq', 'r0, r0, r0')]
>>> d.skipdata=False
>>> list(d.disasm_lite(b"\x00\x00\x80\x08\x00\x00\x00\x00", 0xf000, count=2))
[]
```

PR #630, opened 2026-07-08
testsuite: Skip tests for unavailable kernel modules

sasshka:
When a module is not supported and not loaded, the test suite fails on all the tests as it unconditionally tries to load the module.

Replace the single modprobe -a call with a per-module loop using subprocess.run, so one missing module doesn't blow up the whole suite. Tracks which modules loaded successfully in loaded_modules. Add require_module method to
LinuxKernelTestCase so any test can cleanly skip itself if a needed module isn't available. Add self.require_module("loop") to test_loop_device, and self.require_module("btrfs") to test_btrfs_subvolume and test_super_block_on_block_device.

### DIFF
--- a/tests/linux_kernel/__init__.py
+++ b/tests/linux_kernel/__init__.py
@@ -73,6 +73,11 @@ except ImportError:
 class LinuxKernelTestCase(TestCase):
     prog = None
     skip_reason = None
+    loaded_modules = set()
+
+    def require_module(self, module):
+        if module not in LinuxKernelTestCase.loaded_modules:
+            self.skipTest(f"module {module} is not available")
 
     @staticmethod
     def _load_debug_info(prog):
@@ -119,9 +124,10 @@ class LinuxKernelTestCase(TestCase):
                     )
                 else:
                     # Load modules that are used by test cases.
-                    subprocess.check_call(
-                        ["modprobe", "-a", "btrfs", "configs", "loop"]
-                    )
+                    for module in ["btrfs", "configs", "loop"]:
+                        result = subprocess.run(["modprobe", module])
+                        if result.returncode == 0:
+                            LinuxKernelTestCase.loaded_modules.add(module)
                     try:
                         cls._load_debug_info(prog)
                         LinuxKernelTestCase.prog = prog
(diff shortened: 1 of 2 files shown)


osandov (approved):
Looks reasonable. What test environment did you run into this in?

prudo1:
Hi Alexandra,
Hi Omar,

the issue was found on RHEL where btrfs isn't supported. So running the test suite on RHEL kept failing all the time. @sasshka, IMHO that's a detail worth mentioning in the commit message.

An other tiny nit/comment. Both `tests/linux_kernel/helpers/test_kconfig.py:TestKconfig` and `tests/linux_kernel/crash_commands/test_sys.py:TestSys` make use of the `configs` module and have a way to deal with it not being loaded. So strictly speaking having the check there isn't required. Nevertheless using the same method everywhere would make it more consistent and easier to `grep` for. @osandov any opinion?

Thanks
Philipp

osandov:
Agreed on mentioning that in the commit message. For the kconfig tests, I'd rather keep that gated on the existence of `/proc/config.gz` and save the module checks for cases where we can't easily detect availability.

prudo1:
@osandov sure. I just thought it's worth mentioning the two test cases so it's documented that they make use of one of the modules as well.

tests/linux_kernel/__init__.py
```diff
@@ -119,9 +124,10 @@ def setUpClass(cls):
                     )
                 else:
                     # Load modules that are used by test cases.
-                    subprocess.check_call(
-                        ["modprobe", "-a", "btrfs", "configs", "loop"]
-                    )
+                    for module in ["btrfs", "loop"]:
```
osandov:
```suggestion
                    for module in ["btrfs", "configs", "loop"]:
```

We still need to load `configs`, we just don't need to use `require_module("configs")` anywhere since we check for its presence directly.

osandov (changes requested):
Thanks for updating the commit message. One last thing, then this looks good.

tests/linux_kernel/__init__.py
sasshka: Sorry, removing configs was a mistake.

PR #632, opened 2026-07-20
ci: update to Noble / 24.04 LTS

brenns10:
Just wanted to start a discussion. The motivation here (see commit message) is to get a more recent libbpf in the CI. There are multiple breaking changes in libbpf 1.0, and versions prior to 1.0 don't support BTF_ENUM64.

The issue with supporting older versions of libbpf is that we use it to parse any kernel's data. So running tests against a bleeding-edge kernel with libbpf 0.5.0 is not going to work. Nor will an old version of libbpf understand an ENUM64 type that I included in my synthetic debug data.

So there are three possibilities right here, and I've taken the laziest route first:
1. Use a newer CI image and hope that new kernels don't have any backwards-incompatible BTF features for a while.
2. Download and build a suitably recent libbpf in a CI step (or maybe put it into another dummy git tag).
3. Drop the libbpf dependency altogether, bundle the latest `btf.h` header from the kernel, and parse the BTF format ourselves.

Honestly, number 3 isn't terrible -- that's what my original BTF branch did a long time ago. I migrated to use libbpf because it seemed like the "correct" option. But given how tightly coupled it is to the target kernel, number 3 could be preferred. If not, then number 1, i.e. this PR, would be the best step for right now.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,7 +35,7 @@ concurrency:
 
 jobs:
   test:
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
     strategy:
       matrix:
         python-version: ${{ (github.event_name == 'push' || inputs.test_all_python_versions)
--- a/.github/workflows/vmtest-build.yml
+++ b/.github/workflows/vmtest-build.yml
@@ -13,7 +13,7 @@ jobs:
         arch: [x86_64, aarch64, ppc64, s390x, arm, riscv64]
       fail-fast: false
       max-parallel: 6
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
     permissions:
       contents: write
     env:


osandov (approved):
For some reason I thought that the 24.04 image didn't support Python 3.8 anymore, but that doesn't seem to be the case. Regardless of the libbpf decision, this seems fine to me (assuming that tests pass).

For libbpf, will there be a hard version requirement, or are you planning on supporting old libbpf versions as long as the kernel version is old enough, too?

brenns10:
> I thought that the 24.04 image didn't support Python 3.8 anymore

I thought that as well, so I was pleasantly surprised that this worked fine on [my run](https://github.com/brenns10/drgn/actions/runs/29618038513).

> For libbpf, will there be a hard version requirement, or are you planning on supporting old libbpf versions as long as the kernel version is old enough, too?

I was planning on setting a minimum of libbpf 1.0 (August 2022), so that we don't need to deal with enum64 or the API changes (related to how error codes are returned). I would have liked to set a minimum beyond libbpf/libbpf@5b7613e, but v1.4.0 is just too recent to reasonably require (Oct 2024).

I'm explicitly not trying to support all sorts of ancient versions. For Oracle at least, we have CTF that can work on those older platforms. I deal with ancient libctf versions in that fork and I don't want to contribute those problems upstream ;)

osandov:
> I was planning on setting a minimum of libbpf 1.0 (August 2022), so that we don't need to deal with enum64 or the API changes (related to how error codes are returned). I would have liked to set a minimum beyond [libbpf/libbpf@5b7613e](https://github.com/libbpf/libbpf/commit/5b7613e), but v1.4.0 is just too recent to reasonably require (Oct 2024).

Requiring 1.0 sounds reasonable to me since this will be a brand new feature. We'll have to build and install it from source in the manylinux builds (manylinux2014/CentOS 7 doesn't have libbpf at all, and manylinux_2_28/Alma 8 only has 0.5), but that's totally fine.

PR #634, opened 2026-07-29
vmtest: roll vmtest-build back to Ubuntu 22.04

osandov:
Ubuntu 24.04 uses glibc 2.39, so it produces kernel build tools (e.g., objtool) that are too new for my development machines on CentOS Stream 9, which has glibc 2.34. Roll back to Ubuntu 22.04.

@brenns10 will this cause any issues for the BTF work? I hope not since the CI will still be on Ubuntu 24.04.

### DIFF
--- a/.github/workflows/vmtest-build.yml
+++ b/.github/workflows/vmtest-build.yml
@@ -13,7 +13,7 @@ jobs:
         arch: [x86_64, aarch64, ppc64, s390x, arm, riscv64]
       fail-fast: false
       max-parallel: 6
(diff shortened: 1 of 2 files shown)


brenns10:
Nope, that shouldn't be an issue. I had included vmtest-build for consistency with CI. But I only needed CI on 24.04 for the libbpf 1.0+ dependency. As far as I can tell, most (if not all) libbpf versions are quite backward compatible with the actual BTF, so kernels built with older pahole + libbpf should be fine.

PR #635, opened 2026-07-30
drgn.helpers.linux.printk: fix text offset for wrapped records

nilsstrassle:
When a printk record wraps around the end of the lockless ring buffer, the block is relocated to the start of the ring. Each block begins with the descriptor's id (ulong_size bytes) followed by the text. In `_get_printk_records_lockless()`, lpos_begin was set to 0 when wrapping, so the returned text included the id.

For example on an aarch64 vmcore with a wrapped printk record in the buffer, `get_dmesg()` includes the descriptor's id for that block in the returned string:
```
[  593.756214] p\xfd\xff\xff\x00\x00\x00\x00
```

Setting `lpos_begin` to `ulong_size` when wrapping around the end of the buffer in `helpers/linux/printk.py` fixes the issue.
```python
lpos_begin = ulong_size
```

The first commit unpacks `lpos_begin`, `lpos_next` and `text_len` to int so `lpos_begin` can be set directly, and the second commit contains the fix.

### DIFF
--- a/drgn/helpers/linux/printk.py
+++ b/drgn/helpers/linux/printk.py
@@ -136,7 +136,7 @@ def _get_printk_records_lockless(prog: Program, prb: Object) -> List[PrintkRecor
             return
         if lpos_begin > lpos_next:
             # Data wrapped.
-            lpos_begin = 0
+            lpos_begin = ulong_size
         info = infos[idx].read_()
         text_len = info.text_len.value_()
         if lpos_next - lpos_begin < text_len:
--- a/drgn/helpers/linux/printk.py
+++ b/drgn/helpers/linux/printk.py
@@ -127,8 +127,8 @@ def _get_printk_records_lockless(prog: Program, prb: Object) -> List[PrintkRecor
         if not record_committed(current_id, desc.state_var.counter.value_()):
             return
 
-        lpos_begin = desc.text_blk_lpos.begin & text_data_ring_mask
-        lpos_next = desc.text_blk_lpos.next & text_data_ring_mask
+        lpos_begin = (desc.text_blk_lpos.begin & text_data_ring_mask).value_()
+        lpos_next = (desc.text_blk_lpos.next & text_data_ring_mask).value_()
         lpos_begin += ulong_size
 
         if lpos_begin == lpos_next:
@@ -136,9 +136,9 @@ def _get_printk_records_lockless(prog: Program, prb: Object) -> List[PrintkRecor
             return
         if lpos_begin > lpos_next:
             # Data wrapped.
-            lpos_begin -= lpos_begin
+            lpos_begin = 0
         info = infos[idx].read_()
-        text_len = info.text_len
+        text_len = info.text_len.value_()
         if lpos_next - lpos_begin < text_len:
             # Truncated record.
             text_len = lpos_next - lpos_begin
--- a/drgn/helpers/linux/printk.py
+++ b/drgn/helpers/linux/printk.py
@@ -103,6 +103,8 @@ def _get_printk_records_lockless(prog: Program, prb: Object) -> List[PrintkRecor
     LOG_CONT = prog["LOG_CONT"].value_()
     desc_committed = prog["desc_committed"].value_()
     desc_finalized = prog["desc_finalized"].value_()
+    LPOS_DATALESS = 0x1
+    EMPTY_LINE_LPOS = 0x3
 
     def record_committed(current_id: int, state_var: int) -> bool:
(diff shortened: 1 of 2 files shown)


osandov:
Thanks for looking into this! I modified the unit test to exercise this:

```diff
diff --git a/tests/linux_kernel/helpers/test_printk.py b/tests/linux_kernel/helpers/test_printk.py
index cf5505db..5947ea2e 100644
--- a/tests/linux_kernel/helpers/test_printk.py
+++ b/tests/linux_kernel/helpers/test_printk.py
@@ -3,6 +3,7 @@
 
 import errno
 import os
+from pathlib import Path
 import re
 
 from drgn.helpers.linux.printk import PrintkRecord, get_printk_records
@@ -54,7 +55,21 @@ def get_kmsg_records():
 
 
 class TestPrintk(LinuxKernelTestCase):
+    def _fill_log_buf(self):
+        printk_devkmsg_path = Path("/proc/sys/kernel/printk_devkmsg")
+        old_printk_devkmsg = printk_devkmsg_path.read_text()
+        try:
+            if old_printk_devkmsg.strip() != "on":
+                printk_devkmsg_path.write_text("on\n")
+            with open("/dev/kmsg", "ab", buffering=0) as f:
+                for i in range(self.prog["log_buf_len"].value_() // 16):
+                    f.write(f"<7>drgn test {i}\n".encode())
+        finally:
+            if old_printk_devkmsg.strip() != "on":
+                printk_devkmsg_path.write_text(old_printk_devkmsg)
+
     def test_get_printk_records(self):
+        self._fill_log_buf()
         self.assertEqual(
             get_kmsg_records(),
             [
```

And it looks like there's still an issue:

```console
$ python3 -m vmtest.vm python3 -m unittest tests/linux_kernel/helpers/test_printk.py
...
E
ERROR: test_get_printk_records (tests.linux_kernel.helpers.test_printk.TestPrintk.test_get_printk_records)
  File "/home/osandov/repos/drgn3/tests/linux_kernel/helpers/test_printk.py", line 85, in test_get_printk_records
    for record in get_printk_records(self.prog)
                  ~~~~~~~~~~~~~~~~~~^^^^^^^^^^^
  File "/home/osandov/repos/drgn3/drgn/helpers/common/prog.py", line 103, in wrapper
    return f(*args, **kwds)
  File "/home/osandov/repos/drgn3/drgn/helpers/linux/printk.py", line 255, in get_printk_records
    return _get_printk_records_lockless(prog, prb)
  File "/home/osandov/repos/drgn3/drgn/helpers/linux/printk.py", line 173, in _get_printk_records_lockless
    add_record(current_id)
    ~~~~~~~~~~^^^^^^^^^^^^
  File "/home/osandov/repos/drgn3/drgn/helpers/linux/printk.py", line 158, in add_record
    text=prog.read(text_data_ring_data + lpos_begin, text_len),
         ~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: negative size

Ran 1 test in 0.971s

FAILED (errors=1)
```

nilsstrassle:
I could not reproduce the error with this test, but I think I know why it occurs.

When a record ends exactly at the ring boundary, `lpos_next` is reduced to 0 by the mask, and is then treated like a wrapped record. This causes `lpos_next - lpos_begin` to be negative and sets `text_len` to that negative value in the truncated record path.

I fixed this by using `text_data_ring_size` for `lpos_next` when its masked value is 0.

The live-kernel test I used is quite large for just that one path (it has to position a record to end exactly at the data ring boundary, taking into account that there can be other writers), so I haven't included it in the PR. I instead added some synthetic testing to exercise each of the paths for a record (regular, wrapped, data-less and ending exactly at the boundary).

The testing led me to look into the data-less record handling:
```python
lpos_begin += ulong_size

if lpos_begin == lpos_next:
    # Data-less record.
    return
```
This misses the special lpos values ([`0x1` and `0x3`](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?h=v7.2&id=5113cf5f4c53eb2901ddb8fa70b72e8b6f4084ac)) because `lpos_begin` is advanced to skip the id before the data-less check.

To fix the issue, I check for a data-less block before adding `ulong_size` and use an approach similar to [the kernel](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/kernel/printk/printk_ringbuffer.c?id=bf56987c111372a54ae877934a42f7fb0953a6ca#n1270) for handling the special lpos values.

PR #637, opened 2026-07-31
RFC: bpfwalk — optionally run helper enumeration in-kernel via BPF iterators

imran-kn:
**Summary**

  drgn's structure-enumerating helpers (for_each_task, and lots of others) walk kernel data structures pointer-by-pointer
  from userspace — which is exactly what you want for portability: it works identically on core dumps, live kernels, old
  kernels, everything. This RFC explores an optional fast path for the live-kernel case: let a helper hand its whole
  traversal to a small in-kernel BPF iterator that enumerates the structure and returns the addresses, while drgn keeps
  doing all the type work. It's purely additive — when the fast path isn't available (core dump, remote, old kernel, no
  BPF privilege) the existing helper runs unchanged and nobody notices.

  I'd love feedback on whether this is a direction worth pursuing before I build it out further, hence RFC.

  **The idea**

  On a live kernel, the same enumeration a helper does in userspace can run inside a single BPF iterator invocation. A
  walker is a small SEC("iter/task")-style program that enumerates in-kernel and emits raw addresses; drgn turns each
  into an Object (or does container_of). Because only addresses cross the boundary and all typing stays in drgn, one 
  walker serves every helper that iterates the same structure, and the helpers' behavior/output is unchanged.

  A design note on where this hooks: the natural-looking seam is a BPF-backed drgn_memory_read_fn, but there drgn  still drives the loop and calls the reader once per element, so the number of round-trips doesn't really change. The speedup  comes specifically from moving the loop into the kernel — so the offload lives at the helper layer, where the whole traversal can be handed over at once.

  **What's in this PR**

  - A self-contained bpfwalk/ component (a BPF walker + a small userspace library with a session/cursor API and a
  data-driven walker registry — adding a walker is one .bpf.c plus one table row).
  - A ctypes binding and a gate (bpfwalk_objects()) under drgn/helpers/linux/.
  - for_each_task wired as the first consumer: a ~5-line guard that tries the fast path and otherwise falls through to
  today's walk.

  That guard is the entire integration surface for any future helper.

  **What I measured**

  On a live 5.15 kernel with ~1000 tasks, timing just the enumeration loop (drgn's DWARF load and a warmup pass
  excluded), a full for_each_task walk went from roughly 3.2 ms to about 0.8 ms — a bit over 4×. The reason is simply
  that the in-kernel walk is one read()-to-EOF on the iterator fd rather than a read per task, so the enumeration makes
  far fewer trips across the user/kernel boundary. It's an enumeration-only win — code that then reads many members per task will see a smaller overall difference — but for the common "iterate and filter/count" shape it's a nice speedup,
  and it grows with the number of objects.

  **Scope / limitations**

  - Live, local, BPF-privileged kernels only; everything else falls back transparently.
  - One walker (task) so far — this is about validating the shape, not breadth.
  - Deliberately minimal build/packaging (standalone Makefile, ctypes binding) to keep the RFC small.

  **Open questions**

  1. Is an optional in-kernel fast path for helper enumeration something drgn would want?
  2. Placement & build: standalone bpfwalk/ as here, vs contrib/, vs integrating into drgn's build; ctypes vs a proper
  _drgn C extension.
  3. The addresses-out / types-stay-in-drgn contract — does it fit how you'd want helpers to opt in?
  4. Consistency: a single-batch walk can take RCU/locks (potentially even a bit more consistent than a node-by-node
  userspace walk); multi-batch walks aren't atomic across batches, the same caveat that already applies today.
  Reasonable?
  5. If this is worth continuing, which helpers would be most valuable to add walkers for next (slab, xarray/idr, lists)?

  Happy to discuss this further or any suggestions regarding this


brenns10 (commented):
This looks like a really cool idea. It's great that you got this to work, it's one thing to have the concept, but another to get it functioning! I do have a few observations & questions about it.

1. The performance improvement is good, but as you mention, once you start accessing fields of a task, you lose most of that. The thing is, I can't really think of a use-case where we would iterate without accessing fields, except for counting tasks. And once you start adding in the overhead of the Python interpreter in that loop, I'm not sure that the optimization is worth the extra code and maintenance. Where I _do_ think there could be a clear win is for _lookups_, not iteration. Basically, accelerating helpers like `find_task()`, `find_module()`, `vma_find()`, etc.
2. It seems like this is only applicable to data structures that already have BPF iterator program types (or open-coded BPF iterators, maybe?). That is: tasks, files within a task, TCP/UDP sockets, cgroups. Do you have a strategy to broaden this to support arbitrary data structures? For example, the code explicitly mentions xarrays and slab caches as a potential use-case, but BPF doesn't provide a generic iterator for that.
3. Live drgn programs always have a risk of stale data, but this does seem to widen the gap of time between when we observe a pointer, and when we use it. So I wonder if this would actually make the risk of stale data _worse_, not better? It would be interesting to try this with a workload of rapid process creation and deletion, to see whether this is actually any better or worse.

PR #638, opened 2026-08-04
libdrgn: program: check for virtual address segments in drgn_program_set_linux_kernel_custom()

osandov:
@sdimitro, I think we forgot this from our conversation here: https://github.com/osandov/drgn/pull/587#discussion_r2756492924. Just wanted to make sure I won't be breaking you if I add it now.


sdimitro (approved):
LGTM and doesn't break our internal tool 👍

PR #643, opened 2026-08-26
PPC64: handle address translation when kernel virtual memory is not available

theotherjimmy:
When we only have physical memory, we need a few fixes to avoid running into recursive address translation errors:
 * Detecting radix mmu needs to avoid reading kernel structures
 * we need a method to translate the swapper_pg_dir from virtual memory to physical memory
 
 These two commits should address these two problems with more recent linux kernels.

### DIFF
--- a/libdrgn/arch_ppc64.c
+++ b/libdrgn/arch_ppc64.c
@@ -420,19 +420,21 @@ linux_kernel_pgtable_iterator_next_ppc64(struct drgn_program *prog,
 	static const uint64_t PAGE_PRESENT = UINT64_C(1) << 63;
 	static const uint64_t PAGE_PTE = UINT64_C(1) << 62;
 	static const uint64_t PT_MASK = UINT64_C(0xc0000000000000ff);
+	static const uint64_t KERNELBASE = UINT64_C(0xc000000000000000);
 	static const uint16_t levels = 4;
 	struct drgn_error *err;
 	struct pgtable_iterator_ppc64 *it_arch = it->arch;
 	uint64_t virt_addr = it->virt_addr;
 
+	uint64_t table = it->pgtable;
+	bool table_physical = false;
+	if (table == prog->vmcoreinfo.swapper_pg_dir) {
+		table -= KERNELBASE;
+		table_physical = true;
+	}
 	uint64_t entry;
 	for (uint16_t level = levels;; level--) {
-		uint64_t table;
-		bool table_physical;
-		if (level == levels) {
-			table = it->pgtable;
-			table_physical = false;
-		} else {
+		if (level != levels) {
 			// PAGE_PTE bit represents huge page.
 			if (!(entry & PAGE_PRESENT) || (entry & PAGE_PTE) || level == 0) {
 				uint64_t mask = (UINT64_C(1) << it_arch->pt_levels[level].shift) - 1;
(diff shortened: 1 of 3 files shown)


theotherjimmy:
@osandov This is the fixes I needed from #627

libdrgn/arch_ppc64.c
```diff
@@ -325,6 +325,38 @@ get_index(struct pgtable_iterator_ppc64 *it_arch, uint64_t va, uint16_t level)
...
 }
 
+static struct drgn_error *
+fail_if_not_radix_enabled(struct drgn_program * prog)
+{
+	// Note that this is a shortcut that allows working with physical
+	// memory only programs with VMCOREINFO. This was added in
+	// linux kernel commit 36e826b568e4 ("powerpc/vmcore: Add MMU
+	// information to vmcoreinfo"), in Linux 6.7.
+	if (prog->vmcoreinfo.ppc64_radix_mmu)
```
osandov: We probably want this to be a tristate: unknown (because it wasn't in vmcoreinfo), definitely radix, or definitely not. Right now, the definitely not radix case still falls through to the old path.

libdrgn/arch_ppc64.c
```diff
@@ -325,6 +325,38 @@ get_index(struct pgtable_iterator_ppc64 *it_arch, uint64_t va, uint16_t level)
...
+	    return NULL;
+
+	struct drgn_error *err = NULL;
```
osandov: This is overridden immediately, so the initialization is unnecessary.
```suggestion
	struct drgn_error *err;
```

libdrgn/arch_ppc64.c
```diff
@@ -325,6 +325,38 @@ get_index(struct pgtable_iterator_ppc64 *it_arch, uint64_t va, uint16_t level)
...
+	err = drgn_program_find_object(prog, "cur_cpu_spec", NULL,
+	                               DRGN_FIND_OBJECT_ANY, &tmp);
+	if (err)
+	        return err;
+	err = drgn_object_member_dereference(&tmp, &tmp, "mmu_features");
+	if (err)
+	        return err;
+	uint64_t mmu_features;
+	err = drgn_object_read_unsigned(&tmp, &mmu_features);
+
```
osandov: This stray newline was preexisting, but might as well fix it while you're moving it.
```suggestion
```

libdrgn/arch_ppc64.c
```diff
@@ -403,14 +421,15 @@ linux_kernel_pgtable_iterator_next_ppc64(struct drgn_program *prog,
 	struct pgtable_iterator_ppc64 *it_arch = it->arch;
 	uint64_t virt_addr = it->virt_addr;
 
+	uint64_t table = it->pgtable;
+	bool table_physical = false;
+	if (table == prog->vmcoreinfo.swapper_pg_dir) {
+		table -= prog->vmcoreinfo._stext;
```
osandov: Having looked at this after discussing in the meeting today, let's just go with the constant mask approach.

osandov (changes requested):

libdrgn/arch_ppc64.c
theotherjimmy: I added another boolean, `have_ppc64_radix_mmu`. I'm not sure this is the best name (open to suggestions), and I'm open to other ways to do this.

osandov (approved):
I just fixed some minor formatting issues and will go ahead and land this now, thank you!

PR #644, opened 2026-08-26
RFC: Kvm helpers

theotherjimmy:
This adds some simple helpers for finding and debugging kvm virtual machines.

### DIFF
--- a/drgn/helpers/linux/kvm.py
+++ b/drgn/helpers/linux/kvm.py
@@ -11,7 +11,7 @@ virtual machines and their components.
 
 from typing import Callable, Iterator, Optional, Union
 
-from drgn import FaultError, Object, Program, cast
+from drgn import FaultError, Object, Program, ProgramFlags, cast
 from drgn.helpers.common.prog import takes_program_or_default
 from drgn.helpers.linux.mm import access_remote_vm, for_each_vma
 from drgn.helpers.linux.pid import for_each_task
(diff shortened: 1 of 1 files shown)


drgn/helpers/linux/kvm.py
```diff
@@ -0,0 +1,170 @@
...
+
+    :param prog: :class:`~drgn.Program` or :ref:`omitted <default-program>`.
+    :return: Iterator of ``struct kvm *`` objects.
+    """
+    kvm_addr_set = {}
+    for task in for_each_task(prog):
+        kvm = kvm_from_task(task)
+        if kvm and kvm.value_() not in kvm_addr_set:
+            yield kvm
+            kvm_addr_set[kvm.value_()] = True
```
brenns10, line 43: You can iterate over `vm_list` (defined in `virt/kvm/kvm_main.c`) to directly get a list of `struct kvm` objects:

```suggestion
    try:
        yield from list_for_each_entry("struct kvm", prog["vm_list"].address_of_(), "vm_list")
    except ObjectNotFoundError:
        pass  # KVM module is not loaded, so there are no VMs
```

drgn/helpers/linux/kvm.py
```diff
@@ -0,0 +1,170 @@
...
+        )
+    if search_for_vmcoreinfo:
+        for vmcoreinfo_addr in prog.search_memory(b"VMCOREINFO").set_address_range(
+            physical=True
+        ):
+            vmcoreinfo_name_len = prog.read_u32(vmcoreinfo_addr - 0xC, physical=True)
+            vmcoreinfo_desc_len = prog.read_u32(vmcoreinfo_addr - 0x8, physical=True)
+            vmcoreinfo_type = prog.read_u32(vmcoreinfo_addr - 0x4, physical=True)
+            if vmcoreinfo_type != 0 or vmcoreinfo_name_len != 11:
+                continue
```
brenns10, line 160: This is a neat validation mechanism for the ELF note, nice!

drgn/helpers/linux/kvm.py
```diff
@@ -0,0 +1,170 @@
...
+
+    :param kvm: ``struct kvm *``
+    :return: Iterator of ``struct kvm_memory_slot *`` objects.
+    """
+    return rbtree_inorder_for_each_entry(
+        kvm.prog_.type("struct kvm_memory_slot"), kvm.memslots[0].gfn_tree, "gfn_node"
+    )
+
+
+def _make_read_user_memory(
```
brenns10, line 96: As a special case, if the host is live, it might make sense to try to use `/proc/$pid/mem` for reading the guest memory. Given that at this point you know the userspace address, I assume you could find the VMA and use a similar approach as in `contrib/qemu_mem.py`. I believe (but actually have not performance tested) that this is faster than going through drgn's /proc/kcore reader to read userspace memory.

brenns10 (commented):
Sorry, I sent the initial comment hastily, I had a few more thoughts to share.

I also wanted to make you aware of the drgn-tools kvm code as well: https://github.com/oracle-samples/drgn-tools/blob/main/drgn_tools/kvm.py

One difficulty with KVM-related stuff is testing. Drgn's CI kernels run within a VM, and on Github the CI is of course within its own VM, so it's nested virtualization. Getting KVM up and running inside the vmtest VM for the purpose of testing the KVM helpers is difficult, which is one of the reasons we hadn't upstreamed this code initially. We instead have vmcores we test against, but I honestly don't recommend that as a good testing approach.

drgn/helpers/linux/kvm.py
theotherjimmy: This is great, thanks! I'm also a bit embarrassed that I missed it.

